PR #6, opened 2020-01-31
Setup CI

tomchristie:
Setup our CI, which should...

* Ensure that `black` has been applied.
* Ensure that `mypy` verifies the typing correctly.
* Ensure that the `async`/`sync` variants match properly.
* Run the test suite. (Which will include both async and sync variants.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+dist: xenial
+language: python
+
+cache: pip
+
+branches:
+  only:
+    - master
+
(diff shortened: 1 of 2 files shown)


tomchristie:
Note that:

* We're checking both variants into source control right now. I think that's not a bad policy.
* We're just rolling with a custom `unasync.py` right now - we can always swap that out for a packaged version if/when possible.

PR #7, opened 2020-01-31
Cause `unasync --check` failure

tomchristie:
This PR makes a change in one of the sync modules, rather than making a change in the unasync modules and running `scripts/lint` to perform linting and unasync'ing.

It should fail the CI check stage.


tomchristie:
Screenshot 2020-01-31 at 13 12 24

👍

PR #8, opened 2020-01-31
Timeout as dict

tomchristie:
* Timeouts as `Dict[str, Optional[float]]`.
* Tweak docstring style.

I've used `Optional[float]` here. We could have just used `float`, and mandate that any missing key should be treated as `None`, but:

1. That's a bit awkward for user-code to build up the dict, given a config model.
2. It's nicer if we can have `None` as an explicit "Off". (Eg. allows us to provide non-off defaults if we want to.)


florimondmanca:
Nice. Will we want to specify the keys of the dict at some point, though?

PR #32, opened 2020-02-26
Add keepalive-expiry

tomchristie:


florimondmanca:
Marking this as a follow-up to encode/httpx#627

Issue #9, opened 2020-01-31
Use `aclose()` for async close methods.


tomchristie:

victoraugustolls:
Is this to just rename existing `async def close()` methods to `async def aclose()`? If it is, I can open a quick PR for this!

tomchristie:
Kinda yes. You'll also need to add `aclose`->`close` to the `scripts/unasync` script,  which is used to autogenerate the sync case from the async case. (And also used at CI to check that the two implementations matche up correctly.)

tomchristie:
Closed via  #35

PR #35, opened 2020-03-02
Async close rename

victoraugustolls:
PR to solve #9

- Updated `async close` methods to `async aclose`
- Updated `unasync.py` to map `aclose` -> `close`
- Updates async tests


tomchristie:
✨ *fantastic* ✨

Thank you!

Issue #18, opened 2020-02-06
Provide a `get_connection_stats()` interface.


tomchristie:
We should hide anything except `.request` and `.close` on the ConnectionPool, but we *might well* want to provide a `.get_connection_stats()` that returns eg...

```python
>>> print(http.get_connection_stats())
{(b'https', b'example.org', 443): {ConnectionState.IDLE: 1}}
```

*Above: A connection pool, maintaining a single keep-alive connection to https://example.org*

Noticable in particular in the test cases which happen to reach inside intended-private API in order to deteremine the connection state.

victoraugustolls:
would this include making the properties private?

eg. `max_connections` -> `_max_connections`, `connection_semaphore` -> `_connection_semaphore`, ...

tomchristie:
At this point in time really all the interface on `AsyncConnectionPool` / `SyncConnectionPool` *ought* to be made private, other than `request()` and `.close()`/`.aclose()`.

PR #37, opened 2020-03-02
Made ConnectionPool interface private

victoraugustolls:
Made ConnectionPool interface private, except .request and .aclose /.close

This addresses some comments in #18, but is missing `get_connection_stats()`

Update:

Added naive implementation to `get_connection_stats()`, will change if necessary!
Added only one test case for it, but if needed I can make more.


tomchristie:
Ah fantastic stuff thanks!

I'd probably suggest that we drop the `get_connection_stats` from this PR, and split that into an isolated issue, so that we can discuss it independantly without blocking the "make the interfaces private" part of this?

victoraugustolls:
Will do!

victoraugustolls:
@tomchristie done!

tomchristie:
Great thanks! Let's get this one in.

PR #40, opened 2020-03-03
Added get_connection_stats to pool

victoraugustolls:
Added `def get_connecction_stats()` to `ConnectionPool`.

Second part of #18

### DIFF
--- a/tests/async_tests/test_interfaces.py
+++ b/tests/async_tests/test_interfaces.py
@@ -173,3 +173,42 @@ async def test_http_request_cannot_reuse_dropped_connection():
         assert status_code == 200
         assert reason == b"OK"
         assert len(http._connections[url[:3]]) == 1
+
+
+@pytest.mark.usefixtures("async_environment")
+async def test_http_get_connection_stats():
+    async with httpcore.AsyncConnectionPool() as http:
+        method = b"GET"
+        url = (b"http", b"example.org", 80, b"/")
+        headers = [(b"host", b"example.org")]
+        http_version, status_code, reason, headers, stream = await http.request(
+            method, url, headers
+        )
+        body = await read_body(stream)
+
+        assert http_version == b"HTTP/1.1"
+        assert status_code == 200
+        assert reason == b"OK"
+        assert len(http._connections[url[:3]]) == 1
+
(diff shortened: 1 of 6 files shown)


tests/async_tests/test_interfaces.py
```diff
@@ -173,3 +173,42 @@
...
+        headers = [(b"host", b"example.org")]
+        http_version, status_code, reason, headers, stream = await http.request(
+            method, url, headers
+        )
+        body = await read_body(stream)
+
+        assert http_version == b"HTTP/1.1"
+        assert status_code == 200
+        assert reason == b"OK"
+        assert len(http._connections[url[:3]]) == 1
```
tomchristie: Let's drop this, in favor of the `assert http.get_connection_stats() == ...`

tests/async_tests/test_interfaces.py
```diff
@@ -173,3 +173,42 @@
...
```
tomchristie: Let's drop this in favor of an `http.get_connection_stats()` assert.

tomchristie:
Okay, so the main question here is "what we want the interface to look like for this". Eg. really we'd also like to expose which connections are HTTP/1.1 and which are HTTP/2, so perhaps we should be including that on the string too.

Alternately we could be exposing a differnt kinf of interface here, eg. return a set of `HTTPConnection` instances, and make sure that the interfaces on those are sufficiently constrained.

I think a good acid test on "have we got this right" would be to update the tests where possible to use `.get_connection_stats` instead of reaching into the private `._connections`. I'd suggest that if we update the test cases in this PR to do that where possible, then it'd give us a good indication of if we're exposing enough information through this interface or not.

victoraugustolls:
That makes total sense, will try to work on this today / tomorrow.

victoraugustolls:
@tomchristie I updated the interface to include the protocol version and updated the tests. Don't think `status_code` and `reason` should be included, but it is just a personal opinion!

yeraydiazdiaz:
Woops! Sorry @victoraugustolls, literally just saw this PR 😳 

Closing this as it's now superseded by https://github.com/encode/httpcore/pull/102, thanks 👍

Issue #42, opened 2020-03-06
Proxy tests


tomchristie:
I think a sensible tack onto the proxy tests would be if our test environment included a running client proxy, and we made live requests via it.

I'm wondering what the most well used client proxy services are that we'd be able to have easily  installed into our Travis / GitHub actions / Local developer setup, that we could use to exercise forwarding/tunneling requests?

Anyone got thoughts on which services we could easily have installed into the docker environment that'd meet our requirements here?

yeraydiazdiaz:
Closed via #48

PR #48, opened 2020-03-15
Use mitmproxy as a fixture for proxy tests

yeraydiazdiaz:
Towards #42 

In [socksio](https://github.com/sethmlarson/socksio/) we use [Dante](https://www.inet.no/dante/) to spin up SOCKS 4 and 5 proxies for acceptance tests in Docker but I tend to use [mitmproxy](https://mitmproxy.org/) for debugging proxy issues, since it's written in Python I thought it might be a simple non-docker solution to run it as a thread in the background.

This is a POC only, but it show how we could go about doing that. The tunnel test is failing at an assertion in the connection, not quite sure why but the forwarding ones work well.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -38,6 +38,9 @@ def proxy_server():
 
     class ProxyWrapper(threading.Thread):
         def run(self):
+            # mitmproxy uses asyncio internally but the default loop policy
+            # will only create event loops for the main thread, create one
+            # as part of the thread startup
             asyncio.set_event_loop(asyncio.new_event_loop())
             opts = options.Options(listen_host=host, listen_port=port)
             pconf = proxy.config.ProxyConfig(opts)
@@ -46,14 +49,14 @@ def proxy_server():
             self.master.server = proxy.server.ProxyServer(pconf)
             self.master.run()
 
-        def shutdown(self):
+        def join(self):
             self.master.shutdown()
+            super().join()
 
     try:
         thread = ProxyWrapper()
         thread.start()
-        time.sleep(2)
+        time.sleep(2)  # TODO: there's probably a better way to do this
         yield (host, port)
     finally:
-        thread.shutdown()
         thread.join()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,5 @@
 import asyncio
 import threading
-import time
 import typing
 
 import pytest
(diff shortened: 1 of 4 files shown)


tomchristie:
Ah fantastic, yup great stuff!

Looks like it needs adding to the requirements, then we can see if it's running okay in the Travis environment.

> The tunnel test is failing at an assertion in the connection, not quite sure why but the forwarding ones work well.

I'd be okay with us starting off by adding whatever we can as a first pass, and then incrementing from there. Seem reasonable?

yeraydiazdiaz:
> Seem reasonable?

Yep, this is now ready for review.

tests/conftest.py
```diff
@@ -25,3 +33,40 @@ def async_environment(request: typing.Any) -> str:
...
+        self.master.shutdown()
+        super().join()
+
+
+@pytest.fixture
+def proxy_server():
+    try:
+        thread = ProxyWrapper(PROXY_HOST, PROXY_PORT)
+        thread.start()
+        time.sleep(1)  # TODO: there's probably a better way to do this
```
florimondmanca: One quick dirty way might be to register an [addon](https://docs.mitmproxy.org/stable/addons-overview/) with the `.running()` callback…? [Masters trigger it just before running the mainloop](https://github.com/mitmproxy/mitmproxy/blob/e01f044c33aba51d9e5b6053e759b47594ad0855/mitmproxy/master.py#L77-L82). Eg:

```python
started = threading.Event()

class RunningNotify:
    def __init__(self):
        self.started = threading.Event()

    def running(self):
        self.started.set()

class ProxyWrapper:
    def __init__(self, ...):
        ...
        self.notify = RunningNotify()

    def run(self):
        ...
        master.addons.add(self.notify)
        ...

thread = ProxyWrapper(PROXY_HOST, PROXY_PORT)
thread.start()
thread.notify.started.wait()
```

yeraydiazdiaz: Nice find! I'll try it out, it's certainly dirty than the sleep 😄

yeraydiazdiaz: Worked like a charm, thanks!

yeraydiazdiaz:
I noticed something odd happening while finishing this one up. `unasync` keeps deleting the test for some reason but I can't see anything wrong with it. Any ideas?

yeraydiazdiaz:
Nevermind, fixed it now.

tests/sync_tests/test_interfaces.py
```diff
@@ -173,3 +173,20 @@ def test_http_request_cannot_reuse_dropped_connection():
         assert status_code == 200
         assert reason == b"OK"
         assert len(http._connections[url[:3]]) == 1
+
+
+@pytest.mark.parametrize("proxy_mode", ["DEFAULT", "FORWARD_ONLY"])
+
```
yeraydiazdiaz: This extra line is added by `unasync`, parametrization works fine though.

tomchristie:
@encode/httpx-maintainers Right - I've added `httpcore` write permission to the "httpx-maintainers" team.

PR #50, opened 2020-04-01
Bring dev tools up to par with httpx

yeraydiazdiaz:
Fix some typing and linting errors as a result.

PR #52, opened 2020-04-02
More typing fixes

yeraydiazdiaz:
I _think_ I got them all this time.

I left out changes in the tests as I believe we're not typing those in httpx either.

### DIFF
--- a/httpcore/_async/base.py
+++ b/httpcore/_async/base.py
@@ -2,7 +2,7 @@ import enum
 from types import TracebackType
 from typing import AsyncIterator, Callable, List, Optional, Tuple, Type
 
-from .._types import URL, Headers, TimeoutDict
+from .._types import HeadersType, TimeoutDictType, URLType
 
 
 async def empty() -> AsyncIterator:
@@ -78,10 +78,10 @@ class AsyncHTTPTransport:
     async def request(
         self,
         method: bytes,
-        url: URL,
-        headers: Headers = None,
+        url: URLType,
+        headers: HeadersType = None,
         stream: AsyncByteStream = None,
-        timeout: TimeoutDict = None,
+        timeout: TimeoutDictType = None,
     ) -> Tuple[bytes, int, bytes, List[Tuple[bytes, bytes]], AsyncByteStream]:
         """
         The interface for sending a single HTTP request, and returning a response.
--- a/httpcore/_async/base.py
+++ b/httpcore/_async/base.py
@@ -46,7 +46,7 @@ class AsyncByteStream:
 
     def __init__(
         self,
-        iterator: Optional[AsyncIterator[bytes]] = None,
+        iterator: AsyncIterator[bytes] = None,
         close_func: Callable = None,
     ) -> None:
         self.iterator = empty() if iterator is None else iterator
--- a/httpcore/_async/base.py
+++ b/httpcore/_async/base.py
@@ -1,6 +1,6 @@
 import enum
 from types import TracebackType
-from typing import AsyncIterator, Callable, List, Optional, Tuple, Type
+from typing import AsyncIterator, Callable, List, Tuple, Type
 
 from .._types import URL, Headers, TimeoutDict
 
@@ -45,9 +45,7 @@ class AsyncByteStream:
     """
 
     def __init__(
-        self,
-        iterator: AsyncIterator[bytes] = None,
-        close_func: Callable = None,
+        self, iterator: AsyncIterator[bytes] = None, close_func: Callable = None,
     ) -> None:
         self.iterator = empty() if iterator is None else iterator
         self.close_func = close_func
--- a/httpcore/_async/base.py
+++ b/httpcore/_async/base.py
@@ -47,7 +47,7 @@ class AsyncByteStream:
     def __init__(
         self,
         iterator: Optional[AsyncIterator[bytes]] = None,
-        close_func: Optional[Callable] = None,
+        close_func: Callable = None,
     ) -> None:
         self.iterator = empty() if iterator is None else iterator
         self.close_func = close_func
@@ -79,9 +79,9 @@ class AsyncHTTPTransport:
         self,
         method: bytes,
         url: URL,
-        headers: Optional[Headers] = None,
+        headers: Headers = None,
         stream: AsyncByteStream = None,
-        timeout: Optional[TimeoutDict] = None,
+        timeout: TimeoutDict = None,
     ) -> Tuple[bytes, int, bytes, List[Tuple[bytes, bytes]], AsyncByteStream]:
         """
         The interface for sending a single HTTP request, and returning a response.
@@ -122,8 +122,8 @@ class AsyncHTTPTransport:
 
     async def __aexit__(
         self,
-        exc_type: Optional[Type[BaseException]] = None,
-        exc_value: Optional[BaseException] = None,
-        traceback: Optional[TracebackType] = None,
+        exc_type: Type[BaseException] = None,
+        exc_value: BaseException = None,
+        traceback: TracebackType = None,
     ) -> None:
         await self.aclose()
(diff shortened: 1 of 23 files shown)


florimondmanca:
Thanks! Haven’t taken a look yet, but a note on this...

> I left out changes in the tests as I believe we're not typing those in httpx either.

Personally, I believe we should. Tests are where we get to use the lib as users would do IRL, so enabling type checking there is the only way we have to catch typing issues before they reach users. (Eg, I saw several typing fixes in Starlette due to usage of `List[SomeBaseClass]`, causing « List is invariant ».)

tests/async_tests/test_interfaces.py
```diff
@@ -27,11 +29,11 @@
         assert http_version == b"HTTP/1.1"
         assert status_code == 200
         assert reason == b"OK"
-        assert len(http._connections[url[:3]]) == 1
```
yeraydiazdiaz, line 30: It's slightly awkward to add all these `type: ignore` but I thought it was better than exposing `connections`.

httpcore/_async/base.py
```diff
@@ -43,7 +45,9 @@ class AsyncByteStream:
     """
 
     def __init__(
-        self, iterator: AsyncIterator[bytes] = None, close_func: Any = None
+        self,
+        iterator: Optional[AsyncIterator[bytes]] = None,
+        close_func: Optional[Callable] = None,
```
florimondmanca, line 50: Don't think these `Optional` markers are necessary with the default mypy config? Do you have a stricter mypy config locally?

```suggestion
        iterator: AsyncIterator[bytes] = None,
        close_func: Callable = None,
```

There are a couple of instances of this across this diff, not going to highlight them all, but I don't think we use `strict optional` in HTTPX.

httpcore/_async/base.py
```diff
@@ -74,10 +78,10 @@ class AsyncHTTPTransport:
     async def request(
         self,
         method: bytes,
-        url: Tuple[bytes, bytes, int, bytes],
-        headers: List[Tuple[bytes, bytes]] = None,
+        url: URL,
+        headers: Optional[Headers] = None,
```
florimondmanca, line 82: It doesn't seem like it was discussed in #50, but I _think_ there was originally an intentional decision for using explicit types on these datastructures, instead of type aliases.

I'm _personally_ okay with using private aliases, but maybe we should switch the naming to `URLType`, `HeadersType`, etc, to clarify that they're not objects but just aliases to builtin types.

httpcore/_async/http_proxy.py
```diff
@@ -25,24 +21,30 @@ class AsyncHTTPProxy(AsyncConnectionPool):
...
+    * **proxy_mode** - `str` - A proxy mode to operate in. May be "DEFAULT",
+    "FORWARD_ONLY", or "TUNNEL_ONLY".
+    * **ssl_context** - `Optional[SSLContext]` - An SSL context to use for
+    verifying connections.
+    * **max_connections** - `Optional[int]` - The maximum number of concurrent
+    connections to allow.
+    * **max_keepalive** - `Optional[int]` - The maximum number of connections
+    to allow before closing keep-alive connections.
     * **http2** - `bool` - Enable HTTP/2 support.
     """
```
florimondmanca, line 37: Can we instead ignore these flake8 errors…?

```suggestion
    """  # noqa: E501
```

httpcore/_backends/auto.py
```diff
@@ -1,10 +1,13 @@
...
-from typing import Dict, Optional
+from typing import Optional
 
 import sniffio
 
+from .._types import TimeoutDict
 from .base import AsyncBackend, AsyncLock, AsyncSemaphore, AsyncSocketStream
-from .sync import SyncBackend, SyncLock, SyncSemaphore, SyncSocketStream
+
+# The following line is imported from the _sync modules
```
florimondmanca, line 9: Curious: why don't the `_sync/*.py` modules import from eg `_backends.sync` instead? Is it because of `unasync`?

Treating auto as "works both for sync and async" location seems a bit hacky. Should we instead make `_async` and `_sync` import from `_backends` directly, i.e. define an `__all__` in `_backends/__init__.py`?

(Not a comment on this PR in particular, just something we could address separately.)

tests/async_tests/test_interfaces.py
florimondmanca: Agreed, `# type: ignore` are fine here IMO. :)

httpcore/_async/base.py
yeraydiazdiaz: We don't, no. Those were not mypy errors but I changed them as it seemed clearer to me. I'll change it back 👍

florimondmanca: Oh okay. Yeah in general we've been relying on implicit `Optional` as documented here: https://mypy.readthedocs.io/en/stable/command_line.html#none-and-optional-handling

httpcore/_async/base.py
yeraydiazdiaz: So I went with what was discussed on https://github.com/encode/httpx/issues/789, though I guess I should've held off until it was resolved.

It is kinda the point of this PR though, maybe I should just close it?

httpcore/_backends/auto.py
yeraydiazdiaz: I don't really know why tbh, it did seem a bit odd but out of scope so I left it. Your suggestion sounds sensible though.

httpcore/_async/base.py
yeraydiazdiaz: Actually, since `_types` is already in master it's probably better to just move forwards with this model, even if it's only for consistency.

httpcore/_async/http_proxy.py
yeraydiazdiaz: Looks like we're [not ignoring them in HTTPX either](https://github.com/encode/httpx/blob/master/setup.cfg#L2) and can't find any `noqa: E510` comments either.

I personally prefer not to scroll right to see the whole docstring but happy to go with whatever is decided.

httpcore/_async/base.py
```diff
@@ -43,7 +45,9 @@ class AsyncByteStream:
     """
 
     def __init__(
-        self, iterator: AsyncIterator[bytes] = None, close_func: Any = None
+        self,
+        iterator: Optional[AsyncIterator[bytes]] = None,
```
florimondmanca, line 49: Last one :)

```suggestion
        iterator: AsyncIterator[bytes] = None,
```

httpcore/_types.py
```diff
@@ -4,8 +4,8 @@
...
-StrOrBytes = Union[str, bytes]
-Origin = Tuple[bytes, bytes, int]
-URL = Tuple[bytes, bytes, int, bytes]
-Headers = List[Tuple[bytes, bytes]]
-TimeoutDict = Dict[str, Optional[float]]
+StrOrBytesType = Union[str, bytes]
+OriginType = Tuple[bytes, bytes, int]
+URLType = Tuple[bytes, bytes, int, bytes]
+HeadersType = List[Tuple[bytes, bytes]]
+TimeoutDictType = Dict[str, Optional[float]]
```
florimondmanca, line 11: @yeraydiazdiaz Are you okay with these changes? Now that I see it across the codebase the `*Type` suffix looks kinda redundant, ("of course it's a `Type`, it's in an annotation…"), so I'm actually not too sure it's a good idea…

httpcore/_async/base.py
yeraydiazdiaz: Agh, thanks! 😄

httpcore/_types.py
yeraydiazdiaz: I don't have a strong opinion tbh. I guess it is a bit redundant, I'll switch it back.

Issue #54, opened 2020-04-09
TUNNEL_ONLY proxy mode fails


yeraydiazdiaz:
From https://github.com/encode/httpx/issues/859#issuecomment-611206532

```
Traceback (most recent call last):
  File "client.py", line 13, in fetch
    r = await client.get(target)
  File "/Users/florimond/Developer/python-projects/httpx/httpx/_client.py", line 1279, in get
    return await self.request(
  File "/Users/florimond/Developer/python-projects/httpx/httpx/_client.py", line 1121, in request
    response = await self.send(
  File "/Users/florimond/Developer/python-projects/httpx/httpx/_client.py", line 1142, in send
    response = await self.send_handling_redirects(
  File "/Users/florimond/Developer/python-projects/httpx/httpx/_client.py", line 1169, in send_handling_redirects
    response = await self.send_handling_auth(
  File "/Users/florimond/Developer/python-projects/httpx/httpx/_client.py", line 1206, in send_handling_auth
    response = await self.send_single_request(request, timeout)
  File "/Users/florimond/Developer/python-projects/httpx/httpx/_client.py", line 1238, in send_single_request
    ) = await dispatcher.request(
  File "/Users/florimond/Developer/python-projects/httpx/venv/lib/python3.8/site-packages/httpcore/_async/http_proxy.py", line 81, in request
    return await self._tunnel_request(
  File "/Users/florimond/Developer/python-projects/httpx/venv/lib/python3.8/site-packages/httpcore/_async/http_proxy.py", line 155, in _tunnel_request
    proxy_response = await connection.request(
  File "/Users/florimond/Developer/python-projects/httpx/venv/lib/python3.8/site-packages/httpcore/_async/connection.py", line 52, in request
    assert url[:3] == self.origin
AssertionError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "client.py", line 24, in <module>
    loop.run_until_complete(main())
  File "/Users/florimond/.pyenv/versions/3.8.2/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "client.py", line 20, in main
    await asyncio.gather(*tasks)
  File "client.py", line 15, in fetch
    num_ok += 1
  File "/Users/florimond/Developer/python-projects/httpx/httpx/_client.py", line 1454, in __aexit__
    await self.aclose()
  File "/Users/florimond/Developer/python-projects/httpx/httpx/_client.py", line 1443, in aclose
[...]

PR #55, opened 2020-04-09
Fix tunnel proxy and support HTTPS

yeraydiazdiaz:
Fixes https://github.com/encode/httpcore/issues/54

Probably helps with https://github.com/encode/httpx/issues/897

- First establish an HTTP connection to our proxy since `CONNECT` is an HTTP method handled by h11.
- Once successful discard the connection but prevent closing the socket.
- Upgrade to TLS if required.
- Allow passing a socket to the `AsyncHTTPConnection` and create a new h11 connection to the target.

Only HTTP is covered by the tests using mitmproxy, I couldn't quite get HTTPS requests to work using mitmproxy but it did work using Squid as @florimondmanca [outlined in the HTTPX issue](https://github.com/encode/httpx/issues/859#issuecomment-611206532) (thanks for that, it was very helpful 👍 )

These changes sit a bit at the edge of our abstractions since the socket is now exposed a few levels above what it used to. I'm keen to hear your thoughts on whether we need some refactoring.

### DIFF
--- a/httpcore/_async/http11.py
+++ b/httpcore/_async/http11.py
@@ -123,7 +123,7 @@ class AsyncHTTP11Connection(AsyncHTTPTransport):
             event = await self._receive_event(timeout)
             if isinstance(event, h11.Data):
                 yield bytes(event.data)
-            elif isinstance(event, h11.EndOfMessage):
+            elif isinstance(event, (h11.EndOfMessage, h11.PAUSED)):
                 break
 
     async def _receive_event(self, timeout: TimeoutDict) -> H11Event:
--- a/httpcore/_async/http_proxy.py
+++ b/httpcore/_async/http_proxy.py
@@ -170,9 +170,10 @@ class AsyncHTTPProxy(AsyncConnectionPool):
                 msg = "%d %s" % (proxy_status_code, proxy_reason_phrase.decode("ascii"))
                 raise ProxyError(msg)
 
-            # Upgrade to TLS.
-            # TODO: upgrade only on HTTPS targets?
-            # await proxy_connection.start_tls(target, timeout)
+            # Upgrade to TLS if required
+            # We assume the target speaks TLS on the specified port
+            if url[0] == b"https":
+                await proxy_connection.start_tls(target, timeout)
 
             # Create a new connection to the target
             connection = AsyncHTTPConnection(
--- a/httpcore/_async/http_proxy.py
+++ b/httpcore/_async/http_proxy.py
@@ -140,6 +140,7 @@ class AsyncHTTPProxy(AsyncConnectionPool):
         connection = await self._get_connection_from_pool(origin)
 
         if connection is None:
+            # First, create a connection to the proxy server
             proxy_connection = AsyncHTTPConnection(
                 origin=self.proxy_origin, http2=False, ssl_context=self._ssl_context,
             )
@@ -147,7 +148,7 @@ class AsyncHTTPProxy(AsyncConnectionPool):
                 self._connections.setdefault(origin, set())
                 self._connections[origin].add(proxy_connection)
 
-            # Establish the connection by issuing a CONNECT request...
+            # Issue a CONNECT request...
 
             # CONNECT www.example.org:80 HTTP/1.1
             # [proxy-headers]
@@ -173,7 +174,7 @@ class AsyncHTTPProxy(AsyncConnectionPool):
             # Upgrade to TLS if required
             # We assume the target speaks TLS on the specified port
             if url[0] == b"https":
-                await proxy_connection.start_tls(target, timeout)
+                await proxy_connection.start_tls(url[1], timeout)
 
             # Create a new connection to the target
             connection = AsyncHTTPConnection(
--- a/httpcore/_async/http_proxy.py
+++ b/httpcore/_async/http_proxy.py
@@ -140,12 +140,12 @@ class AsyncHTTPProxy(AsyncConnectionPool):
         connection = await self._get_connection_from_pool(origin)
 
         if connection is None:
-            connection = AsyncHTTPConnection(
-                origin=origin, http2=False, ssl_context=self._ssl_context,
+            proxy_connection = AsyncHTTPConnection(
+                origin=self.proxy_origin, http2=False, ssl_context=self._ssl_context,
             )
             async with self._thread_lock:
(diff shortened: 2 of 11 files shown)


httpcore/_async/http11.py
```diff
@@ -123,7 +123,7 @@ def mark_as_ready(self) -> None:
             event = await self._receive_event(timeout)
             if isinstance(event, h11.Data):
                 yield bytes(event.data)
-            elif isinstance(event, h11.EndOfMessage):
+            elif isinstance(event, (h11.EndOfMessage, h11.PAUSED)):
```
yeraydiazdiaz, line 126: From the [h11 docs](https://h11.readthedocs.io/en/latest/api.html#flow-control):

> PAUSED is a little more subtle: [...] A successful CONNECT or Upgrade: request has caused the connection to switch to some other protocol (see Switching protocols). This PAUSED state is permanent; you should abandon this Connection and go do whatever it is you’re going to do with your new protocol.

httpcore/_async/http_proxy.py
```diff
@@ -140,47 +140,56 @@ def __init__(
...
 
             # Establish the connection by issuing a CONNECT request...
 
             # CONNECT www.example.org:80 HTTP/1.1
             # [proxy-headers]
             target = b"%b:%d" % (url[1], url[2])
             connect_url = self.proxy_origin + (target,)
-            connect_headers = self.proxy_headers
-            proxy_response = await connection.request(
+            connect_headers = self.proxy_headers or [(b"host", self.proxy_origin[1])]
```
yeraydiazdiaz, line 156: The host header is required by h11 but not typically passed as part of the proxy headers.

Not sure if we want to merge the headers here or leave that to the caller to sort out.

tests/async_tests/test_interfaces.py
```diff
@@ -182,14 +183,37 @@
...
         )
-        body = await read_body(stream)
+        _ = await read_body(stream)
+
+        assert http_version == b"HTTP/1.1"
+        assert status_code == 200
+        assert reason == b"OK"
+
+
+@pytest.mark.parametrize("proxy_mode", ["TUNNEL_ONLY"])
```
yeraydiazdiaz, line 200: Forwarding HTTPS requests doesn't work with mitmproxy or squid, but I'm not sure if it's by design or a problem with our code, any ideas?

florimondmanca: Do you mean...

- Sending HTTPS Requests through the proxy? Works for me, at least to external hosts - never managed to set up a proxy that could connect to a localhost server with custom certs.
- Connecting to the proxy via HTTPS (ie the proxy URL is https://...)? Never got it to work, mostly because I don’t know of a proxy that supports it/how to set it up on Squid/mitm.

yeraydiazdiaz: I mean the former, sending requests to `https://example.org` via a `mitmproxy` instance fails with a 400 and a mysterious HTML response content reading `HttpException("Invalid request scheme: https")`. Which is why I was guessing it's simply not supported, but I can't find concrete proof of that.

> Works for me, at least to external hosts

Any details on how you're achieving that? I'm basically recreating what the tests are doing with `trustme-cli`. Adding `FORWARD_ONLY` to the parametrization results in exactly the same problem.

florimondmanca: > Any details on how you're achieving that?

It worked with Squid, haven't tried with MITM proxy though. For mitmproxy, [How mitmproxy works](https://docs.mitmproxy.org/stable/concepts-howmitmproxyworks/) contains very detailed info about what happens when a client issues an HTTPS request to forward, but I can't wrap my ahead around what actually needs to be configured for this to work…

yeraydiazdiaz: Huh, I just tried it with Squid and got:

`b'HTTP/1.1' 501 b'Not Implemented'` and a bunch of headers and HTML from the Squid server.

Here's the script I'm running to test, in case I'm missing something. The Squid setup is as in your comment.

```
import asyncio
import httpcore

async def read_body(stream):
    try:
        body = []
        async for chunk in stream:
            body.append(chunk)
        return b"".join(body)
    finally:
        await stream.aclose()

async def tunnel_proxy(proxy_server, proxy_mode):
    async with httpcore.AsyncHTTPProxy(proxy_server, proxy_mode=proxy_mode) as http:
        method = b"GET"
        url = (b"https", b"example.org", 443, b"/")
        http_version, status_code, reason, headers, stream = await http.request(
            method, url, headers
        )
        print(http_version, status_code, reason, headers)
        body = await read_body(stream)
        print(body)

        assert http_version == b"HTTP/1.1"
        assert status_code == 200
        assert reason == b"OK"


def main():
    proxy_server = (b"http", b"localhost", 3128)
    proxy_mode = "FORWARD_ONLY"
    asyncio.run(tunnel_proxy(proxy_server, proxy_mode))

if __name__ == "__main__":
    main()
```

yeraydiazdiaz: I spent some time researching this and I can't find any definite answer but given what we wrote on the [HTTPX proxy docs](https://www.python-httpx.org/advanced/#http-proxying):

> To make sure that proxies cannot read your traffic, and even if the proxy_url uses HTTPS, it is recommended to use HTTPS and tunnel requests if possible.

The fact that we default to tunnelling HTTPS requests and forward HTTP ones, and that AFAICT neither mitmproxy or Squid implement forwarding HTTPS requests so I'm going to assume it's discouraged but presumably possible to forward HTTPS requests.

I changed the name of the test and made a note to reflect that.

Thanks for looking into it. 👍

yeraydiazdiaz:
A pesky warning in mitmproxy led me to find and fix a bug, yay warnings! 😄

This is now ready for a final review. 🙏

florimondmanca:
Thanks! Haven't got around to reviewing this, but will try soon.

Side note — I noticed the "_and_" in the PR title, is there a way "Fix tunnel proxy" and "HTTPS support" could be two separate, smaller-size (and easier to review?) PRs, or are these two too entangled? I'll admit I'm still a proxy noob so smaller diffs would help me provide better quality reviews on these matters. :)

yeraydiazdiaz:
Fair enough, let me see what I can do.

yeraydiazdiaz:
Closing in favor of #57.

PR #69, opened 2020-05-01
Enable type checking for Trio

JayH5:
Using [trio-typing](https://github.com/python-trio/trio-typing)

### DIFF
--- a/httpcore/_backends/trio.py
+++ b/httpcore/_backends/trio.py
@@ -1,5 +1,5 @@
 from ssl import SSLContext
-from typing import Optional, Union
+from typing import Optional
 
 import trio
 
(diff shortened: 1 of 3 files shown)


httpcore/_backends/trio.py
```diff
@@ -43,7 +43,9 @@ async def start_tls(
             trio.BrokenResourceError: ConnectError,
         }
         ssl_stream = trio.SSLStream(
-            self.stream, ssl_context=ssl_context, server_hostname=hostname
+            self.stream,
+            ssl_context=ssl_context,
+            server_hostname=hostname.decode("ascii"),
```
JayH5, line 48: This wasn't actually a bug I think because the SSLContext always encodes the hostname:
https://github.com/python/cpython/blob/7b3ab5921fa25ed8b97b6296f97c5c78aacf5447/Lib/ssl.py#L486-L492

...but lot's of things (including the typeshed definitions of the stdlib `SSLContext`) type the hostname as `Optional[str]`. And we decode with `ascii` in the asyncio backend as well.

yeraydiazdiaz: Sounds fair enough 👍

PR #70, opened 2020-05-02
Fix forward proxy adding connection to pool

yeraydiazdiaz:
Fixes https://github.com/encode/httpx/issues/922

The issue was that we were manually adding the connection to the pool rather than using the `_add_to_pool` method. By doing this we never acquired (or created) the semaphore, so when attempting to close the connection a new one was created and tried to be released, raising the BoundedSemaphore exception.

Interestingly, this didn't show up in our test suite because when `max_connections` is `None` which is the default we use a `NullSemaphore`. I added both `max_connections` and `max_keepalive` to our proxy tests to cover this particular PR, maybe we should add them to all tests?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+
+- Fix connections using proxy forwarding requests not being added to the
+connection pool properly.
+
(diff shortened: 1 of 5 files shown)


florimondmanca:
@yeraydiazdiaz So if I understood correctly, the new workflow should allow us to issue a 0.8.2 release with this patch after a release PR, correct? :-)

yeraydiazdiaz:
Yep, merging a release PR should be enough. Want to give it a go?

florimondmanca:
Sure! 👍

Issue #64, opened 2020-04-30
Status Badges


tomchristie:
We ought to add our "Test Suite" and PyPI badges to the README and to `docs/index.md`

PR #73, opened 2020-05-04
Add badges to README

yeraydiazdiaz:
Closes https://github.com/encode/httpcore/issues/64

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,14 @@
 # HTTP Core
 
+<p>
+    <a href="https://github.com/encode/httpcore/actions">
+        <img src="https://github.com/encode/httpcore/workflows/Test%20Suite/badge.svg" alt="Test Suite">
+    </a>
+    <a href="https://pypi.org/project/httpcore/">
+        <img src="https://badge.fury.io/py/httpcore.svg" alt="Package version">
+    </a>
+</p>
+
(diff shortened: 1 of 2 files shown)


tomchristie:
Great! I guess let's have the badges on the `docs/index.md` too?

tomchristie:
Related to this: Occurred to me that the Test Suite badge probably *isn't actually* giving us the "show the results of the latest merge to master".

We might need to switch the config so that we *are* building again on the final commit to master? (Right now we're only building against PRs, and have enforced the "PRs must be up to date with master setting") so that the badges are actually reflecting the status of PRs once commited?

yeraydiazdiaz:
Personally I don't mind, but I guess it might be confusing for people clicking on that link and not seeing `master` which could lead to some unnecessary interaction.

tomchristie:
> Personally I don't mind, but I guess it might be confusing for people clicking on that link and not seeing master which could lead to some unnecessary interaction.

It's not the link that's an issue, so much as the badge colour / status check. Right now, I'm not sure it'll even be getting updated, since it's a badge based on "default branch", that's only running against PRs (not the final merge-to-master), and the PRs themselves are never *on* master branch.

It's just not clear to me if the badge behaviour with our current config *is* actually meaningful.

PR #74, opened 2020-05-04
Run test suite on "merge to master" actions.

tomchristie:
So I *think* we need to also run our test suite on the final merge to master, not just on the PR itself.

If we're not doing this, then it's not clear to me that the associated badge is going to actually reflect "colour of the last PR that was merged to master". Having said that, it's actually a bit inscrutable to me.

Something that might help thrash this out would be to leave things as they currently are, and issue a PR with *deliberately* failing tests, and find the answer to...

* Does issuing the PR on it's own update the badge colour?
* Does merging the PR update the badge colour?

I realise I've not worded any of this very clearly, but it's actually not as obvious as you might think that the current test suite behaviour we've got is associated with a meaningful badge.

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -2,6 +2,8 @@
 name: Test Suite
 
 on:
+  push:
+    branches: ["master"]
   pull_request:
     branches: ["master"]
 


florimondmanca:
Related docs: https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository

> By default, badges display the status of your default branch (usually master). You can also display the status of a workflow run for a specific branch or event using the branch and event query parameters in the URL.

Apparently it would be possible to show the status of the latest built PR, but I don’t think it’s what the status badge is aimed at - ie showing whether the build pipeline is successful on the latest _merged_ PR. At this point I’d agree we should just turn on master builds again...

tomchristie:
Okay, let's do this then.
We'll want the same change on a whole bunch of other encode repos too. (httpx in particular)

PR #77, opened 2020-05-06
Fix up proxy header behaviour

tomchristie:
Refs https://github.com/encode/httpx/issues/929#issuecomment-624127681

* Add `Host` and `Accept: */*` headers by default on the tunnel CONNECT requests, if they're not already set.
* De-duplicate proxy headers vs request headers on forwarded requests, rather than naively concatenating them.

I think we should probably roll a new version of `httpcore` from this, then issue an `httpx 0.13.dev1`.

### DIFF
--- a/httpcore/_async/http_proxy.py
+++ b/httpcore/_async/http_proxy.py
@@ -8,14 +8,20 @@ from .connection import AsyncHTTPConnection
 from .connection_pool import AsyncConnectionPool, ResponseByteStream
 
 
-def merge_headers(default_headers: Headers = None, override_headers: Headers = None) -> Headers:
+def merge_headers(
+    default_headers: Headers = None, override_headers: Headers = None
+) -> Headers:
     """
     Append default_headers and override_headers, de-duplicating if a key existing in both cases.
     """
     default_headers = [] if default_headers is None else default_headers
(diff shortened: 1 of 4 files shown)


httpcore/_async/http_proxy.py
```diff
@@ -144,8 +161,10 @@ async def _tunnel_request(
             # [proxy-headers]
             target = b"%b:%d" % (url[1], url[2])
             connect_url = self.proxy_origin + (target,)
+            connect_headers = [(b"Host", target), (b"Accept", b"*/*")]
+            connect_headers = merge_headers(connect_headers, self.proxy_headers)
             proxy_response = await proxy_connection.request(
-                b"CONNECT", connect_url, headers=self.proxy_headers, timeout=timeout
+                b"CONNECT", connect_url, headers=connect_headers, timeout=timeout
```
yeraydiazdiaz, line 167: We should probably remove the [setting of `host` headers in the tests](https://github.com/encode/httpcore/blob/master/tests/async_tests/test_interfaces.py#L188-L193) to cover things are working as expected.

tomchristie: Good call!

PR #78, opened 2020-05-06
Version 0.8.3

tomchristie:
@yeraydiazdiaz I guess we *don't* actually need a `httpx 0.13.dev1` - we can just advise `pip install httpx --pre -U`.

Tho perhaps *might* want to issue `httpcore 0.9` alongside the release version of `httpx 0.13`, so that we can pin httpx to 0.9+?

### DIFF
--- a/httpcore/__init__.py
+++ b/httpcore/__init__.py
@@ -39,4 +39,4 @@ __all__ = [
     "WriteError",
     "CloseError",
 ]
-__version__ = "0.8.2"
+__version__ = "0.8.3"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## Unreleased
+## 0.8.3 (May 6rd, 2020)
 
 ### Fixed
 
+- Include `Host` and `Accept` headers on proxy "CONNECT" requests.
+- De-duplicate any headers also contained in proxy_headers.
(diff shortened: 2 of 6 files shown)


yeraydiazdiaz:
> I guess we don't actually need a httpx 0.13.dev1 - we can just advise pip install httpx --pre -U

I think that alone [wouldn't upgrade `httpcore`](https://pip.pypa.io/en/stable/reference/pip_install/?highlight=upgrade_strategy#cmdoption-u) though. Maybe we should cut the release pinning to `>=0.8.3` so it's less confusing?

> Tho perhaps might want to issue httpcore 0.9 alongside the release version of httpx 0.13, so that we can pin httpx to 0.9+?

Yup, probably should do that too.

httpcore/__init__.py
```diff
@@ -39,4 +39,4 @@
     "WriteError",
     "CloseError",
 ]
-__version__ = "0.8.2"
+__version__ = "0.8.3"
```
yeraydiazdiaz, line 42: Probably should update the CHANGELOG as well.

tomchristie: Blergh, accidentally pushed that to master, but yup.

tomchristie:
## 0.8.3 (May 6rd, 2020)

### Fixed

- Include `Host` and `Accept` headers on proxy "CONNECT" requests.
- De-duplicate any headers also contained in proxy_headers.
- HTTP/2 flag not being passed down to proxy connections.

tomchristie:
> Maybe we should cut the release pinning to >=0.8.3 so it's less confusing?

Sure thing.

tomchristie:
Released.

PR #84, opened 2020-05-12
Use `wait_closed` with asyncio, with socket unwrapping workaround.

tomchristie:
Proposal for working around https://bugs.python.org/issue39758 and addressing resulting bugs such as https://github.com/encode/httpx/issues/825 and https://github.com/encode/httpx/issues/914

Really we *ought* to be calling `.wait_closed` on closing sockets in asyncio, but we've not been able to because of a bug in cpython that can cause it to hang indefinitely on SSL termination. A workaround here is to explicitly deal with the socket unwrapping ourselves.

I've issued this as a draft PR since at the *very* least it'd need some verification.


florimondmanca:
I think this got superseded by #167? Closing then…

PR #89, opened 2020-05-13
Honor max concurrent streams

### LINKED ISSUE #85: Don't exceed `MAX_CONCURRENT_STREAMS` when sending HTTP/2 requests
Bug described in encode/httpx#281 where HTTPX doesn't support waiting for other streams to close before creating a new stream on an HTTP/2 connection.

tomchristie:
Closes #85

Work based off #38

### DIFF
--- a/httpcore/_async/http2.py
+++ b/httpcore/_async/http2.py
@@ -8,8 +8,8 @@ from h2.config import H2Configuration
 from h2.exceptions import NoAvailableStreamIDError
 from h2.settings import SettingCodes, Settings
 
-from .._backends.auto import AsyncLock, AsyncSocketStream, AutoBackend
-from .._exceptions import ProtocolError
+from .._backends.auto import AsyncLock, AsyncSemaphore, AsyncSocketStream, AutoBackend
+from .._exceptions import PoolTimeout, ProtocolError
 from .._types import URL, Headers, TimeoutDict
 from .._utils import get_logger
 from .base import (
@@ -66,6 +66,17 @@ class AsyncHTTP2Connection(AsyncHTTPTransport):
         if not hasattr(self, "_read_lock"):
             self._read_lock = self.backend.create_lock()
         return self._read_lock
+
+    @property
+    def max_streams_semaphore(self) -> AsyncSemaphore:
+        # We do this lazily, to make sure backend autodetection always
+        # runs within an async context.
+        if not hasattr(self, "_max_streams_semaphore"):
+            max_streams = self.h2_state.remote_settings.max_concurrent_streams
+            self._max_streams_semaphore = self.backend.create_semaphore(
+                max_streams, exc_class=PoolTimeout
+            )
+        return self._max_streams_semaphore
 
     async def start_tls(self, hostname: bytes, timeout: TimeoutDict = None) -> None:
         pass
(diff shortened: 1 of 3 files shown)


httpcore/_async/http2.py
```diff
@@ -265,16 +276,21 @@ async def request(
...
-        if has_body:
-            await self.send_body(stream, timeout)
-
-        # Receive the response.
-        status_code, headers = await self.receive_response(timeout)
-        reason_phrase = get_reason_phrase(status_code)
-        stream = AsyncByteStream(
-            aiterator=self.body_iter(timeout), aclose_func=self._response_closed
-        )
+        await self.connection.max_streams_semaphore.acquire()
```
yeraydiazdiaz, line 279: I'm not really familiar with the mechanics of HTTP/2 but it occurs to me that maybe we ought to not block if no more streams are available but rather create a new connection?

That might be a separate issue though.

httpcore/_async/http2.py
```diff
@@ -346,4 +362,5 @@ async def body_iter(self, timeout: TimeoutDict) -> AsyncIterator[bytes]:
                 break
 
     async def _response_closed(self) -> None:
+        self.connection.max_streams_semaphore.release()
```
yeraydiazdiaz, line 365: I'm thinking testing this one is tricky in our current high-level testing policy in `test_interfaces` as we need quite a lot of requests to be performed to an external server which seems a bit much.

Here in httpcore we don't have a `server` fixture, we could introduce one but it'd have to support HTTP/2 to cover this case.

tomchristie:
I’ve not seen that no - is that before or after this change, or both?

yeraydiazdiaz:
That happens on master as well, but usually fails with a "no more streams available" error which your change fixes.

httpcore/_backends/asyncio.py
```diff
@@ -158,6 +158,18 @@ async def aclose(self) -> None:
...
+                try:
+                    ssl_object = self.stream_writer.get_extra_info("ssl_object")
+                    if ssl_object is not None:
+                        ssl_object.unwrap()
+                except SSLWantReadError:
+                    pass
+                else:
+                    if hasattr(self.stream_writer, "wait_closed"):
+                        # Python 3.7+
+                        await self.stream_writer.wait_closed()  # type: ignore
```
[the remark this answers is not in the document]
tomchristie: Ugh, that's me accidentally branching this work off #84 instead of off master.

httpcore/_async/http2.py
tomchristie: Different issue - this is about the maximum number of *concurrent* streams, rather than the maximum stream id before a new connection should be reissued.

yeraydiazdiaz: Right, makes sense 👍

httpcore/_async/http2.py
```diff
@@ -265,16 +276,21 @@ async def request(
...
+            if has_body:
+                await self.send_body(stream, timeout)
+
+            # Receive the response.
+            status_code, headers = await self.receive_response(timeout)
+            reason_phrase = get_reason_phrase(status_code)
+            stream = AsyncByteStream(
+                aiterator=self.body_iter(timeout), aclose_func=self._response_closed
+            )
+        except:
```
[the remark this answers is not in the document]
tomchristie: We don't actually care, we just want to make sure that we release the semaphore in one of these two cases:

* We fail to return, because some (any) exception was raised.
* We return okay, and the client code fulfils it's contract to call `stream.close()`, which calls into `_response_closed`.

httpcore/_async/http2.py
tomchristie: Indeed. We've at least got test *coverage* over the lines, which is a good first pass, even if we're not testing all kinds of behaviour.

httpcore/_async/http2.py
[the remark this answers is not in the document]
yeraydiazdiaz: Right, in that case we ought to use `Exception` rather than a bare `except`, I think flake8 recommends that, not sure why that's not flagged in the linting.

httpcore/_async/http2.py
yeraydiazdiaz: Right, it was more of an overall comment on our testing strategy moving forwards, not that much to do with this particular PR. We'll address separately 👍

tomchristie:
I've addressed the `error_code:ErrorCodes.REFUSED_STREAM` issue now.
The semaphore release in `_response_closed` needs to occur *after* the network close.

tomchristie:
Righty. I think we're there now.
Thanks so much for your review on this one - it was *invaluable*.

yeraydiazdiaz:
My pleasure! Awesome work 🚀

PR #92, opened 2020-05-20
Port becomes optional

tomchristie:
* Port in URL becomes optional.
* Origin stays as `Tuple[bytes, bytes, int]`, since we need a canonical form there.
* `proxy_origin` becomes `proxy_url` for better consistency.

HTTPX would need to update the `proxy_origin` form it uses to `proxy_url`.

### DIFF
--- a/httpcore/_async/http_proxy.py
+++ b/httpcore/_async/http_proxy.py
@@ -3,7 +3,7 @@ from typing import Tuple
 
 from .._exceptions import ProxyError
 from .._types import URL, Headers, Origin, TimeoutDict
-from .._utils import get_logger
+from .._utils import get_logger, url_to_origin
 from .base import AsyncByteStream
 from .connection import AsyncHTTPConnection
 from .connection_pool import AsyncConnectionPool, ResponseByteStream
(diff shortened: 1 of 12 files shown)


httpcore/_async/http_proxy.py
```diff
@@ -51,7 +51,7 @@ class AsyncHTTPProxy(AsyncConnectionPool):
 
     def __init__(
         self,
-        proxy_origin: Origin,
+        proxy_url: URL,
```
yeraydiazdiaz, line 54: OMG thanks! 😅

The docstring above needs to be updated.

httpcore/_utils.py
```diff
@@ -47,3 +48,10 @@ def trace(message: str, *args: typing.Any, **kwargs: typing.Any) -> None:
     logger.trace = trace  # type: ignore
 
     return typing.cast(Logger, logger)
+
+
+def url_to_origin(url: URL) -> Origin:
+    scheme, host, explicit_port = url[:3]
+    default_port = {b'http': 80, b'https': 443}[scheme]
```
florimondmanca, line 55: I was curious why Black didn't cough up on this usage of single quotes, only to realize it's not run in `scripts/check`.

It's been a while since I've been around here but I assume it (and the isort check) was removed on purpose from there?

PR #96, opened 2020-05-27
Implement AsyncHTTPConnection.aclose

schmir-at-bb:
The base class implementation is empty and python will otherwise emit
a "ResourceWarning unclosed <socket.socket ...>" for the unclosed
sockets.

### DIFF
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -126,3 +126,7 @@ class AsyncHTTPConnection(AsyncHTTPTransport):
             await self.connection.start_tls(hostname, timeout)
             logger.trace("start_tls complete hostname=%r timeout=%r", hostname, timeout)
             self.socket = self.connection.socket
+
+    async def aclose(self) -> None:
+        if self.socket:
+            await self.socket.aclose()


tomchristie:
Hiya!

So, httpcore is a low-level HTTP client, which means that you're required to ensure you're closing connections, eg. as documented in the quickstart...

```python
    http_version, status_code, reason_phrase, headers, stream = await http.request(
        method=b'GET',
        url=(b'https', b'example.org', 443, b'/'),
    )

    try:
        body = b''.join(chunk async for chunk in stream)
    finally:
        await stream.close()
```

It's really important to use the `try ... finally` pattern if you're using httpcore at a low level.

Higher level clients like `httpx` will take care of this *for* you. Ie. httpx always manages the close for you, either with basic requests, or with `.stream()` using a context-managed API.

Having said that, we *should* forcibly close outstanding connections, yup, although this PR isn't quite the right approach. (I'm also surprised that it passed the tests - since the sync & async cases aren't in sync here.)

tomchristie:
Closing in favour of #98

schmir-at-bb:
I did run into this using `httpx`, which does call `aclose` on the `AsyncHTTPConnection`. I've used the `async with httpx.AsyncClient() as client` context manager stuff.

Should this have worked without the fix? Is there another bug in httpx?

tomchristie:
Ah gotcha. Anyways yup, without #98 we're not properly closing connections when the pool is cleaned up. We'll release a 0.9.1 version shortly. Thanks for having filed this!

Issue #88, opened 2020-05-13
Advanced connection options.


tomchristie:
Prompted by a comment from @hynek

We ought to add more `__init__` controls to `AsyncConnectionPool(...)` and `SyncConnectionPool(...)`  for advanced connection options, including...

* UDS support. https://github.com/encode/httpx/pull/511, https://github.com/encode/httpx/pull/726
* System IP to use https://github.com/encode/httpx/issues/755
* Socket family and related options.

It'd be useful to do a comprehensive review of...

* What connection options are offered by asyncio?
* What connection options are offered by stdlib's standard sync networking API?
* What connection options are offered by trio?

Compare against controls available in `urllib3`, `aiohttp`.

hynek:
For reference, this is what you can do with `aiohttp`: https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.TCPConnector

tomchristie:
@hynek - Gotcha. The DNS caching isn't something I think we need to introduce, since it adds less reliable side-effects. Perhaps it's a feasible feature at some point, but let's treat it as out of scope for purposes here.

Out of the other stuff:

* What would you consider a priority here?
* Can we enumerate those options / namings, and do they all map to `asyncio.open_connection` options? Are they all also available in sync and trio backends?

hynek:
I _personally_ need to be able to force IPv4 by setting the socket family to `AF_INET` to firewall whitelisting shenanigans. No clue how common that problem is though.

That could be abstracted away of course to be more use friendly.

bwelling:
I'm planning to look at adding source_address support to the ConnectionPool objects.

@hynek - I believe that choosing an address family would be functionally equivalent to specifying a wildcard source address in that family; that is, using `AF_INET` would be the same as using a source address of `0.0.0.0`.  This could mean that the ConnectionPool objects internally converted a family to a source address.

bwelling:
Adding source address support is surprisingly complicated, unfortunately.  Getting the address down to the various `open_tcp_stream` methods isn't a problem, but the backends are more complicated.

In the sync backend, the code (currently) assumes IPv4 when opening a socket.  If this restriction is kept, the only change would be to optionally call `sock.bind()`, but not having IPv6 support seems bad.  Using `socket.create_connection` might be an improvement.

[...]

PR #100, opened 2020-05-27
Local address support.

bwelling:
Based on the discussion in #88 .

This isn't really complete yet.  The open issues are (at least).

- There's no testing.  I haven't even done any manual testing; I mostly wanted to verify that the changes look like a good start.

- The raised exceptions should be improved.

- mypy reports a bogus error.  See https://github.com/python/typeshed/issues/4116

- It would be nice to support passing a raw string as local_addr (if you just want to specify an address, and don't care about port).  I haven't done that yet.


bwelling:
The typing issue has been fixed upstream (https://github.com/python/typeshed/pull/4133).

bwelling:
Any comments on this approach?

florimondmanca:
I'm a bit far away from discussions that happened in #88 but this looks sensible enough to me. :-)

florimondmanca:
> It would be nice to support passing a raw string as local_addr

An option could be to mark the port there as `Optional[int]`, like we did in `URL` via #92.

bwelling:
Making the port an `Optional[int]` would mean having to pass an `(<address>, None)` tuple, which isn't really any better than `(<address>, 0)`.  I think using something like `Union[SocketAddress, str, bytes, bytearray]` would work, but I was running into typing issues when I first tried it.

florimondmanca:
Hmm, a general design philosophy I've seen in `httpcore` is simplicity/minimalism in terms of interfaces and usage.

Eg I'm not sure what the meaning of `Union[SocketAddress, str, bytes, bytearray]` would be; I'd assume this would mostly be for the sake of _convenience_. But from my perspective `httpcore` is not designed to be convenient (eg look at the kind of tuple one needs to pass for URLs, it looks pretty unusual). We should require users pass only what is required and nothing more - any smarts around encoding/decoding of such an address should be up to higher-level clients like HTTPX.

Which is why I was suggesting the `Optional` approach as a way to _literally_ transcribe from "pass a tuple containing the hostname and _an optional port_", while following the lead of #92. :-)

bwelling:
Fair enough.  If passing in a tuple from higher layers is ok, there's no reason for the port to be optional; callers can pass in 0.

bwelling:
There are some tests now, although it's hard to test too much in an automated fashion, since proper testing requires the system to have multiple routable addresses and working IPv6, and testing explicit source port numbers requires that the ports are unused.  I've done manual testing for all of these.

Setting local_addr requires setting family, as asyncio requires that.  Maybe it would be better to use something like a (family, address, port) tuple, instead of separate fields for family and address.  This could allow address=None, or require that address be set to the appropriate wildcard address for the family if the address isn't important.

The bogus error flagged by mypy is now suppressed; hopefully the typeshed fix will be picked up at some point.

I think this is ready for a more formal review.

httpcore/_async/connection_pool.py
```diff
@@ -76,6 +76,9 @@ class AsyncConnectionPool(AsyncHTTPTransport):
     * **keepalive_expiry** - `Optional[float]` - The maximum time to allow
     before closing a keep-alive connection.
     * **http2** - `bool` - Enable HTTP/2 support.
+    * **family** - `Optional[int]` - Address family to use
```
yeraydiazdiaz, line 79: Nit:

```suggestion
    * **family** - `int` - Address family to use, defaults to 0.
```

tests/async_tests/test_interfaces.py
```diff
@@ -204,6 +205,46 @@ async def test_http_proxy(
...
+        assert len(http._connections[url[:3]]) == 1  # type: ignore
+
+
+@pytest.mark.parametrize("local_addr", ["0.0.0.0"])
+@pytest.mark.asyncio
+# This doesn't run with trio, since trio doesn't support local_addr.
+async def test_http_request_local_addr(local_addr: str) -> None:
+    family = socket.AF_INET6 if ':' in local_addr else socket.AF_INET
+    async with httpcore.AsyncConnectionPool(family=family,
+                                            local_addr=(local_addr, 0)) as http:
```
yeraydiazdiaz, line 233: Huh, I would've expected Black to reformat this 🤔

tests/async_tests/test_interfaces.py
```diff
@@ -204,6 +205,46 @@ async def test_http_proxy(
...
+        assert status_code == 200
+        assert reason == b"OK"
+        assert len(http._connections[url[:3]]) == 1  # type: ignore
+
+
+@pytest.mark.parametrize("local_addr", ["0.0.0.0"])
+@pytest.mark.asyncio
+# This doesn't run with trio, since trio doesn't support local_addr.
+async def test_http_request_local_addr(local_addr: str) -> None:
+    family = socket.AF_INET6 if ':' in local_addr else socket.AF_INET
```
yeraydiazdiaz, line 231: This seems leftover from when there was more than one value in the parametrization, might be worth removing it altogether?

tests/async_tests/test_interfaces.py
```diff
@@ -204,6 +205,46 @@ async def test_http_proxy(
         assert reason == b"OK"
 
 
+@pytest.mark.parametrize("family", [socket.AF_INET])
```
yeraydiazdiaz, line 208: Probably worth just setting it below?

httpcore/_async/connection_pool.py
bwelling: I'll fix this.

tests/async_tests/test_interfaces.py
bwelling: The test really should be trying more than one case, but when I tried testing IPv6, it failed due to the test system not supporting IPv6.  I'd rather leave this to make it more clear that it should be parametrized, but if it's important, it can be removed.

tests/async_tests/test_interfaces.py
bwelling: As above, this originally tested AF_INET6 also, but the test machine doesn't support IPv6.

tests/async_tests/test_interfaces.py
bwelling: I must have forgotten to run the reformatting script before my last commit; it's done now.

bwelling:
> Thanks @bwelling! 💯
> 
> Left some minor comments on the current implementation.
> 
> > Setting local_addr requires setting family, as asyncio requires that. Maybe it would be better to use something like a (family, address, port) tuple, instead of separate fields for family and address. This could allow address=None, or require that address be set to the appropriate wildcard address for the family if the address isn't important.
> 
> I think it might be best to stay with the current approach, follow httpcore's "you know what you're doing" philosophy and leave ergonomics to HTTPX. I'd love to hear other's thoughts on this though.

I agree with this, but I'm not sure which approach matches "you know what you're doing" more; having 2 linked parameters or 1 more complicated parameter?  That is:

a) Have separate family (int) and local_addr (tuple of address and port) parameters (as the current patch does)

b) Have a single optional parameter (maybe called `source`) containing (family, address, port), where address could be None.

c) Have a single optional parameter containing (family, address, port), where all 3 fields must not be None (that is, address would be set to the appropriate wildcard address for the family if the specific address was unimportant but family was)

I don't think there's a reason to support more than 1 of those choices, but I don't know which of them best matches the current philosophy.

httpcore/_async/connection_pool.py
ntninja: Shouldn't the type of the `family` parameter be `socket.AddressFamily` rather than just `int`?

njsmith:
Is specifying the source port useful? I wasn't planning to support that in Trio, since it's kind of inherently incompatible with any kind of happy-eyeballs support... (And similarly, Twisted's version of this feature only lets you specify the source host, not the source port.)

bwelling:
> Is specifying the source port useful? I wasn't planning to support that in Trio, since it's kind of inherently incompatible with any kind of happy-eyeballs support... (And similarly, Twisted's version of this feature only lets you specify the source host, not the source port.)

In the case where you care about happy-eyeballs support, probably not.  But if you're specifying a single IPv4 or IPv6 source address (which this patch does), you explicitly don't, and being able to specify the source port is a pretty reasonable thing to do.

httpcore/_async/connection_pool.py
bwelling: If all of occurrences of `0` were replaced with `socket.AF_UNSPEC`, then sure.  I don't want to make this change yet, though, since if there's consensus to combine family and address into a single field, then there wouldn't be a need for either one, because the entire field would be `None`.

njsmith:
> In the case where you care about happy-eyeballs support, probably not. But if you're specifying a single IPv4 or IPv6 source address (which this patch does), you explicitly don't

Happy eyeballs is still useful if you're restricting to just IPv4 or just IPv6. Lots of website DNS names resolve to multiple IPv4/IPv6 addresses. E.g.:

```
❯ host debian.org
debian.org has address 149.20.4.15
debian.org has address 128.31.0.62
debian.org has address 130.89.148.77
debian.org has IPv6 address 2603:400a:ffff:bb8::801f:3e
debian.org has IPv6 address 2001:4f8:1:c::15
debian.org has IPv6 address 2001:67c:2564:a119::77
```

> being able to specify the source port is a pretty reasonable thing to do.

I didn't ask if it was reasonable, I asked if it was useful :-). Supporting source ports isn't trivial: the exact desired semantics aren't obvious, and writing good tests is tricky. It can be done, obviously, but if you want us to do the work then it'd be nice to have more of a justification. I've looked for use cases, and I haven't found any yet that care about source ports. (See: https://github.com/python-trio/trio/issues/275#issuecomment-648771086)

An interesting case is the requests-toolbelt `SourceAddressAdapter`. It *does* support setting the source port, *but* has a big warning saying that if you try to actually use it then it will break stuff, so please don't: https://toolbelt.readthedocs.io/en/latest/adapters.html#sourceaddressadapter

bwelling:
> I didn't ask if it was reasonable, I asked if it was useful :-). Supporting source ports isn't trivial: the exact desired semantics aren't obvious, and writing good tests is tricky. It can be done, obviously, but if you want us to do the work then it'd be nice to have more of a justification. I've looked for use cases, and I haven't found any yet that care about source ports. (See: [python-trio/trio#275 (comment)](https://github.com/python-trio/trio/issues/275#issuecomment-648771086))

After thinking about this some more, you're probably right.  My use case here is DoH (dns-over-https) queries from dnspython, and dnspython does allow specifying source port.  That's mostly historical in this case, though; while it's important for UDP in some cases, it's only really important in archaic situations for TCP, and probably not at all useful for HTTPS.

Would it be better to remove source-port support from this PR altogether, or have the trio backend raise an exception if a nonzero port is specified?

njsmith:
> Would it be better to remove source-port support from this PR altogether, or have the trio backend raise an exception if a nonzero port is specified?

I guess that's a judgement call for Tom to make, but if it were my project I would definitely remove the support entirely until a compelling use case comes along. For this kind of infrastructure project, it's *way* easier to add features later than it is to claw back features you added in a burst of enthusiasm and then later ended up regretting.

bwelling:
Just following up, is there anything else I need to do here?  We need this functionality in dnspython for asynchronous DNS-over-HTTPS and HTTP/2 support.

httpcore/_types.py
```diff
@@ -9,3 +9,4 @@
 URL = Tuple[bytes, bytes, Optional[int], bytes]
 Headers = List[Tuple[bytes, bytes]]
 TimeoutDict = Dict[str, Optional[float]]
+SocketAddress = Tuple[Union[str, bytes, bytearray], int]
```
tomchristie, line 12: Would it be reasonable to keep this as tightly constrained as possible?
Presumably `Tuple[bytes, int]` is sufficient right?

tomchristie:
Just getting back into this...

> I would definitely remove the [source-port] support entirely until a compelling use case comes along. For this kind of infrastructure project, it's way easier to add features later than it is to claw back features you added in a burst of enthusiasm and then later ended up regretting.

100% yup.

Having separate arguments seems to make sense to me, but it might make sense to keep this PR as tightly scoped as possible and *just* start with adding the local address support?

httpcore/_types.py
bwelling: If source-port is removed, this type isn't needed at all, as the address could just be a `bytes`.

tomchristie: Right, yup.

bwelling:
I just pushed an update that removes address family and source-port support; this now just allows setting a local address, and passes it through.

Note that this still doesn't support trio, as the trio change that @ntninja mentioned earlier hasn't been released yet.

bwelling:
It's been about 3 weeks since I addressed the last set of comments, and nothing's happened since then.  Is there anything I can do to get this merged?

httpcore/_async/connection.py
```diff
@@ -23,11 +23,13 @@ def __init__(
         http2: bool = False,
         ssl_context: SSLContext = None,
         socket: AsyncSocketStream = None,
+        local_addr: bytes = None,
```
tomchristie, line 26: Just to check - Do we want `local_addr` or `source_addr` here?
What are trio, asyncio, and the sync stdlib using for their naming?

httpcore/_async/connection_pool.py
```diff
@@ -76,6 +76,7 @@ class AsyncConnectionPool(AsyncHTTPTransport):
     * **keepalive_expiry** - `Optional[float]` - The maximum time to allow
     before closing a keep-alive connection.
     * **http2** - `bool` - Enable HTTP/2 support.
+    * **local_addr** - `Optional[bytes]` - Local address to connect from
```
tomchristie, line 79: Miniature nitpick. Let's use a full stop to match the other cases. "Local address to connect from." 😃

httpcore/_backends/asyncio.py
```diff
@@ -225,13 +225,20 @@ async def open_tcp_stream(
...
         timeout: TimeoutDict,
+        local_addr: Optional[bytes],
     ) -> SocketStream:
         host = hostname.decode("ascii")
         connect_timeout = timeout.get("connect")
         exc_map = {asyncio.TimeoutError: ConnectTimeout, OSError: ConnectError}
         with map_exceptions(exc_map):
+            local_addrport = None
+            if local_addr:
+                local_addrport = (local_addr, 0)
```
tomchristie, line 236:
```python
local_addrport = None if local_addr is None else (local_addr, 0)
```

httpcore/_backends/auto.py
```diff
@@ -34,8 +34,11 @@ async def open_tcp_stream(
         port: int,
         ssl_context: Optional[SSLContext],
         timeout: TimeoutDict,
+        local_addr: Optional[bytes],
     ) -> AsyncSocketStream:
-        return await self.backend.open_tcp_stream(hostname, port, ssl_context, timeout)
+        return await self.backend.open_tcp_stream(
+            hostname, port, ssl_context, timeout, local_addr
```
tomchristie, line 40: I reckon we ought to use `.open_tcp_stream(..., local_addr=local_addr) ` here.
Just makes it super obvious visually that it's an optional extra.

httpcore/_backends/trio.py
```diff
@@ -143,7 +143,10 @@ async def open_tcp_stream(
         port: int,
         ssl_context: Optional[SSLContext],
         timeout: TimeoutDict,
+        local_addr: Optional[bytes],
     ) -> AsyncSocketStream:
+        if local_addr:
```
tomchristie, line 148: Maybe a comment that it is currently supported in `trio` master, and should be expected from version 0.16.1 onwards?

Also, what will our implementation change look like once trio support does land here?

tomchristie:
Referencing the commit in trio that adds their support for this, mostly just because they do a great job on docs, and we might want to follow their lead here. https://github.com/python-trio/trio/pull/1644/files

httpcore/_async/connection.py
bwelling: asyncio uses `local_addr`, trio uses `local_address`, and the stdlib socket module uses `source_address` (and curio uses `source_addr`).  There's really no consistency.

httpcore/_async/connection_pool.py
bwelling: Fixed.

httpcore/_backends/asyncio.py
bwelling: Fixed.

httpcore/_backends/auto.py
bwelling: Fixed.

httpcore/_backends/trio.py
bwelling: I'm adding a comment.

I believe the code would just add something like `local_address=local_addr` to the `open_tcp_stream` call, although it's possible that there would be a type mismatch, as the `local_addr` parameter is `None` or a `bytes`, and trio claims to want `None` or a `str` (although it just passes it into `socket.bind`, which should accept a `bytes`).

tomchristie:
Great stuff - I've made some tiny amendments in https://github.com/encode/httpcore/pull/134. As soon as that's reviewed we'll get it in, and look at rolling our 0.10 release.

bwelling:
I might be missing something, but this was linked to by the httpx 0.14 release PR (https://github.com/encode/httpx/pull/1083).  These changes were merged into #134 and pulled up from there, but nothing was actually added to httpx 0.14, and there's nothing in that release PR or any other PR/issue indicating why not.

tomchristie:
@bwelling Indeed - that's because we don't *actually need* any code changes in httpx to support this.

For "local_address=...", and for our planned "uds=..." support, we'll *require* users to drop down to the low-level `httpcore` API, and install the transport using `Client(transport=...)`

I've drafted a PR with some docs notes that should hopefully clear this up... https://github.com/encode/httpx/pull/1165

bwelling:
The doc referenced explains how to use it, but that seems like an especially fragile interface.  If httpcore isn't designed to be user-friendly (which is fine), then requiring users to use it seems suboptimal.  Having to create a separate transport adapter to set custom and somewhat uncommon options is a perfectly reasonable API design, but requiring users to also pass in a bunch of unrelated configuration options with no defaults is a recipe for disaster, as it means that users who want the standard client behavior, except for one thing, need to poll httpx over time to see if its defaults change, or if new fields are added or removed from the default set.

Providing wrappers in httpx around `httpcore.SyncConnectionPool` and `httpcore.AsyncConnectionPool` that merge a set of user-provided configuration options with the defaults would go a long way toward solving this.

tomchristie:
We could certainly think about switching our policy on httpcore defaults, yup.

It's not really been a very exposed issue until just now.

bwelling:
Should I open an issue for that?

tomchristie:
Welcome to, yup.

There's a bunch of stuff to think through. (Eg. What should our default SSLConfig be & how do we plan to keep things in sync between httpcore/httpx)

Issue #44, opened 2020-03-11
Write Error on reusing potentially closed connection


victoraugustolls:
With a service deployed with httpx version with httpcore interface, I started receiving the following error:

`<class 'httpcore._exceptions.WriteError'>: [Errno 104] Connection reset by peer`

The only problem is that this is when connecting to a service that others applications have no problems. The connection is (should be) HTTP/1.1 in this case.

tomchristie:
I'm going to close this one off given the ancedotal https://github.com/encode/httpx/issues/841#issuecomment-627865072 "I just tested the original gist against httpx==0.13.0.dev2 and got the same result for http1.1 and http2."

If there's something reproducible here or any anecdotal reports of similar connection closing then we can reopen it, but given that it's just "the server closed the connection" there's not a lot here for us to act on.

victoraugustolls:
I have a service that after upgrading httpx to 0.13 started receiving the same error, and rolling back to 0.12 seems to resolve the issue.

victoraugustolls:
I don't know what details I can give here, other that:

- The problems seems to happen when the connections is not used for some minutes.
- My service runs under fast api.
- The only upgraded dependency was httpx.

victoraugustolls:
Stack trace:

File "/usr/local/lib/python3.8/site-packages/httpx/_client.py", line 1364, in post
    return await self.request(
  File "/usr/local/lib/python3.8/site-packages/httpx/_client.py", line 1137, in request
    response = await self.send(
  File "/usr/local/lib/python3.8/site-packages/httpx/_client.py", line 1158, in send
    response = await self.send_handling_redirects(
  File "/usr/local/lib/python3.8/site-packages/httpx/_client.py", line 1185, in send_handling_redirects
    response = await self.send_handling_auth(
  File "/usr/local/lib/python3.8/site-packages/httpx/_client.py", line 1222, in send_handling_auth
    response = await self.send_single_request(request, timeout)
  File "/usr/local/lib/python3.8/site-packages/httpx/_client.py", line 1254, in send_single_request
    ) = await transport.request(
  File "/usr/local/lib/python3.8/site-packages/httpcore/_async/connection_pool.py", line 152, in request
    response = await connection.request(
  File "/usr/local/lib/python3.8/site-packages/httpcore/_async/connection.py", line 78, in request
    return await self.connection.request(method, url, headers, stream, timeout)
[...]

PR #101, opened 2020-05-28
Add trace logs to get_connection_from_pool

yeraydiazdiaz:
To help debug issues like https://github.com/encode/httpcore/issues/44

### DIFF
--- a/httpcore/_async/http11.py
+++ b/httpcore/_async/http11.py
@@ -34,6 +34,9 @@ class AsyncHTTP11Connection(AsyncHTTPTransport):
 
         self.state = ConnectionState.ACTIVE
 
+    def __repr__(self) -> str:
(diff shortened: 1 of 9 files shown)


httpcore/_async/http11.py
```diff
@@ -34,6 +34,9 @@ def __init__(
 
         self.state = ConnectionState.ACTIVE
 
+    def __repr__(self) -> str:
+        return f"<AsyncHTTP11Connection state={self.state}>"
```
florimondmanca, line 38: Are we actually ever using this and the HTTP/2 equivalent? I _think_ the logs in the connection pool will only ever show `AsyncHTTPConnection`.

yeraydiazdiaz: Sure, but while debugging I find it useful to have some additional information about the connection 🙂

PR #102, opened 2020-05-28
Add connection pool get_connection_stats method

yeraydiazdiaz:
Closes https://github.com/encode/httpcore/issues/18

### DIFF
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -1,5 +1,5 @@
 from ssl import SSLContext
-from typing import List, Optional, Tuple, Union, cast
+from typing import List, Optional, Tuple, Union
 
 from .._backends.auto import AsyncLock, AsyncSocketStream, AutoBackend
 from .._types import URL, Headers, Origin, TimeoutDict
@@ -44,11 +44,7 @@ class AsyncHTTPConnection(AsyncHTTPTransport):
             return "Not connected"
         elif self.state == ConnectionState.PENDING:
             return "Connecting"
-        elif self.is_http11:
-            return f"HTTP/1.1, {self.state.name}"
-        else:
-            connection = cast(AsyncHTTP2Connection, self.connection)
-            return f"HTTP/2, {self.state.name}, {len(connection.streams)} streams"
+        return self.connection.info()
 
     @property
     def request_lock(self) -> AsyncLock:
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -1,5 +1,5 @@
 from ssl import SSLContext
-from typing import List, Optional, Tuple, Union
+from typing import List, Optional, Tuple, Union, cast
 
 from .._backends.auto import AsyncLock, AsyncSocketStream, AutoBackend
 from .._types import URL, Headers, Origin, TimeoutDict
@@ -38,6 +38,17 @@ class AsyncHTTPConnection(AsyncHTTPTransport):
         self.connect_failed = False
         self.expires_at: Optional[float] = None
         self.backend = AutoBackend()
+
+    def info(self) -> str:
+        if self.connection is None:
+            return "Not connected"
+        elif self.state == ConnectionState.PENDING:
+            return "Connecting"
+        elif self.is_http11:
+            return f"HTTP/1.1, {self.state.name}"
+        else:
+            connection = cast(AsyncHTTP2Connection, self.connection)
+            return f"HTTP/2, {self.state.name}, {len(connection.streams)} streams"
 
     @property
     def request_lock(self) -> AsyncLock:
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from ssl import SSLContext
 from typing import AsyncIterator, Callable, Dict, Optional, Set, Tuple
 
@@ -294,3 +295,15 @@ class AsyncConnectionPool(AsyncHTTPTransport):
         # Close all connections
         for connection in connections:
             await connection.aclose()
+
+    def get_connection_stats(self) -> Dict[Origin, Dict[str, int]]:
+        """Returns statistics dict around state of the connection pool in the form:
+
+        { origin: { connection_state: number_of_connections_on_that_state } }
+        """
+        stats: Dict[Origin, Dict[str, int]] = {}
+        for origin, connections in self._connections.items():
+            stats[origin] = defaultdict(lambda: 0)
+            for connection in connections:
+                stats[origin][connection.state.name] += 1
+        return stats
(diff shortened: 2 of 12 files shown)


httpcore/_async/connection_pool.py
```diff
@@ -294,3 +295,15 @@ async def aclose(self) -> None:
...
             await connection.aclose()
+
+    def get_connection_stats(self) -> Dict[Origin, Dict[str, int]]:
+        """Returns statistics dict around state of the connection pool in the form:
+
+        { origin: { connection_state: number_of_connections_on_that_state } }
+        """
+        stats: Dict[Origin, Dict[str, int]] = {}
+        for origin, connections in self._connections.items():
+            stats[origin] = defaultdict(lambda: 0)
```
JayH5, line 306: Could use `collections.Counter`?

tomchristie:
Okay, great!

So I'd been thinking about this one for a while actually. Something that we *do* want from this is to provide a useful level of structured debugging info. Something that we *don't* want is to introduce API that starts exposing internal implementation details.

I was wondering if perhaps we should shift to something like this...

`{origin: [str, str, str, ...], origin: [...], ...}`

Where the `str` is an under-defined string representation of the state of a connection.

Perhaps also `get_connection_info()` rather than `get_connection_stats()`?

Eg.

```python
{
    (b'https', b'example.org', 443): ["HTTP/2, 3 active streams"],
    (b'https', b'tomchristie.com', 443): ["HTTP/1.1, idle keep-alive", "HTTP/1.1, active", "HTTP/1.1, active"]
}
```

It's more verbose, but it doesn't start leaking implementation details, since we can document the strings as "opaque, don't rely on the format of this" but descriptive strings.

yeraydiazdiaz:
I agree with leaking implementation details, which is why I added not the actual enum class but the name. Formalizing your suggestion strings a bit:

`<HTTP_VERSION>, <STATE>, <ACTIVE_STREAMS_FOR_HTTP2>`

I don't think `keep_alive` is an attribute of the connection though?

Also, do we want to keep the origin as a 3-tuple?

httpcore/_async/connection_pool.py
yeraydiazdiaz: That was my first idea, but couldn't find a clear way of introspecting the state and counting based on that. Didn't think too hard about it though... 😂

yeraydiazdiaz:
Oh, and 👍  on `get_connection_info`.

tomchristie:
> Also, do we want to keep the origin as a 3-tuple?

Erm, good point. Nope, let's have it as a plain string.

> `<HTTP_VERSION>, <STATE>, <ACTIVE_STREAMS_FOR_HTTP2>`

Yup sounds fantastic. While we're connecting we don't yet have an HTTP version, so we've got a "Connecting" case too.

We probably want a `.info() -> str` method on `HTTPConnection` and `HTTP2Connection`, `HTTP11Connection`.

tomchristie:
So, something like this, then?...

```python
{
    "https://example.org:443": ["HTTP/2, ACTIVE, 3 streams"],
    "https://tomchristie.com:443": ["HTTP/1.1, IDLE", "HTTP/1.1, ACTIVE", "HTTP/1.1, ACTIVE"],
    "https://google.com:443": ["Connecting"],
}
```

yeraydiazdiaz:
Yup, sounds good.

httpcore/_async/connection.py
```diff
@@ -39,6 +39,17 @@ def __init__(
...
+    def info(self) -> str:
+        if self.connection is None:
+            return "Not connected"
+        elif self.state == ConnectionState.PENDING:
+            return "Connecting"
+        elif self.is_http11:
+            return f"HTTP/1.1, {self.state.name}"
+        else:
+            connection = cast(AsyncHTTP2Connection, self.connection)
+            return f"HTTP/2, {self.state.name}, {len(connection.streams)} streams"
```
tomchristie, line 51: I'd approach this slightly differently, so...

```python
if self.connection is None:
    return str(self.state.name)
return self.connection.info()
```

And then have `HTTP11Connection` and `HTTP2Connection` implement `.info()`.

I kinda prefer that since it defers the "how do you want to represent the connection info" up to the connection class itself, and doesn't use any casting, which always feels a bit odd.

yeraydiazdiaz: Yup, that makes sense 👍

httpcore/_sync/connection_pool.py
```diff
@@ -294,3 +294,12 @@ def close(self) -> None:
...
             connection.close()
+
+    def get_connection_info(self) -> Dict[str, List[str]]:
+        """Returns dict with ."""
+        stats = {}
+        for origin, connections in self._connections.items():
+            stats[origin_to_url_string(origin)] = [
+                connection.info() for connection in connections
+            ]
+        return stats
```
cansarigol, line 305: Instead of creating a variable, What do you think about the below code? 

```
return {
    origin_to_url_string(origin): [
        connection.info() for connection in connections
     ]
     for origin, connections in self._connections.items()
}
```

yeraydiazdiaz: Personally I find it less readable that way, not a fan of nested comprehensions 🙂

httpcore/_async/connection_pool.py
```diff
@@ -294,3 +294,12 @@ async def aclose(self) -> None:
         # Close all connections
         for connection in connections:
             await connection.aclose()
+
+    def get_connection_info(self) -> Dict[str, List[str]]:
+        """Returns dict with ."""
```
florimondmanca, line 299: Looks like this one didn't get finished, a suggestion based on what I can see here…

```suggestion
        """Returns a dict containing information about the connection of each origin URL."""
```

yeraydiazdiaz: Woops 👍

yeraydiazdiaz:
> Are there any tests we can update to use this instead of private API?

It did in the first iteration of this PR, but since we're using now a string summary I decided to not to assert on the result of the new method to make things clearer.

yeraydiazdiaz:
Thanks for the reviews @florimondmanca and @cansarigol 🙏

tomchristie:
Fantastic stuff! Love it. 👍

> It did in the first iteration of this PR, but since we're using now a string summary I decided to not to assert on the result of the new method to make things clearer.

Actually I do still think we should update the test cases. It's okay for us to document that the specific format shouldn't be relied upon, and *could* be subject to change, while still asserting against it *ourselves* in test cases.

I'd probably suggest...

* Tweak the docstring, to mention that the result should be treated as an opaque string, as the exact formatting and internal behaviours could be subject to change.
* Update our own test cases to assert against it, since we really *do* want to make sure we're tracking if and when those changes occur.

PR #111, opened 2020-07-05
Use isort 5

florimondmanca:
Refs https://github.com/encode/httpx/pull/1050

See https://timothycrosley.github.io/isort/CHANGELOG/#500-penny-july-4-2020

Also pins `isort==5.*`.

I reckon the breaking change here in `isort` is going to require action from most projects that depend on it… Starlette, Uvicorn, Databases, Typesystem and others are probably affected too. Will have to issue fixes for my projects as well…

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,6 +21,6 @@ pytest-cov
 black
 autoflake
 mypy
-isort
+isort==5.*
 mitmproxy
 trustme
--- a/scripts/lint
+++ b/scripts/lint
@@ -9,6 +9,6 @@ export SOURCE_FILES="httpcore tests"
 set -x
 
 ${PREFIX}autoflake --in-place --recursive $SOURCE_FILES
-${PREFIX}isort --project=httpcore --recursive --apply $SOURCE_FILES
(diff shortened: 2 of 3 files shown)


florimondmanca:
Thanks for the review @JayH5 :-)

hukkinj1:
Not sure if this has been considered yet, but one way to avoid having to batch update projects in one go for reasons like this (and to provide a more stable CI experience for external contributors) would be to use poetry and add a lockfile to the repository.

Ofc poetry and the lockfile bring in some of their own complexities, but at least personally I've been a happy user for some time :)

florimondmanca:
I personally think pinning our dependencies is more than enough for us - this kind of problem doesn't happen very often. And any way, you'd also have to take action and batch-upgrade dependencies through poetry to be able to use the latest features (eg isort's new concept of "profiles"). Another point to take into consideration is that switching to anything other than standard venv+pip when managing a dozen of Python packages is bound to bring its own bunch of maintenance overhead (due to additional external magic), which is exactly what I think we're trying to keep at a minimum. (I initially used Poetry for some personal Python package projects, but reverted to venv+pip for this exact reason.)

PR #112, opened 2020-07-07
Improve handling of server disconnects

### LINKED ISSUE #110: Client/Stream doesn't handle CLOSE_WAIT sockets well (meaning other side closed the connection), loops around forever
### Checklist

- [x] The bug is reproducible against the latest release and/or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

When connecting to HTTP2 streams, if the other side closes the connection (causing the socket to go into a `CLOSE_WAIT` state), the Client/Stream doesn't handle that appropriately, instead it keeps reading `b''` from a closed socket forever.

### To reproduce

Install from master:

```
pip install git+https://github.com/encode/httpx.git@bacc2d18350d9f6645828461435fe6e19b9dc518
```

Generate self-signed keys:

```
openssl req -x509 -newkey rsa:2048 -nodes -sha256 -subj  '/CN=localhost'  -keyout key.pem -out cert.pem
```

I'm using `nodejs` version 14.5 to create a small HTTP2 server that streams continuously, called `server.js`.

```javascript
const http2 = require('http2');
const fs = require('fs');

const server = http2.createSecureServer({
  key: fs.readFileSync('key.pem'),
  cert: fs.readFileSync('cert.pem')
});
server.on('error', (err) => console.error(err));

server.on('stream', (stream, headers, flags) => {
	stream.respond({
			':status': 200,
			'content-type': 'text/event-stream'
	});
	setInterval(() => stream.write(`${Math.random()}\n`), 500);
});

server.listen(8443);
```

To install `nodejs`:
```
$ curl -sL https://deb.nodesource.com/setup_14.x | sudo -E bash -
..
$ sudo apt-get install -y nodejs
...
$ nodejs -v
v14.5.0
$
```

A small python client that talks to the server (`client.py`):

```python
import httpx


def get_client():
    timeout = httpx.Timeout(read_timeout=5.0)
    client = httpx.Client(
        base_url="https://localhost:8443",
        http2=True,
        timeout=timeout,
        verify="cert.pem",
    )
    return client


def main():
    with get_client() as c, c.stream("GET", "") as s:
        for i in s.iter_lines():
            print(i)


if __name__ == '__main__':
    main()
```

I start the server:
```
nodejs server.js
```

In another terminal I then connect with the client and I start getting things like:
```
$ python3.8 client.py
0.5366774977298967
                                                                                                                                                                          
0.282442060319521
                                                                                                                                                                          
0.1150984598113638
                                                                                                                                                                          
0.3190482876162004

...
```

I then kill the server:
```
$ nodejs server.js
^C
$
```

### Expected behavior

The client exits somewhat gracefully, perhaps logging a warning that the other side closed the connection.

### Actual behavior

The client just blocks. On inspecting with strace it seems to be continuously reading from an empty buffer:
```
$ sudo strace -p <PID OF client.py>
read(5, "", 5)                          = 0
ioctl(5, FIONBIO, [1])                  = 0
read(5, "", 5)                          = 0
ioctl(5, FIONBIO, [1])                  = 0
read(5, "", 5)                          = 0
ioctl(5, FIONBIO, [1])                  = 0
read(5, "", 5)                          = 0
...
```

### Debugging material

I was able to track the bug to `SyncSocketStream.read`:

```python
def read(self, n: int, timeout: TimeoutDict) -> bytes:
    read_timeout = timeout.get("read")
    exc_map = {socket.timeout: ReadTimeout, socket.error: ReadError}
    
    with self.read_lock:
        with map_exceptions(exc_map):
    	self.sock.settimeout(read_timeout)
    	return self.sock.recv(n)   # returns b''
```

Here `self.sock.recv(n)` returns simply `b''` but that's never handled further up the call stack:

```python
> /home/<username>/.virtualenvs/sandbox3.8/lib/python3.8/site-packages/httpcore/_backends/sync.py(63)read()
-> return self.sock.recv(n)
(Pdb) p self.sock.recv(n)
b''
```

If I keep pressing `n` it comes back around:

```python
(Pdb) n
> /home/<username>/.virtualenvs/sandbox3.8/lib/python3.8/site-packages/httpcore/_backends/sync.py(63)read()
-> return self.sock.recv(n)
(Pdb) n
--Return--
> /home/<username>/.virtualenvs/sandbox3.8/lib/python3.8/site-packages/httpcore/_backends/sync.py(63)read()->b''
-> return self.sock.recv(n)
(Pdb) 
> /home/<username>/.virtualenvs/sandbox3.8/lib/python3.8/site-packages/httpcore/_sync/http2.py(205)receive_events()
-> events = self.h2_state.receive_data(data)
(Pdb) 
> /home/<username>/.virtualenvs/sandbox3.8/lib/python3.8/site-packages/httpcore/_sync/http2.py(206)receive_events()
-> for event in events:
(Pdb) 
> /home/<username>/.virtualenvs/sandbox3.8/lib/python3.8/site-packages/httpcore/_sync/http2.py(216)receive_events()
-> data_to_send = self.h2_state.data_to_send()
(Pdb) 
> /home/<username>/.virtualenvs/sandbox3.8/lib/python3.8/site-packages/httpcore/_sync/http2.py(217)receive_events()
-> self.socket.write(data_to_send, timeout)
(Pdb) 
--Return--
> /home/<username>/.virtualenvs/sandbox3.8/lib/python3.8/site-packages/httpcore/_sync/http2.py(217)receive_events()->None
-> self.socket.write(data_to_send, timeout)
(Pdb) 
> /home/<username>/.virtualenvs/sandbox3.8/lib/python3.8/site-packages/httpcore/_sync/http2.py(196)wait_for_event()
-> while not self.events[stream_id]:
(Pdb) 
> /home/<username>/.virtualenvs/sandbox3.8/lib/python3.8/site-packages/httpcore/_sync/http2.py(197)wait_for_event()
-> self.receive_events(timeout)
(Pdb) 
> /home/<username>/.virtualenvs/sandbox3.8/lib/python3.8/site-packages/httpcore/_backends/sync.py(63)read()
-> return self.sock.recv(n)
```

### Environment

- OS: Ubuntu 18.04.1
- Python version: 3.8.0
- HTTPX version: Pinned at bacc2d18350d9f6645828461435fe6e19b9dc518
- Async environment: asyncio
- HTTP proxy: no
- Custom certificates: yes (but only because I couldn't get nodejs http2 to work without them)

I'm happy to provide more information if that's not sufficient.

florimondmanca:
Fixes #110, cc @nawarnoori

Now, the test setup detailed in https://github.com/encode/httpcore/issues/110#issuecomment-654891958 results in the following error in all cases:

```console
httpcore._exceptions.ReadError: Server disconnected while attempting read
```

### DIFF
--- a/httpcore/_backends/asyncio.py
+++ b/httpcore/_backends/asyncio.py
@@ -131,9 +131,12 @@ class SocketStream(AsyncSocketStream):
         exc_map = {asyncio.TimeoutError: ReadTimeout, OSError: ReadError}
         async with self.read_lock:
             with map_exceptions(exc_map):
-                return await asyncio.wait_for(
+                data = await asyncio.wait_for(
                     self.stream_reader.read(n), timeout.get("read")
                 )
+                if data == b"":
+                    raise OSError("Server disconnected while attempting read")
+                return data
 
     async def write(self, data: bytes, timeout: TimeoutDict) -> None:
         if not data:
--- a/httpcore/_backends/asyncio.py
+++ b/httpcore/_backends/asyncio.py
@@ -135,7 +135,7 @@ class SocketStream(AsyncSocketStream):
                     self.stream_reader.read(n), timeout.get("read")
                 )
                 if data == b"":
-                    raise OSError("Server disconnected while attempting read")
+                    raise ReadError("Server disconnected while attempting read")
                 return data
 
     async def write(self, data: bytes, timeout: TimeoutDict) -> None:
--- a/httpcore/_backends/sync.py
+++ b/httpcore/_backends/sync.py
@@ -54,12 +54,15 @@ class SyncSocketStream:
 
     def read(self, n: int, timeout: TimeoutDict) -> bytes:
         read_timeout = timeout.get("read")
-        exc_map = {socket.timeout: ReadTimeout, socket.error: ReadError}
+        exc_map = {socket.timeout: ReadTimeout, OSError: ReadError}
 
         with self.read_lock:
             with map_exceptions(exc_map):
                 self.sock.settimeout(read_timeout)
-                return self.sock.recv(n)
+                data = self.sock.recv(n)
+                if data == b"":
+                    raise OSError("Server disconnected while attempting read")
+                return data
 
     def write(self, data: bytes, timeout: TimeoutDict) -> None:
         write_timeout = timeout.get("write")
--- a/httpcore/_backends/sync.py
+++ b/httpcore/_backends/sync.py
@@ -54,14 +54,14 @@ class SyncSocketStream:
 
     def read(self, n: int, timeout: TimeoutDict) -> bytes:
         read_timeout = timeout.get("read")
-        exc_map = {socket.timeout: ReadTimeout, OSError: ReadError}
+        exc_map = {socket.timeout: ReadTimeout, socket.error: ReadError}
 
         with self.read_lock:
             with map_exceptions(exc_map):
                 self.sock.settimeout(read_timeout)
                 data = self.sock.recv(n)
                 if data == b"":
-                    raise OSError("Server disconnected while attempting read")
+                    raise ReadError("Server disconnected while attempting read")
                 return data
 
     def write(self, data: bytes, timeout: TimeoutDict) -> None:
--- a/httpcore/_backends/trio.py
+++ b/httpcore/_backends/trio.py
@@ -55,12 +55,21 @@ class SocketStream(AsyncSocketStream):
 
     async def read(self, n: int, timeout: TimeoutDict) -> bytes:
         read_timeout = none_as_inf(timeout.get("read"))
-        exc_map = {trio.TooSlowError: ReadTimeout, trio.BrokenResourceError: ReadError}
+        exc_map = {trio.TooSlowError: ReadTimeout, OSError: ReadError}
 
         async with self.read_lock:
             with map_exceptions(exc_map):
                 with trio.fail_after(read_timeout):
-                    return await self.stream.receive_some(max_bytes=n)
+                    try:
+                        data = await self.stream.receive_some(max_bytes=n)
+                    except trio.BrokenResourceError:
+                        message = "Server disconnected while attempting read"
+                        raise OSError(message) from None
+
+                    if data == b"":
+                        raise OSError("Server disconnected while attempting read")
+
+                    return data
 
     async def write(self, data: bytes, timeout: TimeoutDict) -> None:
         if not data:
--- a/httpcore/_backends/trio.py
+++ b/httpcore/_backends/trio.py
@@ -55,20 +55,14 @@ class SocketStream(AsyncSocketStream):
 
     async def read(self, n: int, timeout: TimeoutDict) -> bytes:
         read_timeout = none_as_inf(timeout.get("read"))
-        exc_map = {trio.TooSlowError: ReadTimeout, OSError: ReadError}
+        exc_map = {trio.TooSlowError: ReadTimeout, trio.BrokenResourceError: ReadError}
 
         async with self.read_lock:
             with map_exceptions(exc_map):
                 with trio.fail_after(read_timeout):
-                    try:
-                        data = await self.stream.receive_some(max_bytes=n)
-                    except trio.BrokenResourceError:
-                        message = "Server disconnected while attempting read"
-                        raise OSError(message) from None
-
+                    data = await self.stream.receive_some(max_bytes=n)
                     if data == b"":
-                        raise OSError("Server disconnected while attempting read")
-
+                        raise ReadError("Server disconnected while attempting read")
                     return data
 
     async def write(self, data: bytes, timeout: TimeoutDict) -> None:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,7 +5,6 @@ import typing
 
 import pytest
 import trustme
-
 from mitmproxy import options, proxy
 from mitmproxy.tools.dump import DumpMaster
 


httpcore/_backends/trio.py
```diff
@@ -55,12 +55,21 @@ async def start_tls(
...
                 with trio.fail_after(read_timeout):
-                    return await self.stream.receive_some(max_bytes=n)
+                    try:
+                        data = await self.stream.receive_some(max_bytes=n)
+                    except trio.BrokenResourceError:
+                        message = "Server disconnected while attempting read"
+                        raise OSError(message) from None
+
+                    if data == b"":
+                        raise OSError("Server disconnected while attempting read")
```
florimondmanca, line 70: Interestingly, `trio` raises an exception in the HTTP/2 case and returns `b""` in the HTTP/1.1 case (causing a `ProtocolError` from `h11`). I'm not sure why, because in both cases regular TCP sockets are used and we attempt reading 4096 bytes. So the solution I found is to deal with both possibilities…

httpcore/_backends/sync.py
```diff
@@ -54,12 +54,15 @@ def start_tls(
 
     def read(self, n: int, timeout: TimeoutDict) -> bytes:
         read_timeout = timeout.get("read")
-        exc_map = {socket.timeout: ReadTimeout, socket.error: ReadError}
+        exc_map = {socket.timeout: ReadTimeout, OSError: ReadError}
```
florimondmanca, line 57: `socket.error` is a synonym for `OSError` in Python 3.

httpcore/_backends/trio.py
nawarnoori, line 70: very minor: you could reuse `message` here.

florimondmanca:
> do you guys add tests? 

Yes, ideally we'd have a test for this. I did try something - spinning up a Uvicorn server and restart it while we're reading from the stream (which should trigger the error case). It feels pretty heavy just for this particular edge case. But I don't know how we'd go with mocking either, so…

nawarnoori:
Are you happy to merge this? The fix looks good, even without tests it's better than not being in.

florimondmanca:
I'm tempted to be okay with merging this in to issue a bug fix release too, but I'd like to hear thoughts from @encode/maintainers about this. :-)

httpcore/_backends/trio.py
```diff
@@ -55,12 +55,21 @@ async def start_tls(
...
+        exc_map = {trio.TooSlowError: ReadTimeout, OSError: ReadError}
 
         async with self.read_lock:
             with map_exceptions(exc_map):
                 with trio.fail_after(read_timeout):
-                    return await self.stream.receive_some(max_bytes=n)
+                    try:
+                        data = await self.stream.receive_some(max_bytes=n)
+                    except trio.BrokenResourceError:
+                        message = "Server disconnected while attempting read"
```
JayH5, line 66: Not clear why this is in a variable?

florimondmanca:
@JayH5 There's [`ConnectionAbortedError`](https://docs.python.org/3/library/exceptions.html#ConnectionAbortedError) (a subclass of `OSError`) - I was hesitant to use it because we're mapping from `OSError` (so `ReadError` is raised in all cases) and it's not immediate knowledge that `ConnectionAbortedError` will be raised too.

So hmm, actually I think I could just raise an `httpcore.ReadError` directly. :-)

httpcore/_backends/sync.py
```diff
@@ -54,12 +54,15 @@ def start_tls(
...
 
         with self.read_lock:
             with map_exceptions(exc_map):
                 self.sock.settimeout(read_timeout)
-                return self.sock.recv(n)
+                data = self.sock.recv(n)
+                if data == b"":
+                    raise OSError("Server disconnected while attempting read")
```
nawarnoori, line 64: Just a quick thought, will `self.sock` get closed elsewhere, e.g. in `__exit__` or when explicitly closing the client? I guess it's something you want to leave up to the client whether to close the socket in these cases.

florimondmanca: Yes, closing the socket is already handled by the connection pool, which closes all connections (which themselves close their sockets) on `__aexit__`.

nawarnoori:
Hi chaps, is there something I could do to help this along? I've patched the bug locally with a temporary hack but would really benefit from having this in the library.

florimondmanca:
@nawarnoori I just need to make a new pass on this to raise a `ReadError` exceptions directly instead of `OSError`, and then this should be good for a final review… Unfortunately I don't think outside contributors can push to this branch, but if you'd like to speed things up feel free to bundle these ideas into a new PR.

nawarnoori:
@florimondmanca I'm getting 403 on either pushing to this branch or pushing a new branch, but I've formatted the commit as a patch that I've attached here (had to add `.txt` to get github to upload it). It's very small, I've removed some exception mappings that I thought were unnecessary (since I couldn't see what would raise those exceptions) and just raise `ReadError` directly.

[0001-Just-raise-ReadError-directly.patch.txt](https://github.com/encode/httpcore/files/4911877/0001-Just-raise-ReadError-directly.patch.txt)

httpcore/_backends/asyncio.py
```diff
@@ -131,9 +131,12 @@ async def read(self, n: int, timeout: TimeoutDict) -> bytes:
         exc_map = {asyncio.TimeoutError: ReadTimeout, OSError: ReadError}
         async with self.read_lock:
             with map_exceptions(exc_map):
-                return await asyncio.wait_for(
+                data = await asyncio.wait_for(
                     self.stream_reader.read(n), timeout.get("read")
                 )
+                if data == b"":
+                    raise OSError("Server disconnected while attempting read")
```
tomchristie, line 138:
```suggestion
                    raise ReadError("Server disconnected while attempting read")
```

httpcore/_backends/sync.py
tomchristie, line 64:
```suggestion
                    raise ReadError("Server disconnected while attempting read")
```

httpcore/_backends/trio.py
```diff
@@ -55,12 +55,21 @@ async def start_tls(
...
+                        raise OSError(message) from None
```
tomchristie, line 67:
```suggestion
                        raise ReadError(message) from None
```

httpcore/_backends/trio.py
tomchristie, line 70:
```suggestion
                        raise ReadError("Server disconnected while attempting read")
```

httpcore/_backends/asyncio.py
j178:
```suggestion
                    raise OSError("Server disconnected while attempting read")
```

florimondmanca:
> Would be nice to test

For the record, to test this as part of the test suite we would _need_ these:

- Spin up an HTTP/1.1 server locally (the bug was _also_ reproducible on HTTP/2 but it's not an absolute requirement - it also occurred on HTTP/1.1).
- Programmatically close the client connection on the server during the test (eg restart the server).
- Synchronize all of this (eg server in one thread vs tests in the main thread).

We have / used to have something similar in HTTPX for connection pooling tests, but since at this point the test suite is very minimal (integration tests against a live website), we'd need some more work to be able to automatically test this properly.

Any case, I did some manual testing and my claim in the PR description holds: we get a proper `ReadError` in all cases instead of undefined behavior (eg loop forever on `sock.recv()`).

Shall we :shipit:?

florimondmanca:
Thanks all! 🎉

PR #114, opened 2020-07-08
Prettify unasync output

florimondmanca:
Mostly a developer quality of life improvement — I figured it was easy enough to tweak our `lint` script so that it runs code formatters on the unasynced code, so that when navigating the sync code we can save things without the IDE reformatting everything.

The important bit here is the modification to `scripts/lint`: https://github.com/encode/httpcore/pull/114/files#diff-6694ef873aea668ceffff73715edc0fd
The rest comes from running `scripts/lint`.

### DIFF
--- a/httpcore/_sync/base.py
+++ b/httpcore/_sync/base.py
@@ -1,6 +1,6 @@
 import enum
 from types import TracebackType
-from typing import Iterator, Callable, List, Tuple, Type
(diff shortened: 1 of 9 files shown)


florimondmanca:
Yeah, okay, obviously black, isort & others would then modify the unasync'd output and unasync's `--check` would fail because it's not aware of those changes.

PR #115, opened 2020-07-08
Add comment about why unasync is run after code formatters

florimondmanca:
Prompted by https://github.com/encode/httpcore/pull/114#issuecomment-655509240

A note to my future self that might be useful for other contributors. 😄

### DIFF
--- a/scripts/lint
+++ b/scripts/lint
@@ -11,4 +11,7 @@ set -x
 ${PREFIX}autoflake --in-place --recursive $SOURCE_FILES
 ${PREFIX}isort --project=httpcore $SOURCE_FILES
 ${PREFIX}black --exclude 'httpcore/_sync/.*' --exclude 'tests/sync_tests/.*' --target-version=py36 $SOURCE_FILES
+
+# Run unasync last because its `--check` mode is not aware of code formatters.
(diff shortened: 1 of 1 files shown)


florimondmanca:
There's line length-sensitive formatting, but also a couple of other things due to how we sometimes remove lines during unasync.

Examples:

- Too many blank lines:

https://github.com/encode/httpcore/blob/337aff9eef9f3ee4c121857967758917f15181cb/tests/sync_tests/test_interfaces.py#L84-L88

- Emtpy line between decorators (I'm surprised this is valid Python!):

https://github.com/encode/httpcore/blob/337aff9eef9f3ee4c121857967758917f15181cb/tests/sync_tests/test_interfaces.py#L207-L211

`isort` would also sometimes apply changes (because `Async` and `Sync` don't typically end up in the same position), but that is fine.

Anyway, my initial motivation for #114 was really to just be able to go through the code, make some hacky changes for testing, save and… not have my IDE black-ify everything. :-) But I reckon there are probably ways I can configure my editor to not do that in `_sync/` and `sync_tests`, so…

tomchristie:
Gotcha 👍👍

PR #117, opened 2020-07-16
close connection on exception

### LINKED ISSUE #116: connection leak in async connection pool
in AsyncConnectionPool.request

```python
            except Exception:
                logger.trace("remove from pool connection=%r", connection)
                await self._remove_from_pool(connection)
                raise
```

the connection is not closed.

hzliu:
fix #116


tomchristie:
Thanks for this!
Looks like it might be valid, but will need to dig into it more closely, and make sure we've got a test case to properly cover it.

elupus:
Just did the exact same fix locally to fix an issue. This leaks session on server that for example close the socket before returning data, in my case eventually depleting sockets in the server. I'm not entirely sure it's the only issue at play. But it's one for sure in the below linked issues.

https://github.com/home-assistant/core/issues/57093
https://github.com/home-assistant/core/issues/49801#issuecomment-931645725

PR #120, opened 2020-07-18
Tweak TimeoutDict type variance

florimondmanca:
Fixes a mypy error when `timeout` is defined as a separate variable:

```python
import httpcore

timeout = {"pool": 5.0}

with httpcore.SyncConnectionPool() as http:
    _ = http.request(b"GET", (b"https", b"example.com", 443, b"/"), timeout=timeout)
```

```console
debug/test.py:6: error: Argument "timeout" to "request" of "SyncHTTPTransport" has incompatible type "Dict[str, float]"; expected "Optional[Dict[str, Optional[float]]]"
debug/test.py:6: note: "Dict" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance
debug/test.py:6: note: Consider using "Mapping" instead, which is covariant in the value type
Found 1 error in 1 file (checked 1 source file)
```

We've seen instances of this several times in the past for HTTPCore and HTTPX, it's a common gotcha…

### DIFF
--- a/httpcore/_types.py
+++ b/httpcore/_types.py
@@ -2,10 +2,10 @@
 Type definitions for type checking purposes.
 """
 
-from typing import Dict, List, Optional, Tuple, Union
+from typing import List, Mapping, Optional, Tuple, Union
 
 StrOrBytes = Union[str, bytes]
 Origin = Tuple[bytes, bytes, int]
 URL = Tuple[bytes, bytes, Optional[int], bytes]
 Headers = List[Tuple[bytes, bytes]]
-TimeoutDict = Dict[str, Optional[float]]
+TimeoutDict = Mapping[str, Optional[float]]


httpcore/_types.py
```diff
@@ -2,10 +2,10 @@
 Type definitions for type checking purposes.
 """
 
-from typing import Dict, List, Optional, Tuple, Union
+from typing import List, Mapping, Optional, Tuple, Union
 
 StrOrBytes = Union[str, bytes]
 Origin = Tuple[bytes, bytes, int]
 URL = Tuple[bytes, bytes, Optional[int], bytes]
 Headers = List[Tuple[bytes, bytes]]
```
JayH5, line 10: I see this is a `List` and not a `Sequence` but the value type doesn't vary so that's ok. Still, it might be nice to have some standardisation on when to use `List` vs `Sequence`/`Dict` vs `Mapping` instead of trying to fix these every time somebody's mypy complains.

florimondmanca: I guess the standard is what you've just mentioned - invariant type is OK if inner types don't vary, but covariant is required when they do.

An option to catch these more early on would be to systematically define these arguments as separate variables first in tests, since this is how these MyPy errors fire off usually... It's a bit tedious but we could see what a PR for this looks like...

tomchristie:
Urgh. This has a knock on effect to httpx, which would need the same changes too.

TBH. I'm having a hard time figuring why `Dict[str, Optional[float]]` would raise typing errors for...

```python
import httpcore

timeout = {"pool": 5.0}

with httpcore.SyncConnectionPool() as http:
    _ = http.request(b"GET", (b"https", b"example.com", 443, b"/"), timeout=timeout)
```

Or how being forced by our type checker to use a more general type here than we *actually* want is an improvement. 🙃

I guess I need to suck it up and just switch to `Mapping` in `httpx` too.

PR #121, opened 2020-07-21
Optional HTTP/2

tomchristie:
For consideration.
We might want to think again about having HTTP/2 be an optional install for our `httpx` 1.0 release onwards.

Done:

* Provide a base HTTP Connection class interface, that is agnostic of HTTP/1.1 and HTTP/2.
* Use the base HTTP Connection class for the `.connection` signature in `AsyncHTTPConnection`.
* Lazily import either `AsyncHTTP11Connection` or `AsyncHTTP2Connection` as required.

To do:

* Import `h2` in `AsyncConnectionPool` if `http2=True` is set, so that we get nice early errors if it is not installed.
* Move `h2` to an optional install, requiring `pip install httpcore[http2]`

### DIFF
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -10,7 +10,7 @@ from .base import (
     ConnectionState,
     NewConnectionRequired,
 )
-from .http import BaseHTTPConnection
+from .http import AsyncBaseHTTPConnection
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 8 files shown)


tomchristie:
@florimondmanca - I think we ought to look at getting this one in, and treat the remaining issues as a follow up...

* Import h2 in AsyncConnectionPool if http2=True is set, so that we get nice early errors if it is not installed.
* Move h2 to an optional install, requiring pip install httpcore[http2]

That'll get us in good shape and allow us to make a judgement call on if we want optional HTTP/2 for `httpx` 0.14 or if we're waiting for 1.0.

florimondmanca:
@tomchristie Okay, I thought you meant to bring the remaining TODOs in this PR which is why I held back on reviewing. Will review soon then. :-)

tomchristie:
Right, let's do this...

PR #127, opened 2020-08-02
Support eg. `AsyncByteStream(content=b'...')`

### LINKED ISSUE #125: Add `content=b""` to base stream implementations.
From https://github.com/encode/httpx/pull/1096

The `httpcore.SyncByteStream` and `httpcore.AsyncByteStream` classes provide a simple base implementation that allows users to pass `iterator`/`close` or `aiterator`/`aclose` parameters. That's actually a bit fiddly if you've just got a non-iterating bytestring that you want to pass. We ought to also support passing plain bytes to the base implementation. So...

```python
    """
    The base interface for request and response bodies.
    Concrete implementations should subclass this class, and implement
    the `\\__aiter__` method, and optionally the `close` method.
    """

    def __init__(
        self, content: Union[bytes, AsyncIterator[bytes]] = b"", aclose_func: Callable = None,
    ) -> None:
        self.content = content
        self.aclose_func = aclose_func

    async def __aiter__(self) -> AsyncIterator[bytes]:
        """
        Yield bytes representing the request or response body.
        """
        if isinstance(self.content, bytes):
            yield self.content
        else:
            async for chunk in self.content:
                yield chunk

        """
        Must be called by the client to indicate that the stream has been closed.
        """
        if self.aclose_func is not None:
            await self.aclose_func()
```

tomchristie:
Nicer base class implementations for `AsyncByteStream` and `SyncByteStream`.

Allows for eg `AsyncByteStream(content=b'...')` usages.

Closes #125

### DIFF
--- a/httpcore/_async/base.py
+++ b/httpcore/_async/base.py
@@ -5,10 +5,6 @@ from typing import AsyncIterator, Callable, List, Tuple, Type
 from .._types import URL, Headers, TimeoutDict
 
 
-async def empty() -> AsyncIterator:
-    yield b""
-
-
 class NewConnectionRequired(Exception):
     pass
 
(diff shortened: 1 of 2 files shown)


florimondmanca:
Any docs changes required here too…?

tomchristie:
@florimondmanca - Nope, should all be picked up by our autodocs. We've not got *great* documentation around using the interface, but addressing that more fully is something for a different day! :)

PR #128, opened 2020-08-02
Add an `UnsupportedProtocol` exception

### LINKED ISSUE #124: Introduce `UnsupportedProtocol` exception.
To be raised for non `http`/`https` requests.

Note that this will allow use to drop `enforce_http_url` in `httpx`, and instead defer that handling entirely to the transport class.

tomchristie:
Closes #124

Once we tie this exception through to `httpx`, this will allow us to drop `enforce_http_url` in `httpx`, and instead defer the decision entirely to the transport layer.

Will later allow us to support mounting transports on schemes other than `http`, `https`.

### DIFF
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -126,7 +126,7 @@ class AsyncConnectionPool(AsyncHTTPTransport):
     ) -> Tuple[bytes, int, bytes, Headers, AsyncByteStream]:
         if url[0] not in (b"http", b"https"):
             scheme = url[0].decode("latin-1")
-            raise UnsupportedProtocol("Unsupported URL protocol '{scheme}'")
+            raise UnsupportedProtocol(f"Unsupported URL protocol {scheme!r}")
 
         origin = url_to_origin(url)
 
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -2,7 +2,7 @@ from ssl import SSLContext
 from typing import AsyncIterator, Callable, Dict, List, Optional, Set, Tuple
 
 from .._backends.auto import AsyncLock, AsyncSemaphore, AutoBackend
-from .._exceptions import PoolTimeout
+from .._exceptions import PoolTimeout, UnsupportedProtocol
 from .._threadlock import ThreadLock
 from .._types import URL, Headers, Origin, TimeoutDict
 from .._utils import get_logger, origin_to_url_string, url_to_origin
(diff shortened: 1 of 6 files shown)


httpcore/_async/connection_pool.py
```diff
@@ -124,7 +124,10 @@ async def request(
         stream: AsyncByteStream = None,
         timeout: TimeoutDict = None,
     ) -> Tuple[bytes, int, bytes, Headers, AsyncByteStream]:
-        assert url[0] in (b"http", b"https")
+        if url[0] not in (b"http", b"https"):
+            scheme = url[0].decode("latin-1")
+            raise UnsupportedProtocol("Unsupported URL protocol '{scheme}'")
```
florimondmanca, line 129: Missing `f` for this f-string :-) Also using `!r` to automatically show the `'` from the string repr…

```suggestion
            raise UnsupportedProtocol(f"Unsupported URL protocol {scheme!r}")
```

tests/async_tests/test_interfaces.py
```diff
@@ -50,6 +50,18 @@ async def test_https_request() -> None:
         assert len(http._connections[url[:3]]) == 1  # type: ignore
 
 
+@pytest.mark.usefixtures("async_environment")
+async def test_ftp_request() -> None:
```
florimondmanca, line 54: Don't think testing our behavior when requesting `ftp` is the primary goal of this test, so suggesting an alternative name?

```suggestion
async def test_request_unsupported_protocol() -> None:
```

tests/async_tests/test_interfaces.py
```diff
@@ -50,6 +50,18 @@ async def test_https_request() -> None:
...
+    async with httpcore.AsyncConnectionPool() as http:
+        method = b"GET"
+        url = (b"ftp", b"example.org", 443, b"/")
+        headers = [(b"host", b"example.org")]
+        with pytest.raises(httpcore.UnsupportedProtocol):
+            http_version, status_code, reason, headers, stream = await http.request(
+                method, url, headers
+            )
```
florimondmanca, line 62: (Nit) I know we don't have `flake8` on this repo anymore, but mine wouldn't be happy with these unused variables (and I'd agree…), so maybe…?

```suggestion
            await http.request(method, url, headers)
```

Don't remember - what was the rationale again for removing `flake8`? I have the intuition we could bring it back with looser configuration (eg omit annoying errors such as "lines too long").

tomchristie: Not sure, but yes we might be able to look at it again.

PR #130, opened 2020-08-03
Finalize optional HTTP/2

tomchristie:
We'd need to follow up on this with either...

* Making http/2 optional in `httpx` 0.14
* Using `httpcore[http2]` in the `httpx` 0.14 setup.py info.

Of those two I think we should *probably?* push on and make the change with `httpx` 0.14, since it's where we're headed anyway.

### DIFF
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -96,6 +96,15 @@ class AsyncConnectionPool(AsyncHTTPTransport):
         self._backend = AutoBackend()
         self._next_keepalive_check = 0.0
 
+        if http2:
+            try:
+                import h2
+            except ImportError:
+                raise ImportError(
+                    "Attempted to use http2=True, but the 'h2' "
+                    "package is not installed. Use 'pip install httpcore[http2]'."
+                )
+
     @property
     def _connection_semaphore(self) -> AsyncSemaphore:
         # We do this lazily, to make sure backend autodetection always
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -53,7 +53,10 @@ def get_packages(package):
     packages=get_packages("httpcore"),
     include_package_data=True,
     zip_safe=False,
-    install_requires=["h11>=0.8,<0.10", "h2==3.*", "sniffio==1.*"],
+    install_requires=["h11>=0.8,<0.10", "sniffio==1.*"],
+    extras_require={
+        "http2": ["h2==3.*",]
```
florimondmanca, line 58: (Nit)

```suggestion
        "http2": ["h2==3.*"],
```

httpcore/_async/connection_pool.py
```diff
@@ -96,6 +96,15 @@ def __init__(
...
         self._next_keepalive_check = 0.0
 
+        if http2:
+            try:
+                import h2
+            except ImportError:
+                raise ImportError(
+                    "Attempted to use http2=True, but the 'h2' "
+                    "package is not installed. Use 'pip install httpcore[http2]'."
+                )
```
florimondmanca, line 106: I assume we'll do the same dance in HTTPX when a client is given `http2=True`, such that an appropriate `ImportError` hinting at `httpx[http2]` is raised before this code is even reached?

tomchristie: Yes I think so.

PR #132, opened 2020-08-03
No typing on bytestream attributes

tomchristie:
We need to do a bit of a dance on `AsyncByteStream`/`SyncByteStream`, since we're using them *both* as an interface class, and include a plain base implementation for simple usages.

By using `setattr`/`getattr` instead of typical attribute setting we can avoid type checkers making assumptions about how what types base classes should expect on attributes they might use, like `.content`.

### DIFF
--- a/httpcore/_async/base.py
+++ b/httpcore/_async/base.py
@@ -46,27 +46,39 @@ class AsyncByteStream:
         aiterator: AsyncIterator[bytes] = None,
         aclose_func: Callable = None,
     ) -> None:
+        #  We're doing a bit of a funny dance here to ensure that type checkers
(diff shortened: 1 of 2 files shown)


florimondmanca:
Hmm, not sure about this.

This *could* be a smell that we should really separate the interface from the handy basic implementation.

But at the same time I don't know how I'd call things if they had to be split apart, so...

What's your feeling about this @tomchristie?

tomchristie:
That's a great observation. We could keep the interfaces pure, and provide `SimpleByteStream`, `IteratorByteStream`, and `AiteratorByteStream` concrete implementations.

tomchristie:
Closing in favour of #133

PR #133, opened 2020-08-05
Pure ByteStream interface

tomchristie:
This pull request ensures that our `AsyncByteStream` and `SyncByteStream` classes are pure interface classes, without any implementations themselves.

It also introduces the following concrete implementations, which act as helpful classes for developers, so that it's not always strictly necessary to implement a concrete class.

```python
stream = httpcore.PlainByteStream(b"123")
stream = httpcore.IteratorByteStream(iterator=...)  # Optionally, also `close_func=...`.
stream = httpcore.AsyncIteratorByteStream(aiterator=...)  # Optionally, also `aclose_func=...`.
```

Developers don't *need* to use our concrete implementations, as they could instead provide their own implementations, but this change gives us both a nice purity of interface, and the convenience of providing a few very basic implementations.

### DIFF
--- a/httpcore/_bytestreams.py
+++ b/httpcore/_bytestreams.py
@@ -3,13 +3,13 @@ from ._async.base import AsyncByteStream
 from ._sync.base import SyncByteStream
 
 
-class SimpleByteStream(AsyncByteStream, SyncByteStream):
+class PlainByteStream(AsyncByteStream, SyncByteStream):
     """
     A concrete implementation for either sync or async byte streams.
     Just handles a plain byte string as the content of the stream.
 
     ```
-    stream = httpcore.SimpleByteStream(b"123")
+    stream = httpcore.PlainByteStream(b"123")
     ```
     """
 
--- a/httpcore/_bytestreams.py
+++ b/httpcore/_bytestreams.py
@@ -49,7 +49,7 @@ class IteratorByteStream(SyncByteStream):
             self._close_func()
 
 
-class AIteratorByteStream(AsyncByteStream):
+class AsyncIteratorByteStream(AsyncByteStream):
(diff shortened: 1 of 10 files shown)


httpcore/_bytestreams.py
```diff
@@ -0,0 +1,77 @@
+from typing import AsyncIterator, Iterator, Callable
+from ._async.base import AsyncByteStream
+from ._sync.base import SyncByteStream
+
+
+class SimpleByteStream(AsyncByteStream, SyncByteStream):
```
florimondmanca, line 6: What do we think of…?

```suggestion
class PlainByteStream(AsyncByteStream, SyncByteStream):
```

Or maybe just…?

```suggestion
class ByteStream(AsyncByteStream, SyncByteStream):
```

httpcore/_bytestreams.py
```diff
@@ -0,0 +1,77 @@
...
+    def __iter__(self) -> Iterator[bytes]:
+        for chunk in self._iterator:
+            yield chunk
+
+    def close(self) -> None:
+        if self._close_func is not None:
+            self._close_func()
+
+
+class AIteratorByteStream(AsyncByteStream):
```
florimondmanca, line 52: Longer, but for consistency with our naming elsewhere…?

```suggestion
class AsyncIteratorByteStream(AsyncByteStream):
```

tomchristie: Absolutely yup - great call!

httpcore/_bytestreams.py
tomchristie: Tricky one, but a fair call yup. I've gone with `PlainByteStream`.

PR #134, opened 2020-08-05
Local address support

tomchristie:
Based on #100 with a few extra minor changes.

* Use `local_address` as the keyword argument name. (*If in doubt, follow the Trio API.*)
* Use `str` as the argument type. (Granted we're using `bytes` throughtout in our `request()` API, but I'm not sure I see the point in that here - all the underlying backends either expect or allow the address as a string. *If in doubt, follow the Trio API.*)
* Support `local_address=...` on Trio. (currently only usable against trio master branch)
* Pass with keyword-only argument usage internally to `open_tcp_stream()`


httpcore/_backends/asyncio.py
```diff
@@ -225,13 +225,19 @@ async def open_tcp_stream(
...
         ssl_context: Optional[SSLContext],
         timeout: TimeoutDict,
+        *,
+        local_address: Optional[str],
     ) -> SocketStream:
         host = hostname.decode("ascii")
         connect_timeout = timeout.get("connect")
         exc_map = {asyncio.TimeoutError: ConnectTimeout, OSError: ConnectError}
         with map_exceptions(exc_map):
+            local_addrport = None if local_address is None else (local_address, 0)
```
florimondmanca, line 235: (Nit) Consider moving this line alongside computation of `host` and `connect_timeout`?

tests/async_tests/test_interfaces.py
```diff
@@ -212,6 +212,25 @@ async def test_http_proxy(proxy_server: URL, proxy_mode: str) -> None:
         assert reason == b"OK"
 
 
+@pytest.mark.parametrize("local_address", [b"0.0.0.0"])
+@pytest.mark.asyncio
+# This doesn't run with trio, since trio doesn't support local_address.
+async def test_http_request_local_address(local_address: str) -> None:
+    async with httpcore.AsyncConnectionPool(local_address=local_address) as http:
```
florimondmanca, line 219: Possible simplification?

```suggestion
@pytest.mark.asyncio
# This doesn't run with trio, since trio doesn't support local_address.
async def test_http_request_local_address() -> None:
    async with httpcore.AsyncConnectionPool(local_address="0.0.0.0") as http:
```

tests/async_tests/test_interfaces.py
```diff
@@ -212,6 +212,25 @@ async def test_http_proxy(proxy_server: URL, proxy_mode: str) -> None:
...
+        headers = [(b"host", b"example.org")]
+        http_version, status_code, reason, headers, stream = await http.request(
+            method, url, headers
+        )
+        body = await read_body(stream)
+
+        assert http_version == b"HTTP/1.1"
+        assert status_code == 200
+        assert reason == b"OK"
+        assert len(http._connections[url[:3]]) == 1  # type: ignore
```
florimondmanca, line 231: It doesn't look like this test tests that the `local_address` actually gets used — I suppose that's intended for simplicity since testing that w/o mocking is not trivial?

tomchristie:
@florimondmanca Fantastic review points as always - thank you!

PR #139, opened 2020-08-06
Add Unix Domain Socket support

### LINKED ISSUE #138: UDS Support
This httpx chanelog [PR](https://github.com/encode/httpx/pull/971), as part of a movement to sunset usage of urllib3 in favor of httpcore, mentions UDS support as a temporary casualty of the process because "maintaining that would have meant having to push back our work towards a 1.0 release".

Regarding putting this support back into httpcore, there has been recent work done in this [proof of concept](https://gist.github.com/florimondmanca/110133096a4d025eaea2dd08d9bcc380) (thanks @florimondmanca ) that suggests that including UDS support inside the library would not be an overwhelming task.

I personally use a lot of inter-service communication via unix sockets so this would be (opinion) a welcome addition as a first-class citizen of the new release.

I am brand new to this library; I have only used pre-httpcore httpx. After upgrading past 0.12 of httpx, I was surprised that my code could no longer use the uds= keyword when creating clients, enough to [blow up](https://gitter.im/python-trio/general?at=5f288fa46334d26a7e71a57b) on the trio gitter (apologies). I now understand that keeping to a release schedule and making everyone happy is an extremely hard task!

@tomchristie suggested this issue be created to start a discussion here. Go!

florimondmanca:
"For consideration" :-)

Closes #138 

Based on work in https://github.com/encode/httpx/pull/511

I think it turns out the best option is really just to add `uds=...` support like for `local_address`. UDS is just a different network transport layer than TCP, but all of HTTP still applies, and so we can reuse our connection pool, HTTPS support, HTTP/2 support, etc.

It's a bit disappointing because it'd be nice it this was easy to create as a separate transport, but a/ reusing the connection pool while customizing how the socket is established is pretty much impossible, and b/ re-implementing everything just to change a few lines in `AsyncConnectionPool` and `AsyncHTTPConnection` seems overkill (and if I had to implement this as a third-party package, I'd find it pretty daunting).


tomchristie:
Will need to take a comprehensive review over it, but yeah I'm in favour of this.

tests/async_tests/test_interfaces.py
```diff
@@ -304,3 +306,25 @@ async def test_connection_pool_get_connection_info(
...
+        url = (b"http", b"localhost", None, b"/")
+        headers = [(b"host", b"localhost")]
+        http_version, status_code, reason, headers, stream = await http.request(
+            method, url, headers
+        )
+        assert http_version == b"HTTP/1.1"
+        assert status_code == 200
+        assert reason == b"OK"
+        body = await read_body(stream)
+        assert body == b"Hello, world!"
```
florimondmanca, line 330:
```suggestion
        assert body == b"Hello, world!"

```

lundberg:
Awesome ❤️

PR #140, opened 2020-08-07
`max_keepalive` -> `max_keepalive_connections`

tomchristie:


tests/async_tests/test_interfaces.py
```diff
@@ -195,12 +195,8 @@ async def test_http_proxy(proxy_server: URL, proxy_mode: str) -> None:
     url = (b"http", b"example.org", 80, b"/")
     headers = [(b"host", b"example.org")]
     max_connections = 1
-    max_keepalive = 2
     async with httpcore.AsyncHTTPProxy(
-        proxy_server,
-        proxy_mode=proxy_mode,
-        max_connections=max_connections,
-        max_keepalive=max_keepalive,
+        proxy_server, proxy_mode=proxy_mode, max_connections=max_connections,
```
florimondmanca, line 199: Hmm, so are we not testing `max_keepalive_connections` elsewhere still? (This looks like you initially meant to just rename the argument.)

tomchristie: Dropped `max_keepalive` here, since it *wasn't* actually testing anything, as setting `max_keepalive` >= `max_connections` will end up having no effect. And yup it'd be good to have a *meaningful* `max_keepalive_connections` test.

PR #143, opened 2020-08-08
Run scripts/lint

florimondmanca:
Hasn't been run for a while it seems, and `scripts/check` doesn't check for black or isort so we wouldn't have known…

Will submit a follow-up PR to add `black --check` and `isort --check`.


tomchristie:
@florimondmanca @JayH5 Lovely stuff! I'll leave this in your hands. 😊

PR #144, opened 2020-08-08
Add isort and black to scripts/check

florimondmanca:
Requires #143 for these new checks to pass.

### DIFF
--- a/scripts/lint
+++ b/scripts/lint
@@ -9,8 +9,8 @@ export SOURCE_FILES="httpcore tests"
 set -x
 
 ${PREFIX}autoflake --in-place --recursive $SOURCE_FILES
-${PREFIX}isort --project=httpcore $SOURCE_FILES
-${PREFIX}black --exclude 'httpcore/_sync/.*' --exclude 'tests/sync_tests/.*' --target-version=py36 $SOURCE_FILES
+${PREFIX}isort --skip-glob 'httpcore/_sync/*' --skip-glob 'tests/sync_tests/*' --project=httpcore $SOURCE_FILES
+${PREFIX}black --exclude 'httpcore/_sync/.*|tests/sync_tests/.*' --target-version=py36 $SOURCE_FILES
 
 # Run unasync last because its `--check` mode is not aware of code formatters.
 # (This means sync code isn't prettified, and that's mostly okay.)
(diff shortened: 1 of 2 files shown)


florimondmanca:
I think Black is drunk :-( It fails to check `_sync/http11.py`, `_sync/connection_pool.py` and `_sync/http2.py` even though we told it to `--exclude` the entire `_sync` folder…

https://github.com/encode/httpcore/pull/144/checks?check_run_id=964114004

```console
Oh no! 💥 💔 💥
3 files would be reformatted, 29 files would be left unchanged.
```

euri10:
No I think the correct way to exclude would be 
`${PREFIX}black --verbose --check --diff --exclude 'tests/sync_tests|httpcore/_sync' --target-version=py36 $SOURCE_FILES`

JayH5:
> `${PREFIX}black --verbose --check --diff --exclude 'tests/sync_tests|httpcore/_sync' --target-version=py36 $SOURCE_FILES`

Yeah looks like it only takes the flag once 😕 

I also noticed that the `isort` command in the lint script is missing the `--skip-glob` options.

florimondmanca:
Awesome, I think that did the trick - thanks both :100:

scripts/lint
```diff
@@ -9,8 +9,8 @@ export SOURCE_FILES="httpcore tests"
 set -x
 
 ${PREFIX}autoflake --in-place --recursive $SOURCE_FILES
-${PREFIX}isort --project=httpcore $SOURCE_FILES
-${PREFIX}black --exclude 'httpcore/_sync/.*' --exclude 'tests/sync_tests/.*' --target-version=py36 $SOURCE_FILES
+${PREFIX}isort --skip-glob 'httpcore/_sync/*' --skip-glob 'tests/sync_tests/*' --project=httpcore $SOURCE_FILES
+${PREFIX}black --exclude 'httpcore/_sync/.*|tests/sync_tests/.*' --target-version=py36 $SOURCE_FILES
```
JayH5, line 13: Don't think the `.*` is necessary but eh 🤷‍♂️. Would be nice if black used globs instead of regexes.

jcugat:
Small suggestion, it would be better to move all the arguments of `isort` and `black` to `pyproject.toml`. If we do this, IDEs that have support for them (for example by autoformatting on save) will pickup the configuration automatically.

Example of `pyproject.toml` from one of my repos (could be extended to include `isort` too):
```toml
[tool.black]
line-length = 120
exclude = '''
/(
    \.eggs
  | \.venv
  | venv
)/
'''
```

florimondmanca:
Closing in favor of #156.

PR #153, opened 2020-08-11
Fix max_keepalive_connections config

tomchristie:
Our `max_keepalive_connections` config was not actually being passed along correctly, because we were still setting from the deprecated `max_connections` internally.

We need test cases here, but I'd be okay with us resolving this PR and issue a hotfix release without blocking on that, and then make it a priority.

### DIFF
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -105,7 +105,7 @@ class AsyncConnectionPool(AsyncHTTPTransport):
 
         self._ssl_context = SSLContext() if ssl_context is None else ssl_context
         self._max_connections = max_connections
-        self._max_keepalive = max_keepalive
+        self._max_keepalive_connections = max_keepalive_connections
         self._keepalive_expiry = keepalive_expiry
         self._http2 = http2
         self._uds = uds
@@ -259,8 +259,8 @@ class AsyncConnectionPool(AsyncHTTPTransport):
         elif connection.state == ConnectionState.IDLE:
             num_connections = len(self._get_all_connections())
             if (
-                self._max_keepalive is not None
-                and num_connections > self._max_keepalive
+                self._max_keepalive_connections is not None
+                and num_connections > self._max_keepalive_connections
             ):
                 remove_from_pool = True
                 close_connection = True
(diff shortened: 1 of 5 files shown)


httpcore/_async/connection_pool.py
```diff
@@ -105,7 +105,7 @@ def __init__(
 
         self._ssl_context = SSLContext() if ssl_context is None else ssl_context
         self._max_connections = max_connections
-        self._max_keepalive = max_keepalive
```
tomchristie, line 108: We were setting our state from the deprecated setting.

httpx is passing `max_keepalive_connections`, which won't be having any effect at the moment.

tomchristie:
Sure thing!

dalf:
Can ```__slots__``` be helpful ?

tomchristie:
Thanks but no. httpcore/httpx will generally be network bound, so adding `__slots__` at the moment would mostly just be extra noise.

Perhaps someday, once there's not much change in the codebase over a long period of time, we might think about it, but I'd prefer not to at the moment.

dalf:
My suggestion is not about performance. In this issue, even without a test, an exception would have been raise because ```_max_keepalive``` would have been declared. ```__slots__``` can be seen as a way to declare the fields for each object (in the same way than typing adds meta-information). [/subject close]

cdeler:
@dalf 
> Can \_\_slots__ be helpful ?

Let's image the class like:
```python3
class Foo:
    __slots__ = "bar"

    def __init__(self, a):
        self._not_bar = a
```

Even if you define `__slots__`, `self._not_bar = a` creates a `__dict__` in the object, so it doesn't protect you from the misspell

@tomchristie , @jcugat 
I created a PR (https://github.com/encode/httpcore/pull/160) with the test for this feature, but how I can combine it with the fix?

dalf:
@cdeler : https://docs.python.org/3/reference/datamodel.html#object.__slots__  (the sample code triggers an AttributeError)
(sorry for the out of topic comment)

jcugat:
I'm going to merge this now since it's blocking #156 and #160. Feel free to open a separate issue to discuss about `slots` since it's not really related to this PR anyway.

PR #154, opened 2020-08-12
Always include the port on CONNECT request

j178:
Fixes https://github.com/encode/httpx/issues/1169

Before:
image

After the fix:
image

### DIFF
--- a/httpcore/_async/http_proxy.py
+++ b/httpcore/_async/http_proxy.py
@@ -181,7 +181,7 @@ class AsyncHTTPProxy(AsyncConnectionPool):
         connection = await self._get_connection_from_pool(origin)
 
         if connection is None:
-            scheme, host, port, _ = url
+            scheme, host, port = origin
 
             # First, create a connection to the proxy server
             proxy_connection = AsyncHTTPConnection(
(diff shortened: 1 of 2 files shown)


unasync.py
```diff
@@ -36,7 +36,7 @@ def unasync_line(line):
 
 def unasync_file(in_path, out_path):
     with open(in_path, "r") as in_file:
-        with open(out_path, "w") as out_file:
+        with open(out_path, "w", newline="") as out_file:
```
florimondmanca, line 39: Could you clarify what this change is about?

j178: I am working on Windows, I have to set `newline=""` to prevent `universal newlines`, which auto converts "\n" to "\r\n" on Windows:

https://docs.python.org/3/library/functions.html#open
> When writing output to the stream, if newline is None, any '\n' characters written are translated to the system default line separator, os.linesep. If newline is '' or '\n', no translation takes place. If newline is any of the other legal values, any '\n' characters written are translated to the given string.

florimondmanca: Sounds good. Could we try and submit that as a separate pull request, for traceability purposes? (I might be maniac sometimes on clearly separating unrelated changes in separate PRs, but experience has proven it's beneficial when doing bug forensics…)

j178: Sure 👍

j178: Splited into #157

PR #156, opened 2020-08-12
Add checks

jcugat:
I noticed while reviewing https://github.com/encode/httpcore/pull/153 that some imports were not ordered correctly and some of them were unused. I've pretty much copied the approach from `httpx` regarding the scripts and configs.

Depends on https://github.com/encode/httpcore/pull/153 since one of the check failures is regarding `max_keepalive_connections` which will be fixed in the other PR, and wanted to avoid any conflicts. Funny enough, if we had `flake8` enabled before we would have noticed that bug 😅


httpcore/__init__.py
```diff
@@ -24,29 +24,31 @@
 from ._sync.http_proxy import SyncHTTPProxy
 
 __all__ = [
```
jcugat, line 26: The list has been sorted alphabetically, and added `ProtocolError` and `ProxyError` that were missing I guess due to a mistake.

httpcore/_backends/sync.py
```diff
@@ -137,10 +137,9 @@ def open_tcp_stream(
         exc_map = {socket.timeout: ConnectTimeout, socket.error: ConnectError}
 
         with map_exceptions(exc_map):
-            local_addrport = None
```
jcugat, line 140: `local_addrport` was unused, leftover?

requirements.txt
```diff
@@ -5,23 +5,26 @@ trio
...
 mkautodoc
+mkdocs
 mkdocs-material
 
 # Packaging
 twine
 wheel
 
 # Tests & Linting
-pytest
```
jcugat, line 17: List ordered, and added the `flake8*` packages.

scripts/check
```diff
@@ -8,5 +8,8 @@ export SOURCE_FILES="httpcore tests"
 
 set -x
 
-${PREFIX}mypy --ignore-missing-imports httpcore
+${PREFIX}isort --check --diff --project=httpx $SOURCE_FILES
+${PREFIX}black --check --diff $SOURCE_FILES
+${PREFIX}flake8 $SOURCE_FILES
+${PREFIX}mypy $SOURCE_FILES
```
jcugat, line 14: `--ignore-missing-imports` argument removed since it's already in `setup.cfg`

scripts/test
```diff
@@ -5,22 +5,12 @@ if [ -d 'venv' ] ; then
     export PREFIX="venv/bin/"
 fi
 
+set -ex
+
 if [ -z $GITHUB_ACTIONS ]; then
-    set +e
     scripts/check
-    while [ $? -ne 0 ]; do
-        read -p "Running 'scripts/check' failed. Do you want to run 'scripts/lint' now? [y/N]  " yn
```
jcugat, line 12: Removed this just like we did with `httpx` https://github.com/encode/httpx/pull/1107

scripts/lint
```diff
@@ -10,7 +10,7 @@ set -x
 
 ${PREFIX}autoflake --in-place --recursive $SOURCE_FILES
 ${PREFIX}isort --project=httpcore $SOURCE_FILES
-${PREFIX}black --exclude 'httpcore/_sync/.*' --exclude 'tests/sync_tests/.*' --target-version=py36 $SOURCE_FILES
```
jcugat, line 13: Moved arguments to `pyproject.toml`, this way is also picked up by IDEs.

httpcore/__init__.py
florimondmanca: > added ProtocolError and ProxyError that were missing I guess due to a mistake.

I guess we could add an extra validation step in `test_exceptions.py` for this, as a follow up? :-) Or maybe even a more general test case for "verify that all exported members of `httpcore` as identified by `vars(httpcore)` are present in `__all__`". Maybe even check that the list is sorted… (Without validation the list will go unsorted again soon!)

httpcore/_backends/sync.py
florimondmanca: Yes, leftover from moving this code to computing `source_address` above. :+1:

tests/async_tests/test_interfaces.py
```diff
@@ -313,7 +316,7 @@ async def test_connection_pool_get_connection_info(
     reason="Unix Domain Sockets only exist on Unix",
 )
 @pytest.mark.usefixtures("async_environment")
-async def test_http_request_unix_domain_socket(uds_server) -> None:
+async def test_http_request_unix_domain_socket(uds_server: Server,) -> None:
```
florimondmanca, line 319: (Nit)

```suggestion
async def test_http_request_unix_domain_socket(uds_server: Server) -> None:
```

florimondmanca:
> Funny enough, if we had flake8 enabled before we would have noticed that bug 😅

Indeed…

```console
+ flake8 httpcore tests
httpcore/_async/connection_pool.py:104:13: F841 local variable 'max_keepalive_connections' is assigned to but never used
##[error]Process completed with exit code 1.
```

Well done, flake8!

tests/async_tests/test_interfaces.py
jcugat: Weird, I thought black would normalize this.

httpcore/__init__.py
jcugat: That's a great idea, I've added the test for both the order and make sure that everything is in `__all__`.

jcugat:
Completely forgot about #144 (and I even commented there!).

requirements.txt
```diff
@@ -5,23 +5,26 @@ trio
...
-pytest-asyncio
-pytest-trio
-pytest-cov
-black
 autoflake
-mypy
+black
+flake8
+flake8-bugbear
+flake8-pie
```
JayH5, line 21: I'm definitely +1 on flake8, but less keen on these extras, particularly `flake8-pie` is not very useful as 3/4 of its checks are for Celery. I know in the past @tomchristie has pushed back against using lots of extra linting tools.

jcugat: I added those to be consistent with `httpx`

https://github.com/encode/httpx/blob/655773e1c1b75895eda927d5a9d22a3b5b8f572d/requirements.txt#L21

JayH5: Ah I think I was misremembering when `flake8-comprehensions` got removed from httpx. To be clear, this is a minor nit. I'm happy with `flake8-bugbear`, but `flake8-pie` is literally one check ('PIE781: You are assigning to a variable and then returning. Instead remove the assignment and return.') which seems a bit unnecessary.

Issue #155, opened 2020-08-12
Add tests for ConnectionPool


jcugat:
In https://github.com/encode/httpcore/pull/153 was noticed that the parameters `max_keepalive_connections` / `max_connections` were not handled correctly in `AsyncConnectionPool` and `SyncConnectionPool`. Add some tests to cover it.

cdeler:
@jcugat I added the tests, but they are failing without the changes from https://github.com/encode/httpcore/pull/153 (I mean without `self._max_keepalive = max_keepalive` -> `self._max_keepalive = max_keepalive_connections`)

PR #160, opened 2020-08-13
Added new tests for max_keepalive_connections param (AsyncConnectionPool and SyncConnectionPool) (#155)

cdeler:
Note: the tests are failing unless the fix mentioned in https://github.com/encode/httpcore/issues/155#issuecomment-673683869 has been merged.

### DIFF
--- a/tests/async_tests/test_interfaces.py
+++ b/tests/async_tests/test_interfaces.py
@@ -328,3 +328,31 @@ async def test_http_request_unix_domain_socket(uds_server) -> None:
         assert reason == b"OK"
         body = await read_body(stream)
         assert body == b"Hello, world!"
+
+
+@pytest.mark.usefixtures("async_environment")
+@pytest.mark.parametrize("max_keepalive", [1, 3, 5])
+@pytest.mark.parametrize("connections_number", [4])
+async def test_max_keepalive_connections_handled_correctly(
+    max_keepalive, connections_number
+) -> None:
+    async with httpcore.AsyncConnectionPool(
+        max_keepalive_connections=max_keepalive, keepalive_expiry=60
+    ) as http:
+        method = b"GET"
+        url = (b"http", b"example.org", 80, b"/")
+        headers = [(b"host", b"example.org")]
+
+        connections_streams = []
+        for _ in range(connections_number):
+            _, _, _, _, stream = await http.request(method, url, headers)
+            connections_streams.append(stream)
+
+        try:
+            for i in range(len(connections_streams)):
+                await read_body(connections_streams[i])
+        finally:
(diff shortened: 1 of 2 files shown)


tests/async_tests/test_interfaces.py
```diff
@@ -331,3 +331,31 @@ async def test_http_request_unix_domain_socket(uds_server: Server) -> None:
         assert reason == b"OK"
         body = await read_body(stream)
         assert body == b"Hello, world!"
+
+
+@pytest.mark.usefixtures("async_environment")
+@pytest.mark.parametrize("max_keepalive", [1, 3, 5])
+@pytest.mark.parametrize("connections_number", [4])
+async def test_max_keepalive_connections_handled_correctly(
+    max_keepalive, connections_number
```
j178, line 340: Some missing annotations:
```suggestion
    max_keepalive: int, connections_number: int
```

tests/async_tests/test_interfaces.py
```diff
@@ -331,3 +331,31 @@ async def test_http_request_unix_domain_socket(uds_server: Server) -> None:
...
+        connections_streams = []
+        for _ in range(connections_number):
+            _, _, _, _, stream = await http.request(method, url, headers)
+            connections_streams.append(stream)
+
+        try:
+            for i in range(len(connections_streams)):
+                await read_body(connections_streams[i])
+        finally:
+            stats = await http.get_connection_info()
```
j178, line 358: May be we can just use `http._connections` like other tests?

tests/async_tests/test_interfaces.py
cdeler: I'll fix it as soon as I reach my laptop

florimondmanca: If not done already I'd recommend we allow untyped defs in `tests/` (I _think_ we have this tweak in HTTPX, for reference?), so that we're not forced to make add these annotations that often only feel like busy work. :-)

cdeler: > Some missing annotations:
> ```suggestion
>     max_keepalive: int, connections_number: int
> ```

I fixed the typehints (I'm sorry for splitting that into 2 commits, I'm working from the mobile)

j178: > If not done already I'd recommend we allow untyped defs in `tests/` (I _think_ we have this tweak in HTTPX, for reference?), so that we're not forced to make add these annotations that often only feel like busy work. :-)

I see that in httpx:
https://github.com/encode/httpx/blob/78cf16ace92b42de5972a490787002631d558b32/setup.cfg#L9-L11

I'll add it to httpcore later.

PR #151, opened 2020-08-11
Pin sortedcontainers version to avoid requirements' installation error

cdeler:

PR #163, opened 2020-08-15
Use pip 2020 resolver

florimondmanca:
Promted by https://github.com/encode/httpcore/pull/151#issuecomment-672290787 and https://github.com/encode/httpcore/pull/162/files#r470961811, alternative to #151.

It's better we use the (still beta, but soon enabled-by-default) new pip resolver anyway!

https://blog.python.org/2020/07/upgrade-pip-20-2-changes-20-3.html

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,5 @@
+--use-feature="2020-resolver"
+
 -e .[http2]
 
 # Optionals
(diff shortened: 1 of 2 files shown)


florimondmanca:
Closing since this will most likely be integrated via #162 or #184 anyway.

PR #164, opened 2020-08-18
Fix server disconnect behaviour.

tomchristie:
Based on #159 but narrowing things down a bit.

Fixes an issue in HTTP/1.1 when the server does not include the `Content-Length` or set `Transfer-Encoding: chunked` but instead uses a server disconnect to signal the end of the response body.

If the server disconnects under HTTP/2 then raise an appropriate error.


tomchristie:
We've also still got some behaviour we can improve on this, around raising the error on multiple streams. (And a related issue about better handling connection-wide HTTP/2 error conditions.) But let's treat those as follow-ups.

florimondmanca (approved):

PR #165, opened 2020-08-19
Version 0.10.2

tomchristie:
### Added

- Added Unix Domain Socket support. (Pull #139)

### Fixed

- Always include the port on proxy CONNECT requests. (Pull #154)
- Fix `max_keepalive_connections` configuration. (Pull #153)
- Fixes behaviour in HTTP/1.1 where server disconnects can be used to signal the end of the response body. (Pull #164)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.10.2 (August 19th, 2020)
+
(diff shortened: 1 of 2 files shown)


florimondmanca (approved):
Thoughts on observing stricter SemVer and bumping to 0.11.0? Or does that require issuing an HTTPX release as well?

tomchristie:
I think SEMVER says all bets are off on 0.x releases, so I'm okay with our policy of just "API incompatible changes require a new median point release", but yes once we're 1.x we'll likely want to observe that for feature bumps.

Issue #94, opened 2020-05-22
Curio support


brettcannon:
Other than asyncio and trio, [curio](https://github.com/dabeaz/curio) seems to be the other "big" event loop implementation.

tomchristie:
Well indeed. I’ve not yet reached out to @dabeaz to gauge interest here. The backends are a pretty small amount of code, so it’s probably not going to be a *huge* amount of work if someone’s motivated enough to take it on.

Also, curio isn’t the only other place we could be reaching out to here... We’re following [the lead of the `urllib3` team](https://github.com/python-trio/hip) in our approach here, who had also been looking into providing a Twisted backend too.

I don’t really know anything about that ecosystem, but could be worth reaching out to Amber or Glyph there to figure out if that’s also a valuable avenue to look into.

dabeaz:
I'd definitely be interested in seeing Curio support here, but don't really know what's involved.  Curio recently went through a "version 1.0" release that reduced its set of features to a smaller core and likely broke some things for earlier adopters.  However, I consider the current version of Curio to be stable and more-or-less feature complete.

How would this work?   Should I just submit a drive-by pull request adding Curio support?  (I'm more than happy to have someone else do it if they're motivated though).

tomchristie:
Oh cool, hiya! 👋

> but don't really know what's involved

Implementing a concurrency backend requires implementing the `AsyncBackend` interface... https://github.com/encode/httpcore/blob/master/httpcore/_backends/base.py

That provides the bare bones for "give me a semaphore", "give me a lock", and "open a TCP stream", which is everything the backend needs to provide.

The [asyncio](https://github.com/encode/httpcore/blob/master/httpcore/_backends/asyncio.py) and [trio](https://github.com/encode/httpcore/blob/master/httpcore/_backends/trio.py) implementations are worth a quick look over, as the two existing async examples. (There's an equivalent sync implementation too.)

Once there's an implementation, the other bit is to tie it in with the concurrency autodetection... https://github.com/encode/httpcore/blob/master/httpcore/_backends/auto.py#L17-L28 

For the tests I think it'd be a case of adding curio to https://github.com/encode/httpcore/blob/master/tests/conftest.py#L15-L20 in order to include it in the test parameterization, but I don't recall off the top of my head if there's more beyond that to do.

> Should I just submit a drive-by pull request adding Curio support?

That'd be pretty fantastic if you're up for it.

brettcannon:
[...]

PR #168, opened 2020-08-24
Implement curio backend

cdeler:
I'm trying to implement `curio` backend (closes https://github.com/encode/httpcore/issues/94)

**Warning** : work in progress

What I did:
1. implemented `curio` backend

What I had to change, but **didn't want** (I'm not an `asycn`-guru, I didn't find a way how to clearly invoke a coroutine from the sync context):
1. I made `AsyncLock` and `AsyncSemaphore` release methods async
1. I made `AsyncBackend::time` async
1. I added `pytest.mark.curio` definition to the `conftest.py`

### DIFF
--- a/httpcore/_backends/curio.py
+++ b/httpcore/_backends/curio.py
@@ -1,0 +1,189 @@
+from ssl import SSLContext
+from typing import Optional
+
+import curio
+import curio.io
+from curio.network import _wrap_ssl_client
+
+from .._exceptions import (
+    ConnectError,
+    ConnectTimeout,
+    ReadError,
+    ReadTimeout,
+    WriteError,
+    WriteTimeout,
+    map_exceptions,
+)
+from .._types import TimeoutDict
+from .._utils import get_logger
+from .base import AsyncBackend, AsyncLock, AsyncSemaphore, AsyncSocketStream
+
+logger = get_logger("curio_backend")
+
+one_day_in_seconds = 60 * 60 * 24
+
+
+def convert_timeout(value: Optional[float]) -> int:
+    return int(value) if value is not None else one_day_in_seconds
+
+
+class Lock(AsyncLock):
+    def __init__(self) -> None:
+        self._lock = curio.Lock()
+
+    async def acquire(self) -> None:
+        await self._lock.acquire()
+
+    async def release(self) -> None:
+        await self._lock.release()
+
+
+class Semaphore(AsyncSemaphore):
+    def __init__(self, max_value: int, exc_class: type) -> None:
+        self.max_value = max_value
+        self.exc_class = exc_class
+
+    @property
+    def semaphore(self) -> curio.Semaphore:
+        if not hasattr(self, "_semaphore"):
+            self._semaphore = curio.Semaphore(value=self.max_value)
+        return self._semaphore
+
+    async def acquire(self, timeout: float = None) -> None:
+        await self.semaphore.acquire()
+
+    async def release(self) -> None:
+        await self.semaphore.release()
+
+
+class SocketStream(AsyncSocketStream):
+    def __init__(self, socket: curio.io.Socket) -> None:
+        self.read_lock = curio.Lock()
+        self.write_lock = curio.Lock()
+        self.socket = socket
+
+    def get_http_version(self) -> str:
+        if hasattr(self.socket._socket, "_sslobj"):
+            ident = self.socket._socket._sslobj.selected_alpn_protocol()
+        else:
+            ident = "http/1.1"
+        return "HTTP/2" if ident == "h2" else "HTTP/1.1"
+
+    async def start_tls(
+        self, hostname: bytes, ssl_context: SSLContext, timeout: TimeoutDict
+    ) -> "AsyncSocketStream":
+        connect_timeout = convert_timeout(timeout.get("connect"))
+        exc_map = {
+            curio.TaskTimeout: ConnectTimeout,
+            curio.CurioError: ConnectError,
+            OSError: ConnectError,
+        }
+
+        with map_exceptions(exc_map):
+            wrapped_sock = await curio.timeout_after(
+                connect_timeout,
+                _wrap_ssl_client(
+                    self.socket,
+                    ssl=ssl_context,
+                    server_hostname=hostname,
+                    alpn_protocols=["h2", "http/1.1"],
+                ),
+            )
+
+            return SocketStream(wrapped_sock)
+
+    async def read(self, n: int, timeout: TimeoutDict) -> bytes:
+        read_timeout = convert_timeout(timeout.get("read"))
+        exc_map = {
+            curio.TaskTimeout: ReadTimeout,
+            curio.CurioError: ReadError,
+            OSError: ReadError,
+        }
+
+        with map_exceptions(exc_map):
+            async with self.read_lock:
+                socket_stream = self.socket.as_stream()
+
+                return await curio.timeout_after(read_timeout, socket_stream.read(n))
+
+    async def write(self, data: bytes, timeout: TimeoutDict) -> None:
+        write_timeout = convert_timeout(timeout.get("write"))
+        exc_map = {
+            curio.TaskTimeout: WriteTimeout,
+            curio.CurioError: WriteError,
+            OSError: WriteError,
+        }
+
+        with map_exceptions(exc_map):
+            async with self.write_lock:
+                socket_stream = self.socket.as_stream()
+                await curio.timeout_after(write_timeout, socket_stream.write(data))
+
+    async def aclose(self) -> None:
+        await self.socket.close()
+
+    def is_connection_dropped(self) -> bool:
+        return self.socket._closed
+
+
+class CurioBackend(AsyncBackend):
+    async def open_tcp_stream(
+        self,
+        hostname: bytes,
+        port: int,
+        ssl_context: Optional[SSLContext],
+        timeout: TimeoutDict,
+        *,
+        local_address: Optional[str],
+    ) -> AsyncSocketStream:
+        connect_timeout = convert_timeout(timeout.get("connect"))
+        exc_map = {
+            curio.TaskTimeout: ConnectTimeout,
+            curio.CurioError: ConnectError,
+            OSError: ConnectError,
+        }
+        host = hostname.decode("ascii")
+        kwargs = (
+            {} if not ssl_context else {"ssl": ssl_context, "server_hostname": host}
+        )
+
+        with map_exceptions(exc_map):
+            sock: curio.io.Socket = await curio.timeout_after(
+                connect_timeout, curio.open_connection(hostname, port, **kwargs)
+            )
+
+            return SocketStream(sock)
+
+    async def open_uds_stream(
+        self,
+        path: str,
+        hostname: bytes,
+        ssl_context: Optional[SSLContext],
+        timeout: TimeoutDict,
+    ) -> AsyncSocketStream:
+        connect_timeout = convert_timeout(timeout.get("connect"))
+        exc_map = {
+            curio.TaskTimeout: ConnectTimeout,
+            curio.CurioError: ConnectError,
+            OSError: ConnectError,
+        }
+        host = hostname.decode("ascii")
+        kwargs = (
+            {} if not ssl_context else {"ssl": ssl_context, "server_hostname": host}
+        )
+
+        with map_exceptions(exc_map):
+            sock: curio.io.Socket = await curio.timeout_after(
+                connect_timeout, curio.open_unix_connection(path, **kwargs)
+            )
+
+            return SocketStream(sock)
+
+    def create_lock(self) -> AsyncLock:
+        return Lock()
+
+    def create_semaphore(self, max_value: int, exc_class: type) -> AsyncSemaphore:
+        return Semaphore(max_value, exc_class)
+
+    async def time(self) -> float:
+        return float(await curio.clock())
--- a/httpcore/_backends/curio.py
+++ b/httpcore/_backends/curio.py
@@ -1,5 +1,7 @@
-from ssl import SSLContext
-from typing import Optional
+import select
+import socket
+from ssl import SSLContext, SSLSocket
+from typing import Dict, Optional, Type, Union
 
 import curio
 import curio.io
@@ -10,6 +12,7 @@ from .._exceptions import (
     ConnectTimeout,
     ReadError,
     ReadTimeout,
+    TimeoutException,
     WriteError,
     WriteTimeout,
     map_exceptions,
@@ -50,7 +53,13 @@ class Semaphore(AsyncSemaphore):
         return self._semaphore
 
     async def acquire(self, timeout: float = None) -> None:
-        await self.semaphore.acquire()
+        exc_map: Dict[Type[Exception], Type[Exception]] = {
+            curio.TaskTimeout: TimeoutException,
+        }
+        acquire_timeout: int = convert_timeout(timeout)
+
+        with map_exceptions(exc_map):
+            return await curio.timeout_after(acquire_timeout, self.semaphore.acquire())
 
     async def release(self) -> None:
         await self.semaphore.release()
@@ -64,10 +73,14 @@ class SocketStream(AsyncSocketStream):
         self.stream = socket.as_stream()
 
     def get_http_version(self) -> str:
-        if hasattr(self.socket, "_socket") and hasattr(self.socket._socket, "_sslobj"):
-            ident = self.socket._socket._sslobj.selected_alpn_protocol()
-        else:
-            ident = "http/1.1"
+        ident: Optional[str] = "http/1.1"
+
+        if hasattr(self.socket, "_socket"):
+            raw_socket: Union[SSLSocket, socket.socket] = self.socket._socket
+
+            if isinstance(raw_socket, SSLSocket):
+                ident = raw_socket.selected_alpn_protocol()
+
         return "HTTP/2" if ident == "h2" else "HTTP/1.1"
 
     async def start_tls(
@@ -118,11 +131,13 @@ class SocketStream(AsyncSocketStream):
                 await curio.timeout_after(write_timeout, self.stream.write(data))
 
     async def aclose(self) -> None:
-        # we dont need to close the self.socket, since it's closed by stream closing
         await self.stream.close()
+        await self.socket.close()
 
     def is_connection_dropped(self) -> bool:
-        return self.socket._closed
+        rready, _, _ = select.select([self.socket.fileno()], [], [], 0)
+
+        return bool(rready)
 
 
 class CurioBackend(AsyncBackend):
--- a/httpcore/_backends/curio.py
+++ b/httpcore/_backends/curio.py
@@ -18,7 +18,7 @@ from .._types import TimeoutDict
 from .._utils import get_logger
 from .base import AsyncBackend, AsyncLock, AsyncSemaphore, AsyncSocketStream
 
-logger = get_logger("curio_backend")
+logger = get_logger(__name__)
 
 ONE_DAY_IN_SECONDS = float(60 * 60 * 24)
 
@@ -199,4 +199,4 @@ class CurioBackend(AsyncBackend):
         return Semaphore(max_value, exc_class)
 
     async def time(self) -> float:
-        return float(await curio.clock())
+        return await curio.clock()
--- a/httpcore/_backends/curio.py
+++ b/httpcore/_backends/curio.py
@@ -1,7 +1,6 @@
 import select
-import socket
 from ssl import SSLContext, SSLSocket
-from typing import Dict, Optional, Type, Union
+from typing import Optional
 
 import curio
 import curio.io
@@ -11,7 +10,6 @@ from .._exceptions import (
     ConnectTimeout,
     ReadError,
     ReadTimeout,
-    TimeoutException,
     WriteError,
     WriteTimeout,
     map_exceptions,
@@ -22,27 +20,11 @@ from .base import AsyncBackend, AsyncLock, AsyncSemaphore, AsyncSocketStream
 
 logger = get_logger("curio_backend")
 
-one_day_in_seconds = 60 * 60 * 24
+ONE_DAY_IN_SECONDS = float(60 * 60 * 24)
 
 
-async def wrap_ssl_client(
-    sock: curio.io.Socket,
-    ssl_context: SSLContext,
-    server_hostname: bytes,
-) -> curio.io.Socket:
-    kwargs = {
-        "server_hostname": server_hostname,
-        "do_handshake_on_connect": sock._socket.gettimeout() != 0.0,
-    }
-
-    socket = curio.io.Socket(ssl_context.wrap_socket(sock._socket, **kwargs))
-    await socket.do_handshake()
-
-    return socket
-
-
-def convert_timeout(value: Optional[float]) -> int:
-    return int(value) if value is not None else one_day_in_seconds
+def convert_timeout(value: Optional[float]) -> float:
+    return value if value is not None else ONE_DAY_IN_SECONDS
 
 
 class Lock(AsyncLock):
@@ -68,13 +50,12 @@ class Semaphore(AsyncSemaphore):
         return self._semaphore
 
     async def acquire(self, timeout: float = None) -> None:
-        exc_map: Dict[Type[Exception], Type[Exception]] = {
-            curio.TaskTimeout: TimeoutException,
-        }
-        acquire_timeout: int = convert_timeout(timeout)
+        timeout = convert_timeout(timeout)
 
-        with map_exceptions(exc_map):
-            return await curio.timeout_after(acquire_timeout, self.semaphore.acquire())
+        try:
+            return await curio.timeout_after(timeout, self.semaphore.acquire())
+        except curio.TaskTimeout:
+            raise self.exc_class()
 
     async def release(self) -> None:
         await self.semaphore.release()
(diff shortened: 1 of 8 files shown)


httpcore/_backends/curio.py
```diff
@@ -0,0 +1,189 @@
...
+        read_timeout = convert_timeout(timeout.get("read"))
+        exc_map = {
+            curio.TaskTimeout: ReadTimeout,
+            curio.CurioError: ReadError,
+            OSError: ReadError,
+        }
+
+        with map_exceptions(exc_map):
+            async with self.read_lock:
+                socket_stream = self.socket.as_stream()
```
florimondmanca, line 105: Is this "gimme a stream then read/write" dance idiomatic in curio? Building objects on each read and write might not be as efficient so wondering if there's a way to just eg `self.socket.read(n)`.

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,189 @@
...
+        exc_map = {
+            curio.TaskTimeout: ConnectTimeout,
+            curio.CurioError: ConnectError,
+            OSError: ConnectError,
+        }
+
+        with map_exceptions(exc_map):
+            wrapped_sock = await curio.timeout_after(
+                connect_timeout,
+                _wrap_ssl_client(
```
florimondmanca, line 85: How confident are we using this private piece of curio API? Anything like a public `TLSStream` wrapper that could be used instead?

tests/conftest.py
```diff
@@ -18,15 +23,61 @@
...
+
+# Fixture for explicitly running in Kernel instance.
+@pytest.fixture(scope="session")
+def kernel(request):
+    """Provide a Curio Kernel object for running co-routines."""
+    k = curio.Kernel(debug=[curio.debug.longblock, curio.debug.logcrash])
+    m = curio.monitor.Monitor(k)
+    request.addfinalizer(lambda: k.run(shutdown=True))
+    request.addfinalizer(m.close)
+    return k
```
florimondmanca, line 68: Is all of this mostly vendorized code from https://github.com/johnnoone/pytest-curio?

I see ^ is really tiny, so it not looking maintained anymore (5y+) isn't really a problem since the project is most likely "done" and doesn't even need maintenance. So I wouldn't be against adding it as a test dependency. If we do prefer to have this glue code as-is in our repo then _maybe_ at least let's put it somewhere isolated, eg in `tests/marks/curio.py`?

setup.cfg
```diff
@@ -23,3 +23,4 @@ skip = httpcore/_sync/,tests/sync_tests/
 addopts = --cov-report= --cov=httpcore --cov=tests -rxXs
 markers =
   copied_from(source, changes=None): mark test as copied from somewhere else, along with a description of changes made to accodomate e.g. our test setup
+  curio: mark the test as a coroutine, it will be run using a Curio kernel.
```
florimondmanca, line 26: I _think_ there's a programmatic way to tell pytest "mark this mark as known" from one of the pytest hooks, though I don't exactly remember which one. Might be preferable to keep the curio pytest integration self-contained in a single file (and so easy to remove eg if/when we move to anyio).

tests/conftest.py
```diff
@@ -18,15 +23,61 @@
...
     params=[
         pytest.param("asyncio", marks=pytest.mark.asyncio),
         pytest.param("trio", marks=pytest.mark.trio),
+        pytest.param("curio", marks=pytest.mark.curio),
     ]
 )
 def async_environment(request: typing.Any) -> str:
     """
-    Mark a test function to be run on both asyncio and trio.
+    Mark a test function to be run on asyncio, trio and curio.
```
florimondmanca, line 80: Nice yup, we also need to update the rest of the fix string to account for this 3rd async environment.

florimondmanca (commented):
Looks brilliant!

So far most of the backend looks good to me, and the test suite passes. What areas are you least confident about?

httpcore/_backends/curio.py
tomchristie: Or `self.socket_stream = self.socket.as_stream()` just once, in the `__init__`, right?

cdeler: I need to have an access to socket to implement `get_http_version`, so I'd rewrite it to `self.socket.read(n)` mentioned by @florimondmanca

tomchristie:
> Looks brilliant!

Seconded! 🌟

tests/conftest.py
cdeler: There is a problem with this package. It's obsolete.
[This line](https://github.com/johnnoone/pytest-curio/blob/6c93257244c8bd9c0bcf5f717758fdf712a70ca3/pytest_curio/plugin.py#L16
) caused a pytest deprecation warning (which is an Exception during the build). So that I had to copy the plugin example from [there](https://github.com/dabeaz/curio/blob/ee27c4515759616747b9991857169924fdb52824/examples/pytest_plugin.py).

tests/conftest.py
cdeler: Done in https://github.com/encode/httpcore/pull/168/commits/e4238a2213c55ad287f043f698c5bc53b1ede211

httpcore/_backends/curio.py
cdeler: I thought about that. This function did exactly the same which I wanted: takes a curio Socket and wrap it using ssl context.

If I reimplement it by my own, the result will look like the `_wrap_ssl_client` function body

httpcore/_backends/curio.py
cdeler: Done.

But I had to store both `socket` and `stream` in the class, because some methods requires an access to the socket

tests/conftest.py
cdeler: So after that ^ I added the new commit a44909e70bde469157f21ec2430159e0c082df68 

If someone need to add new backend with custom pytest marks etc, they will need to do the same

tests/marks/curio.py
```diff
@@ -0,0 +1,53 @@
...
+
+# Fixture for explicitly running in Kernel instance.
+@pytest.fixture(scope="session")
+def curio_kernel_fixture(request):
+    """Provide a Curio Kernel object for running co-routines."""
+    k = curio.Kernel(debug=[curio.debug.longblock, curio.debug.logcrash])
+    m = curio.monitor.Monitor(k)
+    request.addfinalizer(lambda: k.run(shutdown=True))
+    request.addfinalizer(m.close)
+    return k
```
florimondmanca, line 53: (Nit) Consider using non-abbreviated variable names:

```suggestion
    kernel = curio.Kernel(debug=[curio.debug.longblock, curio.debug.logcrash])
    monitor = curio.monitor.Monitor(kernel)
    request.addfinalizer(lambda: kernel.run(shutdown=True))
    request.addfinalizer(monitor.close)
    return k
```

tests/marks/curio.py
```diff
@@ -0,0 +1,53 @@
...
+    def inner(**kwargs):
+        coro = func(**kwargs)
+        curio.Kernel().run(coro, shutdown=True)
+
+    return inner
+
+
+# Fixture for explicitly running in Kernel instance.
+@pytest.fixture(scope="session")
+def curio_kernel_fixture(request):
```
florimondmanca, line 47: It seems that this fixture is not actually used by any of our test code. Is the setup done here required so that `curio.Kernel().run()` can run at all?

tests/conftest.py
```diff
@@ -14,22 +14,36 @@
...
 
 
+@pytest.mark.tryfirst
+def pytest_pycollect_makeitem(collector, name, obj):
+    curio_pytest_pycollect_makeitem(collector, name, obj)
+
+
+@pytest.hookimpl(tryfirst=True, hookwrapper=True)
+def pytest_pyfunc_call(pyfuncitem):
+    yield from curio_pytest_pyfunc_call(pyfuncitem)
```
florimondmanca, line 31: Very nice namespacing convention there. :+1:

tests/marks/curio.py
cdeler: Let me check...

cdeler: Thank you, I've removed the fixture

setup.cfg
florimondmanca: There — found this programmatic method here: https://docs.pytest.org/en/stable/example/markers.html#custom-marker-and-command-line-option-to-control-test-runs

```python
def pytest_configure(config):
    # register an additional marker
    config.addinivalue_line(
        "markers", "env(name): mark test to run only on named environment"
    )
```

cdeler: Thank you, I added that to `test/conftest.py` in https://github.com/encode/httpcore/pull/168/commits/5c1e69bbf8ee8be31beb391bc76f6e1a5a92aa75 . So that the "curio" marker is discoverable using `pytest --markers`:
```
% pytest --markers | grep curio
@pytest.mark.curio: mark the test as a coroutine, it will be run using a Curio kernel.
```

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,188 @@
...
+        self.exc_class = exc_class
+
+    @property
+    def semaphore(self) -> curio.Semaphore:
+        if not hasattr(self, "_semaphore"):
+            self._semaphore = curio.Semaphore(value=self.max_value)
+        return self._semaphore
+
+    async def acquire(self, timeout: float = None) -> None:
+        await self.semaphore.acquire()
```
tomchristie, line 53: It looks like we're missing some timeout behaviour here...

We ought to use the timeout value, and raise a `self.exc_class` exception if we timeout.

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,188 @@
...
+class SocketStream(AsyncSocketStream):
+    def __init__(self, socket: curio.io.Socket) -> None:
+        self.read_lock = curio.Lock()
+        self.write_lock = curio.Lock()
+        self.socket = socket
+        self.stream = socket.as_stream()
+
+    def get_http_version(self) -> str:
+        if hasattr(self.socket, "_socket") and hasattr(self.socket._socket, "_sslobj"):
+            ident = self.socket._socket._sslobj.selected_alpn_protocol()
```
tomchristie, line 68: Is there any public API in Curio that'd allow us to get the raw socket instance, or the SSL information?

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,188 @@
...
+        with map_exceptions(exc_map):
+            async with self.write_lock:
+                await curio.timeout_after(write_timeout, self.stream.write(data))
+
+    async def aclose(self) -> None:
+        # we dont need to close the self.socket, since it's closed by stream closing
+        await self.stream.close()
+
+    def is_connection_dropped(self) -> bool:
+        return self.socket._closed
```
tomchristie, line 125: Is there any public API in Curio that we can use instead of relying on private API here?

tests/conftest.py
tomchristie: Okay, that all makes sense. I'd recommend we maybe just have a comment in our own `marks/curio` module explaining that it's a vendoring of `pytest-curio`, because that package is out of date.

httpcore/_backends/curio.py
cdeler: I haven't found such API. I'm digging it again

httpcore/_backends/curio.py
cdeler: I did it like it's done in #169 (using `select` with `0` timeout)

httpcore/_backends/curio.py
cdeler: Sure, I forgot about the timeout

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,214 @@
...
+    sock: curio.io.Socket,
+    ssl_context: SSLContext,
+    server_hostname: bytes,
+) -> curio.io.Socket:
+    kwargs = {
+        "server_hostname": server_hostname,
+        "do_handshake_on_connect": sock._socket.gettimeout() != 0.0,
+    }
+
+    socket = curio.io.Socket(ssl_context.wrap_socket(sock._socket, **kwargs))
```
florimondmanca, line 38: (Nit) The `**kwargs` style is a bit surprising, should we consider passing kwargs directly to `.wrap_socket()`?

```suggestion
    socket = curio.io.Socket(
        ssl_context.wrap_socket(
            sock._socket,   
            server_hostname=server_hostname,
            do_handshake_on_connect=sock._socket.gettimeout() != 0,
        )
    )
```

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,214 @@
...
+    def get_http_version(self) -> str:
+        ident: Optional[str] = "http/1.1"
+
+        if hasattr(self.socket, "_socket"):
+            raw_socket: Union[SSLSocket, socket.socket] = self.socket._socket
+
+            if isinstance(raw_socket, SSLSocket):
+                ident = raw_socket.selected_alpn_protocol()
+
+        return "HTTP/2" if ident == "h2" else "HTTP/1.1"
```
florimondmanca, line 99: (Nit) Curious if the code paths look clearer if we reorganize things like this to not have a "mutable" `ident` variable…? (Also tentatively dropping the `Union` type hint since I'm assuming it's not required because we're casting to an `SSLSocket` in the `isinstance` branch, but I might be wrong there.)

```suggestion
        if hasattr(self.socket, "_socket"):
            raw_socket = self.socket._socket

            if isinstance(raw_socket, SSLSocket):
                ident = raw_socket.selected_alpn_protocol()
                return "HTTP/1.1" if ident == "h2" else "HTTP/1.1"

        return "HTTP/1.1"
```

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,214 @@
...
+            curio.TaskTimeout: ConnectTimeout,
+            curio.CurioError: ConnectError,
+            OSError: ConnectError,
+        }
+
+        with map_exceptions(exc_map):
+            wrapped_sock = await curio.timeout_after(
+                connect_timeout,
+                wrap_ssl_client(self.socket, ssl_context, hostname),
+            )
```
florimondmanca, line 115: Since the `wrap_ssl_client` function is only used here, should we consider just moving code in here…? (Benefit: it's clearer that the TLS handshake is the only bit that might cause a timeout.)

```suggestion
            wrapped_socket = curio.io.Socket(
                ssl_context.wrap_socket(
                    self.socket._socket,   
                    server_hostname=hostname,
                    do_handshake_on_connect=self.socket._socket.gettimeout() != 0,
                )
            )

            await curio.timeout_after(
                connect_timeout,
                wrapped_sock.do_handshake(),
            )
```

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,214 @@
...
+    ) -> AsyncSocketStream:
+        connect_timeout = convert_timeout(timeout.get("connect"))
+        exc_map = {
+            curio.TaskTimeout: ConnectTimeout,
+            curio.CurioError: ConnectError,
+            OSError: ConnectError,
+        }
+        host = hostname.decode("ascii")
+        kwargs = (
+            {} if not ssl_context else {"ssl": ssl_context, "server_hostname": host}
```
florimondmanca, line 171: (Nit)

```suggestion
            {} if ssl_context is None else {"ssl": ssl_context, "server_hostname": host}
```

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,214 @@
...
```
florimondmanca, line 197: (Nit)

```suggestion
            {} if ssl_context is None else {"ssl": ssl_context, "server_hostname": host}
```

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,214 @@
...
+    WriteTimeout,
+    map_exceptions,
+)
+from .._types import TimeoutDict
+from .._utils import get_logger
+from .base import AsyncBackend, AsyncLock, AsyncSemaphore, AsyncSocketStream
+
+logger = get_logger("curio_backend")
+
+one_day_in_seconds = 60 * 60 * 24
```
florimondmanca, line 25: (Nit)

```suggestion
ONE_DAY_IN_SECONDS = 60 * 60 * 24
```

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    socket = curio.io.Socket(ssl_context.wrap_socket(sock._socket, **kwargs))
+    await socket.do_handshake()
+
+    return socket
+
+
+def convert_timeout(value: Optional[float]) -> int:
+    return int(value) if value is not None else one_day_in_seconds
```
florimondmanca, line 45: (Nit)

Plus, I suppose we're using "a very long timeout" because Curio doesn't support passing `float("inf")` — or does it?

```suggestion
    return int(value) if value is not None else ONE_DAY_IN_SECONDS
```

tests/conftest.py
```diff
@@ -14,22 +14,43 @@
...
 from httpcore._types import URL
 
+from .marks.curio import curio_pytest_pycollect_makeitem, curio_pytest_pyfunc_call
+
 PROXY_HOST = "127.0.0.1"
 PROXY_PORT = 8080
 
 
+def pytest_configure(config):
+    # register an additional marker
```
florimondmanca, line 24: (Nit) I think the code is self-explanatory here, we don't really need this comment (I think it came from the pytest docs example anyway?).

```suggestion
```

florimondmanca (commented):
Leaving a bunch of nits, but overall this is looking _great_ :100: Great work, thank you! :clap:

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,214 @@
...
+
+    @property
+    def semaphore(self) -> curio.Semaphore:
+        if not hasattr(self, "_semaphore"):
+            self._semaphore = curio.Semaphore(value=self.max_value)
+        return self._semaphore
+
+    async def acquire(self, timeout: float = None) -> None:
+        exc_map: Dict[Type[Exception], Type[Exception]] = {
+            curio.TaskTimeout: TimeoutException,
```
florimondmanca, line 72: I think this should map to `self.exc_class`?

```suggestion
            curio.TaskTimeout: self.exc_class,
```

(Also, nit: we could get away with `exc_map: dict` instead of the very-precise-but-large type hint here, but not fussed either way!)

httpcore/_backends/curio.py
cdeler: Once I tried to do it like in `trio` (treat `None` as `float("inf")`, I stuck with the following problem:
<details>
<summary>Error log for float("inf")</summary>
    ```
self = <selectors.KqueueSelector object at 0x10f2ef310>, timeout = inf

    def select(self, timeout=None):
        timeout = None if timeout is None else max(timeout, 0)
        max_ev = len(self._fd_to_key)
        ready = []
        try:
>           kev_list = self._selector.control(None, max_ev, timeout)
E           TypeError: timeout argument must be a number or None, got float
    ```
</details>

Once I tried to keep `None` as `None`, I got a `curio` error:
<details>
<summary>Error log for None</summary>
    ```
self = <curio.time._TimeoutAfter object at 0x104031520>, ty = None, val = None
tb = None

    async def __aexit__(self, ty, val, tb):
        current_clock = await _unset_timeout(self._prior)
    
        # Discussion.  If a timeout has occurred, it will either
        # present itself here as a TaskTimeout or TimeoutCancellationError
        # exception.  The value of this exception is set to the current
        # kernel clock which can be compared against our own deadline.
        # What happens next is driven by these rules:
        #
        # 1.  If we are the outer-most context where the timeout
        #     period has expired, then a TaskTimeout is raised.
        #
        # 2.  If the deadline has expired for at least one outer
        #     context, (but not us), a TimeoutCancellationError is
        #     raised.  This means that time has expired elsewhere.
        #     We're being cancelled because of that, but the reason
        #     for the cancellation wasn't due to a timeout on our
        #     part.
        #
        # 3.  If the timeout period has not expired on ANY remaining
        #     timeout context, it means that a timeout has escaped
        #     some inner timeout context where it should have been
        #     caught. This is an operational error.  We raise
        #     UncaughtTimeoutError.
    
        try:
            if ty in (TaskTimeout, TimeoutCancellationError):
                timeout_clock = val.args[0]
                # Find the outer most deadline that has expired
                for n, deadline in enumerate(self._deadlines):
                    if deadline <= timeout_clock:
                        break
                else:
                    # No remaining context has expired. An operational error
                    raise UncaughtTimeoutError('Uncaught timeout received')
    
                if n < len(self._deadlines) - 1:
                    if ty is TaskTimeout:
                        raise TimeoutCancellationError(val.args[0]).with_traceback(tb) from None
                    else:
                        return False
                else:
                    # The timeout is us.  Make sure it's a TaskTimeout (unless ignored)
                    self.result = self._timeout_result
                    self.expired = True
                    if self._ignore:
                        return True
                    else:
                        if ty is TimeoutCancellationError:
                            raise TaskTimeout(val.args[0]).with_traceback(tb) from None
                        else:
                            return False
            elif ty is None:
>               if current_clock > self._deadlines[-1]:
E               TypeError: '>' not supported between instances of 'float' and 'NoneType'
    ```
</details>

Moreover, I tried to find a maximum value of timeout, which do not cause an error. The value I found is `2**26-1` == `67108863`. I thought about that and decided to choose any significant time interval less than `2**26-1`, for example one day

httpcore/_backends/curio.py
cdeler: By `Union` I wanted to underline that the socket could be either `SSLSocket` or `socket.socket`

httpcore/_backends/curio.py
florimondmanca: Yes, I know eg trio doesn't allow `None` and requires users to pass `float("inf")` to mean "don't ever timeout", so I was curious if this was supported by `curio`, so that we'd do:

```python
    return float("inf") if value is None else int(value)

```

Also, wait… Should this be `float(value)`? 🤔🤔🤔

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    socket = curio.io.Socket(ssl_context.wrap_socket(sock._socket, **kwargs))
+    await socket.do_handshake()
+
+    return socket
+
+
+def convert_timeout(value: Optional[float]) -> int:
+    return int(value) if value is not None else ONE_DAY_IN_SECONDS
```
florimondmanca, line 45: Right? 🤔

```suggestion
def convert_timeout(value: Optional[float]) -> float:
    return float(value) if value is not None else ONE_DAY_IN_SECONDS
```

cdeler: No, as you write above, `curio` do not accept `float` as a timeout, so it's `int`

httpcore/_backends/curio.py
cdeler: Have done it

httpcore/_backends/curio.py
cdeler: Removed `wrap_ssl_client`, and done the suggested changes.

By the way I set `do_handshake_on_connect = False` as we do the handshake right after the socket creation

httpcore/_backends/curio.py
cdeler: Yes, kwargs wasn't a good idea here :-) changed

httpcore/_backends/curio.py
florimondmanca: It actually _does_ accept float numbers — the error message in response to using `float("inf")` is misleading.

```python
import curio

curio.run(curio.sleep(1.5))  # Sleeps to 1.5 seconds
```

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,203 @@
...
+        return self._semaphore
+
+    async def acquire(self, timeout: float = None) -> None:
+        exc_map: dict = {
+            curio.TaskTimeout: self.exc_class,
+        }
+        acquire_timeout: int = convert_timeout(timeout)
+
+        with map_exceptions(exc_map):
+            return await curio.timeout_after(acquire_timeout, self.semaphore.acquire())
```
florimondmanca, line 59: (Nit) Using a simpler `try/except` style here since there's only one exception mapping case, and this is also what we do in the `asyncio` backend.

Also once `convert_timeout()` is changed to return a `float`, it looks like we'll be able to drop the `int` type hint (which I think is not actually necessary even now?), and use the same variable name (since `timeout` will remain a `float`).

```suggestion
        timeout = convert_timeout(timeout)

        try:
            return await curio.timeout_after(timeout, self.semaphore.acquire())
        except curio.TaskTimeout:
            raise self.exc_class()
```

florimondmanca (commented):
Last small comments! :tada:

httpcore/_backends/curio.py
cdeler: You are right, I checked it again. Changing it

_For me_: first check, then say, not vice versa

httpcore/_backends/curio.py
cdeler: Good point, going to change it

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,202 @@
...
+    ReadTimeout,
+    WriteError,
+    WriteTimeout,
+    map_exceptions,
+)
+from .._types import TimeoutDict
+from .._utils import get_logger
+from .base import AsyncBackend, AsyncLock, AsyncSemaphore, AsyncSocketStream
+
+logger = get_logger("curio_backend")
```
florimondmanca, line 21: The standard practice for loggers seems to be just using `__name__` (which here will be something like `httpcore._backends.curio`). This way, any logging setup on a parent module (here `httpcore`) is used automatically.

So this should be:

```suggestion
logger = get_logger(__name__)
```

httpcore/_backends/curio.py
```diff
@@ -0,0 +1,202 @@
...
+            return SocketStream(sock)
+
+    def create_lock(self) -> AsyncLock:
+        return Lock()
+
+    def create_semaphore(self, max_value: int, exc_class: type) -> AsyncSemaphore:
+        return Semaphore(max_value, exc_class)
+
+    async def time(self) -> float:
+        return float(await curio.clock())
```
florimondmanca, line 202: (Nit that I'll leave since I have another one already!) `curio.clock()` already returns a `float`:

```suggestion
        return await curio.clock()
```

florimondmanca (approved):
I think we're mostly there! I found a small thing wrt the logger name, and then another small nit sneaked in. Once those are fixed, we should be good to go!

Thank you so much for staying on top of this PR! Hope you don't mind the back and forth — having added trio via https://github.com/encode/httpx#276, it seems it's actually a natural part of adding a backend (which is really mostly a one-off thing, so we might as well get it all right first time!).

florimondmanca (approved):
Stellar, thank you! 🌟

PR #170, opened 2020-08-25
Making release() and time() async

cdeler:
In https://github.com/encode/httpcore/pull/169 and #168 you can find pretty the same changes:
`AsyncBackend::time` , `AsyncSemaphore::release` and `AsyncLock::release` have been made async.

I extracted these changes into the separated PR to make the changeset be discussed here.

### DIFF
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -25,7 +25,7 @@ class NullSemaphore(AsyncSemaphore):
     async def acquire(self, timeout: float = None) -> None:
         return
 
-    def release(self) -> None:
+    async def release(self) -> None:
         return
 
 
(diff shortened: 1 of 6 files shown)


tomchristie (approved):
Yup, I think this'll be fairly uncontroversial.
Wouldn't mind a second review over it, but otherwise I'm good with this.

Note to all: We need this because curio has different expectations (for perfectly decent reasons) about which methods should/shouldn't expose an async interface. Given that this async interface can still call into a sync implementation for `trio` and `asyncio`, but that the converse would not be true, we're not really losing anything by adopting curio's constraints in our own internal backend API here.

florimondmanca (approved):
Looks good, thank you!

PR #171, opened 2020-08-29
Allowed expiring connections where idle time == expiration time

agronholm:
On uvloop, the timer is only accurate to a millisecond. This negatively affects the `test_connection_pool_get_connection_info` test which may fail because the code runs so fast that the check is done on the same millisecond the connections have expired. This should have no practical implications elsewhere.


tomchristie (commented):
Yup, good call, tho missing the `_sync` change.

agronholm:
Oops, I'm not used to this workflow yet.

florimondmanca (approved):

PR #175, opened 2020-09-05
Mention curio in supported async backends

florimondmanca:
Cleanup for #168

PR #176, opened 2020-09-05
Add local_address support for curio

florimondmanca:
Realized that #175 did not make use of the `local_address` parameter (refs #134).

Relevant part of the curio docs is here: https://curio.readthedocs.io/en/latest/reference.html#open_connection

> _source_addr_ specifies the source address to use on the socket.

### DIFF
--- a/httpcore/_backends/curio.py
+++ b/httpcore/_backends/curio.py
@@ -155,9 +155,13 @@ class CurioBackend(AsyncBackend):
             OSError: ConnectError,
         }
         host = hostname.decode("ascii")
-        kwargs = (
-            {} if ssl_context is None else {"ssl": ssl_context, "server_hostname": host}
-        )
+
+        kwargs: dict = {}
+        if ssl_context is not None:
+            kwargs["ssl"] = ssl_context
+            kwargs["server_hostname"] = host
+        if local_address is not None:
+            kwargs["source_addr"] = (local_address, 0)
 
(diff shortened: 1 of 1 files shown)


tests/sync_tests/test_interfaces.py
```diff
@@ -211,8 +212,10 @@ def test_http_proxy(proxy_server: URL, proxy_mode: str) -> None:
 
 
 
-# This doesn't run with trio, since trio doesn't support local_address.
 def test_http_request_local_address() -> None:
+    if current_async_library() == "trio":
+        pytest.skip("trio does not support local_address.")
+
     with httpcore.SyncConnectionPool(local_address="0.0.0.0") as http:
```
florimondmanca, line 219: Note: as I mentioned in https://github.com/encode/httpcore/pull/134#discussion_r465971408, this test doesn't actually _assert_ anything / verify that `local_address` had any effect. And even after some investigation, I'm not sure there's a way to test that cross-platform and cross-machine. Eg it seems I can only bind on `10.6.2.175` locally. But that's what gets selected when using `0.0.0.0` anyway, so passing it as `local_address` wouldn't prove that `local_address` has any effect either…

florimondmanca:
Dropped the dependency of this PR on #177, since we can add the test assertion separately. This PR should be reviewable and mergeable now!

cdeler (approved):

PR #177, opened 2020-09-05
Add assertion in local_address test

florimondmanca:
Refs https://github.com/encode/httpcore/pull/134#discussion_r465971408

Our `local_address` test didn't actually verify that `local_address` was used to bind the client socket.

Scratched my head a little, and after some investigation I found out that using `local_address=<local machine IP>` seems to be a repeatable pattern that we can assert against (though we can't know for sure what that IP will look like exactly).

### DIFF
--- a/tests/async_tests/test_interfaces.py
+++ b/tests/async_tests/test_interfaces.py
@@ -214,11 +214,11 @@ async def test_http_proxy(proxy_server: URL, proxy_mode: str) -> None:
 @pytest.mark.asyncio
 # This doesn't run with trio, since trio doesn't support local_address.
 async def test_http_request_local_address() -> None:
-    # By default systems seem to bind to a private IP like 10.x.x.x.
-    # We forcefully use the system local IP, something like 192.168.x.x,
-    # to be able to distinguish that local_address is correctly applied.
+    # Forcefully use the system local IP, instead of a randomly assigned
+    # client IP address. Its shape may differ based on the type of network we're
+    # connected to (eg 10.x.x.x, 192.168.x.x, or 172.20.10.x) so we can't make any
+    # assumption there.
     local_address = get_local_ip_address()
-    assert local_address.startswith("192.168.")
 
     async with httpcore.AsyncConnectionPool(local_address=local_address) as http:
         method = b"GET"
--- a/tests/async_tests/test_interfaces.py
+++ b/tests/async_tests/test_interfaces.py
@@ -6,6 +6,7 @@ import pytest
 import httpcore
 from httpcore._types import URL
 from tests.conftest import Server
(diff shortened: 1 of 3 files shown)


tests/async_tests/test_interfaces.py
```diff
@@ -213,7 +214,13 @@ async def test_http_proxy(proxy_server: URL, proxy_mode: str) -> None:
 @pytest.mark.asyncio
 # This doesn't run with trio, since trio doesn't support local_address.
 async def test_http_request_local_address() -> None:
-    async with httpcore.AsyncConnectionPool(local_address="0.0.0.0") as http:
+    # Forcefully use the system local IP, instead of a randomly assigned
```
JayH5, line 217: I think there's a bit of an overlap of terms here. All of 10.x.x.x, 192.168.x.x, and 172.20.10.x are not system local IP ranges, they're private IP ranges. But OTOH we call the argument `local_address`, so I'm not sure 🤔 Just seems strange to me that `get_local_ip_address` returns the address of the current/"local" machine, but not necessarily a "local IP address".

I don't think it's true to say that "Its shape may differ based on the type of network we're connected to", most likely it just depends on how `socket.gethostname()` is resolved, possibly depending on the contents of `/etc/hosts`.

florimondmanca: @JayH5 Hmm I see… So the comment is misguiding, but what do you think about the general approach? What should we name `get_local_ip_address`? The goal is to bind to a known bindable IP on the machine, and verify that the socket was effectively bound that IP instead of any random bindable IP. (At least on macOS, it seems that the default behavior is to bind to a random one, rather than eg `127.0.0.1`.)

JayH5: > what do you think about the general approach?

I don't think it's going to work 😞. This test doesn't currently pass on my local machine because it can't bind to a port at `127.0.0.1` that could reach `example.org:80`. The only way I would see this working is if there are two IP addresses on the specific host that can be used to reach an external address like `example.org`, and that's not something that can really be relied on.

I think we need a local test server (#109). Then one could bind the server to localhost/127.0.0.1 and then it's much more likely that the server will be reachable from multiple "local" IP addresses. Generally, you can then reach the server from _any_ of the "local" IP addresses, whether that's the address for `lo0`/`eth0`/`en0`/whatever. (This is probably also not true 100% of the time on all systems with all configs but I think it's a reasonable assumption in most circumstances.)

(Also please take all of this with a grain of salt as I don't claim to be a networking expert)

florimondmanca: Okay, hum. I see this needs more thought, and #109 is definitely one we need to start taking a look at… Going to close this for now, good to have thought this through still!

PR #179, opened 2020-09-08
Add 'backend=<str>' parameter

tomchristie:
Add support for `httpcore.AsyncConnectionPool(backend=...)`.

The default behaviour is to use:

* `httpcore.AsyncConnectionPool(backend="auto")`

Which will auto-select an appropriate backend.
The following are also valid...

* `httpcore.AsyncConnectionPool(backend="asyncio")`
* `httpcore.AsyncConnectionPool(backend="trio")`
* `httpcore.AsyncConnectionPool(backend="curio")`

One motivation here is that this PR would allow us to merge a new `AnyIO` backend, without committing ourselves to a switch. We could add `backend="anyio"` in addition to the existing `backend="auto"`.

In the sync case `backend` is currently always just `"sync"`.

### DIFF
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -1,7 +1,7 @@
 from ssl import SSLContext
 from typing import List, Optional, Tuple
 
-from .._backends.auto import AsyncLock, AsyncSocketStream, AutoBackend
+from .._backends.auto import AsyncBackend, AsyncLock, AsyncSocketStream, AutoBackend
 from .._types import URL, Headers, Origin, TimeoutDict
 from .._utils import get_logger, url_to_origin
 from .base import (
@@ -24,6 +24,7 @@ class AsyncHTTPConnection(AsyncHTTPTransport):
         ssl_context: SSLContext = None,
         socket: AsyncSocketStream = None,
         local_address: str = None,
+        backend: AsyncBackend = None,
     ):
         self.origin = origin
         self.http2 = http2
@@ -40,7 +41,7 @@ class AsyncHTTPConnection(AsyncHTTPTransport):
         self.is_http2 = False
         self.connect_failed = False
         self.expires_at: Optional[float] = None
-        self.backend = AutoBackend()
+        self.backend = AutoBackend() if backend is None else backend
 
     def __repr__(self) -> str:
         http_version = "UNKNOWN"
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -2,7 +2,8 @@ import warnings
 from ssl import SSLContext
 from typing import AsyncIterator, Callable, Dict, List, Optional, Set, Tuple
 
-from .._backends.auto import AsyncLock, AsyncSemaphore, AutoBackend
+from .._backends.auto import AsyncLock, AsyncSemaphore
+from .._backends.base import lookup_async_backend
 from .._exceptions import LocalProtocolError, PoolTimeout, UnsupportedProtocol
 from .._threadlock import ThreadLock
 from .._types import URL, Headers, Origin, TimeoutDict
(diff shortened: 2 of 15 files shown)


florimondmanca:
So looking at how things evolve if we do/don't make the switch to anyio:

- If we don't switch: we now have an explicit `backend` knob, and the only sensible scenario where users would use it would be to pass `backend="anyio"` (it's the only sensible use case, since eg running with `backend="trio"` within an `asyncio` async environment would lead to failures).
- If we do switch (and abandon our hand-made backends): we end up with an explicit `backend` know that allows `"auto"` (?) and `"anyio"`. So most likely we'll end up removing this parameter.

It's good that this allows us to _offer the possibility_ of running with `anyio` (if users have it installed), but also I'm not sure if we're not just delaying the switch too much by introducing non-necessary temporary API. Well, most likely nobody else than HTTPX is using HTTPCore right now, and as I said the knob wouldn't be used except for enabling `"anyio"` explicitly…

Another option could be: Change `AutoBackend` to use `AnyIOBackend` _if `anyio` is installed_.

Thinking out loud here…

Pros:

* Follows the same "auto-detection" principle than we already use for auto-detecting the async environment.
* Non-intrusive and non-breaking for users that _don't_ use anyio (also true for the `backend` approach).
* Better consistency for users that _do_ use `anyio` (most likely someone running an asyncio program and using `anyio` for structured concurrency APIs), since HTTPX will also use `anyio` in that case.
* Allows exposing our `AnyIOBackend` to real world usage without requiring an explicit switch / intervention from users.

Cons:

* Extra parameter to ask for when debugging, in order to know which of anyio or our backends was used. (But this is also true if we add an explicit `backend="anyio"` option.)
* Perhaps too smart? Eg what if end-users don't use `anyio`, but another dependency relies on `anyio` _internally_?

The last con is probably a big one… So, yay for a `backend` option for now?

florimondmanca (approved):

tomchristie:
Great, thank you!

My take here is that I don't think we should ever have any behaviours where we *use behaviour X if package Y is installed* in `httpx` or `httpcore`.

It introduces the possibility of things unexpectedly changing underneath user's feet. (Eg. `anyio` might not have been installed directly, but could be some indirect dependency of some other package.)

Why `backend=<str>` start to feel useful to me is for really low-level control for very specific cases. (Eg. Let's introduce an `AnyIO` backend, and ensure that our test suite works against both that and against our native backends.) It also later naturally extends into `backend=<str|AsyncBackend>` / `backend=<str|SyncBackend>`.

That'd be useful because we could then either publicly or semi-publicly expose the backend interface, in order to support some advanced use-cases:

* Allow new backends to be implemented and tested much more easily.
* Allow testing gnarly behaviours against stub backends that mock out the socket behaviour.
* Allow nifty stuff such as middleware-like backends, that either log the exact socket behaviour, or allow recording and replaying exact socket behaviours.

> I'm not sure if we're not just delaying the switch too much by introducing non-necessary temporary API

I guess I ought to dig into where I'm at on this myself...

TBH I'm in two minds wrt. if switching out our native backends for an anyio backend *will or won't be something that we want to do*. I get why it sounds attractive on the surface, but we're not really removing any complexity by doing that, we're just submerging it into another layer. That's not a hard *"no"* from me, it's more an *"I'm not sure"*. Also there's some different places we might end up in here depending on how the landscape evolves.

What's more concrete at this point is that I *certainly* wouldn't be comfortable making any switch until we've got a much more comprehensive test suite - it's relatively high risk given the small handful of gnarly edge case `asyncio` bugs that we're working around in places.

What makes sense to me at this point is getting `AnyIOBackend` in, making sure we're testing comprehensively against both that and the existing behaviours, working on really expanding our our test suite for `httpcore` to cover all the gnarly stuff, and being able to feedback into `anyio` while we're doing so. I'd suggest we probably want to wait until we've got 100% coverage before we assess if any switch makes sense.

PR #180, opened 2020-09-09
Add support for `backend="anyio"`

tomchristie:
Builds on #169, with a minimal as possible change footprint.

Last hurdle to clear would be switching out the cases where I've used `backend="anyio"` to be parameterized to use both `backend="auto"|"anyio"`. The `unasync.py` script will need to strip out that parameterisation for the sync case, which needs to just always use `backend="sync"`.

I'm really keen on us also using anyio to slim down any gnarly in the test suite that it can help us address, but we should treat that as an entirely isolated issue. (Using `anyio` to run our tests is an independent concern to what backend implementation we want to use.)

### DIFF
--- a/httpcore/_backends/anyio.py
+++ b/httpcore/_backends/anyio.py
@@ -1,0 +1,198 @@
+import select
+from ssl import SSLContext
+from typing import Optional
+
+import anyio.abc
+from anyio import BrokenResourceError, EndOfStream
+from anyio.abc import SocketAttribute
+from anyio.streams.tls import TLSAttribute, TLSStream
+
+from .._exceptions import (
+    CloseError,
+    ConnectError,
+    ConnectTimeout,
+    ReadError,
+    ReadTimeout,
+    WriteError,
+    WriteTimeout,
+)
+from .._types import TimeoutDict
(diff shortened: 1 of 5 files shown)


tomchristie:
Also, I think this change bumps up our minimum required `trio` version, which we're currently just being lax about.

(I happened to be running locally with trio 0.14.0, which passed our existing test suite, but failed with `backend="anyio"`, but upgrading to trio 0.16.0 resolved this.)

agronholm:
This won't test the explicit backends used by the "auto" backend though, does it?

tomchristie:
> This won't test the explicit backends used by the "auto" backend though, does it?

At this point in time, yup. It's *only* testing `"anyio"`, eg...

```python
async def test_http_request() -> None:
    async with httpcore.AsyncConnectionPool(backend="anyio") as http:  # Explicitly testing against anyio only.
        method = b"GET"
        url = (b"http", b"example.org", 80, b"/")
```

The last step here would be to parameterise it so that both `backend="anyio"` and `backend="auto"` end up being tested. (And making a little nudge in `unasync.py` so that the sync case is only parametereized with a single case, for `backend="sync"`)

florimondmanca:
Closing tentatively since it looks like the suggestions from this PR were integrated into #169. :-)

PR #181, opened 2020-09-09
Replaced other test plugins and harnesses with AnyIO

agronholm:


agronholm:
I'm having some trouble with `test_http_request_local_address`. I'm supposed to skip it on trio, but I'm having difficulties because unasync is interfering with my every effort.

tests/async_tests/test_interfaces.py
```diff
@@ -2,11 +2,14 @@
 import ssl
 
 import pytest
+import sniffio
 
 import httpcore
 from httpcore._types import URL
 from tests.conftest import Server, detect_backend
 
+pytestmark = pytest.mark.anyio
```
florimondmanca, line 11: I guess we'd rather keep the decorator style, with `@pytest.mark.anyio`? I know in the past we ended up eg having some case that are fine running on any one async environment (eg just asyncio), and in that case we could just switch to `@pytest.mark.asyncio`. (I also _personally_ prefer this "decorate every test case explicitly" for its, well, explicitness…)

tests/async_tests/test_interfaces.py
```diff
@@ -210,9 +204,14 @@ async def test_http_proxy(proxy_server: URL, proxy_mode: str) -> None:
...
 
-@pytest.mark.asyncio
-# This doesn't run with trio, since trio doesn't support local_address.
 async def test_http_request_local_address() -> None:
+    exc = getattr(sniffio, "Async" + "LibraryNotFoundError")
+    try:
+        if sniffio.current_async_library() == "trio":
+            pytest.skip("The trio backend does not support local_address")
+    except exc:
+        pass
```
florimondmanca, line 213: One way to make this work would be to use the same trick than `lookup_async_backend()`/`lookup_sync_backend()` that we already have, e.g. have these helpers:

```python
# tests/utils.py
def lookup_async_library():
    return sniffio.current_async_library()

def lookup_sync_library():
    return "sync"
```

Then in async test cases:

```suggestion
    if lookup_async_library() == "trio":
        pytest.skip("The trio backend does not support local_address")
```

Add `("lookup_async_library", "lookup_sync_library")` to `unasync.py`, and things _should_ fall into place nicely?

florimondmanca (approved):
Great stuff — using `anyio` for testing purposes only is a great idea _regardless_ of what happens wrt backends. :-)

tests/async_tests/test_interfaces.py
agronholm: Switched back to the decorator style.

tests/async_tests/test_interfaces.py
agronholm: I added the module you suggested and changed the test accordingly.

florimondmanca (approved):
💯

tomchristie:
Lovely stuff. 👍

PR #183, opened 2020-09-11
Ignored a new error from updated flake8-pie

agronholm:


StephenBrown2:
See discussion in https://github.com/encode/httpx/pull/1283.
Completely ignoring a lint should be done with careful thought and agreement. I for one would rather ignore it where it's necessary, rather than across the entire project, but others in @encode/maintainers need to chime in as well.

agronholm:
Only 4 lines are affected so I could do that too.

florimondmanca (approved):

florimondmanca (approved):
I'm merging this to unblock other PRs, we can reassess later whether we're okay keeping inline `# noqa`, or if we'd like to just ignore this error in `setup.cfg`, or just pin to the previous version. :+1:

tomchristie:
Just to mention that I really do think it's worth ignoring this in `setup.cfg`. I wouldn't have an issue with this if it only failed linting on `except BaseException`, but `except Exception` is a *perfectly* normal sensible thing to do, and shouldn't need per-line linter exclusions. 👍 

We *could* also review all our `except Exception` cases in tandem with that change, in order to assess if any of them *could* be more tightly defined.

PR #184, opened 2020-09-13
Pin dependencies in requirements.txt and add Dependabot

florimondmanca:
Alternative to #162 

This PR pins all entries in our current `requirements.txt`.

Does not include a freeze process, so that we don't have to deal with transitive dependencies cross-platform (see https://github.com/encode/httpcore/pull/162#issuecomment-691639992).

This way this would work is:

* Developers and CI install from `requirements.txt`. Top-level dependencies are pinned, underlying dependencies may change. We assume the pip 2020 resolver will always be able to find transitive dependencies that work. Should work on Linux, macOS, Windows, and any supported Python version.
* Dependabot comes in once a week to check in and update any top-level dev dependencies that need updating.

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -15,5 +15,6 @@ else
     PIP="pip"
 fi
 
+"$PIP" install -U "pip >= 20.2" setuptools wheel
 "$PIP" install -r "$REQUIREMENTS"
 "$PIP" install -e .
--- a/scripts/install
+++ b/scripts/install
@@ -17,4 +17,3 @@ fi
 
 "$PIP" install -U "pip >= 20.2" setuptools wheel
 "$PIP" install -r "$REQUIREMENTS"
-"$PIP" install -e .
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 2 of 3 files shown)


scripts/install
```diff
@@ -15,5 +15,6 @@ else
     PIP="pip"
 fi
 
+"$PIP" install -U "pip >= 20.2" setuptools wheel
 "$PIP" install -r "$REQUIREMENTS"
 "$PIP" install -e .
```
j178, line 20: We can remove this line since `-e .[http2]` has been included in requirements.txt .

florimondmanca: Good point, though it's actually the case not even considering this PR :-)

```suggestion
```

j178 (approved):

tomchristie:
Right, that's interesting.

It's not *actually* as strictly frozen as #162 - we could still get bitten by a transitive dependancy updating, *but* practically it's probably actually enough for us almost always, and it's nice and simple this ways.

🤷‍♂️

florimondmanca:
@tomchristie Which do you consider simpler? In my mind this PR is a simpler approach at first - at least were explicitly pinning what were using. Even if transitive deps might shift, all "sudden CI breaking" issues I've personally had in the past were due to top level dependencies shifting with breaking changes, not transitive dependencies.

tomchristie:
Indeed yup. Let's roll with this, lovely and simple.

We could always reassess another day if we needed.

StephenBrown2 (approved):
👍

PR #185, opened 2020-09-13
Tweak dropped connection detection

### LINKED ISSUE #182: Using `select` is causing issues when many files are open
In the internals of the httpcore sync client, there is a call to select.select that fails when there is not a fd available < 1024. This can occur when there are already many open files prior to using httpx (and thereby httpcore).  A similar problem occurred in urllib3 a while ago (see references below).

References: 
- https://github.com/boto/botocore/issues/756
- https://github.com/urllib3/urllib3/issues/589

Python Verson: 3.6.8

Traceback:
```
...
File "/var/starterview/common/py/http_client/httpx.py", line 1440, in post
    json=json)
  File "/var/starterview/common/py/http_client/httpx.py", line 379, in do_sync
    resp = func(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/httpx/_client.py", line 932, in post
    timeout=timeout,
  File "/usr/local/lib/python3.6/dist-packages/httpx/_client.py", line 674, in request
    request, auth=auth, allow_redirects=allow_redirects, timeout=timeout
  File "/var/starterview/common/py/http_client/httpx.py", line 640, in wrapper
    resp = func(request, *args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/httpx/_client.py", line 704, in send
    request, auth=auth, timeout=timeout, allow_redirects=allow_redirects
  File "/usr/local/lib/python3.6/dist-packages/httpx/_client.py", line 733, in _send_handling_redirects
    request, auth=auth, timeout=timeout, history=history
  File "/usr/local/lib/python3.6/dist-packages/httpx/_client.py", line 769, in _send_handling_auth
    response = self._send_single_request(request, timeout)
  File "/usr/local/lib/python3.6/dist-packages/httpx/_client.py", line 803, in _send_single_request
    timeout=timeout.as_dict(),
  File "/usr/local/lib/python3.6/dist-packages/httpcore/_sync/connection_pool.py", line 172, in request
    connection = self._get_connection_from_pool(origin)
  File "/usr/local/lib/python3.6/dist-packages/httpcore/_sync/connection_pool.py", line 217, in _get_connection_from_pool
    if connection.is_connection_dropped():
  File "/usr/local/lib/python3.6/dist-packages/httpcore/_sync/connection.py", line 148, in is_connection_dropped
    return self.connection is not None and self.connection.is_connection_dropped()
  File "/usr/local/lib/python3.6/dist-packages/httpcore/_sync/http11.py", line 193, in is_connection_dropped
    return self.socket.is_connection_dropped()
  File "/usr/local/lib/python3.6/dist-packages/httpcore/_backends/sync.py", line 81, in is_connection_dropped
    rready, _wready, _xready = select.select([self.sock], [], [], 0)
ValueError: filedescriptor out of range in select()```

### PR DESCRIPTION
**Pull request by florimondmanca**
Closes #182 

urllib3 had to deal with an issue similar to #182 a while ago: https://github.com/urllib3/urllib3/issues/589

Their approach was to backport a bunch of 3.5+ logic to maintain 2.7 compatibility, but we're lucky enough to be able to use that logic from the stdlib in the form of the [`selectors`](https://docs.python.org/3/library/selectors.html) module, a "higher-level" alternative to `select` that handles cross-platform subtleties among other things.

This PR switches the sync backend to use `selectors.select()` instead of `select.select()` for detecting whether a connection was dropped (i.e. whether the underlying socket has become immediately readable).

Confirmed that this solves #182, since the CI build passes here but fails on #219.

### DIFF
--- a/httpcore/_backends/sync.py
+++ b/httpcore/_backends/sync.py
@@ -77,7 +77,7 @@ class SyncSocketStream:
             with map_exceptions({socket.error: CloseError}):
                 self.sock.close()
 
-    def is_connection_dropped(self) -> bool:
+    def is_readable(self) -> bool:
         return is_socket_readable(self.sock.fileno())
 
 
--- a/httpcore/_backends/sync.py
+++ b/httpcore/_backends/sync.py
@@ -1,4 +1,3 @@
-import select
 import socket
 import threading
 import time
@@ -17,6 +16,7 @@ from .._exceptions import (
     map_exceptions,
 )
 from .._types import TimeoutDict
+from .._utils import wait_for_read
 
 
 class SyncSocketStream:
@@ -78,8 +78,8 @@ class SyncSocketStream:
                 self.sock.close()
 
     def is_connection_dropped(self) -> bool:
-        rready, _wready, _xready = select.select([self.sock], [], [], 0)
-        return bool(rready)
+        rready = wait_for_read([self.sock], timeout=0)
+        return len(rready) > 0
 
 
 class SyncLock:
--- a/httpcore/_backends/sync.py
+++ b/httpcore/_backends/sync.py
@@ -16,7 +16,7 @@ from .._exceptions import (
     map_exceptions,
 )
 from .._types import TimeoutDict
-from .._utils import wait_for_read
+from .._utils import is_socket_at_eof
 
 
 class SyncSocketStream:
@@ -78,8 +78,7 @@ class SyncSocketStream:
                 self.sock.close()
 
     def is_connection_dropped(self) -> bool:
-        rready = wait_for_read([self.sock], timeout=0)
-        return len(rready) > 0
+        return is_socket_at_eof(self.sock.fileno())
 
 
 class SyncLock:
--- a/httpcore/_backends/sync.py
+++ b/httpcore/_backends/sync.py
@@ -16,7 +16,7 @@ from .._exceptions import (
     map_exceptions,
 )
 from .._types import TimeoutDict
-from .._utils import is_socket_at_eof
+from .._utils import is_socket_readable
 
 
 class SyncSocketStream:
@@ -78,7 +78,7 @@ class SyncSocketStream:
                 self.sock.close()
 
     def is_connection_dropped(self) -> bool:
-        return is_socket_at_eof(self.sock.fileno())
+        return is_socket_readable(self.sock.fileno())
 
 
 class SyncLock:
--- a/httpcore/_utils.py
+++ b/httpcore/_utils.py
@@ -66,8 +66,9 @@ def origin_to_url_string(origin: Origin) -> str:
     return f"{scheme.decode('ascii')}://{host.decode('ascii')}{port}"
 
 
-def _wait_io_events(socks: list, events: int, timeout: float = None) -> list:
-    # Better cross-platform approach than low-level `select.select()` calls.
+def _wait_for_io_events(socks: list, events: int, timeout: float = None) -> list:
+    # Prefer the `selectors` module rather than the lower-level `select` module to
+    # improve cross-platform support.
     # See: https://github.com/encode/httpcore/issues/182
     sel = selectors.DefaultSelector()
     for sock in socks:
@@ -76,4 +77,4 @@ def _wait_io_events(socks: list, events: int, timeout: float = None) -> list:
 
 
 def wait_for_read(socks: list, timeout: float = None) -> list:
-    return _wait_io_events(socks, events=selectors.EVENT_READ, timeout=timeout)
+    return _wait_for_io_events(socks, events=selectors.EVENT_READ, timeout=timeout)
--- a/httpcore/_utils.py
+++ b/httpcore/_utils.py
@@ -66,15 +66,18 @@ def origin_to_url_string(origin: Origin) -> str:
     return f"{scheme.decode('ascii')}://{host.decode('ascii')}{port}"
 
 
-def _wait_for_io_events(socks: list, events: int, timeout: float = None) -> list:
-    # Prefer the `selectors` module rather than the lower-level `select` module to
-    # improve cross-platform support.
+def is_socket_at_eof(sock_fd: int) -> bool:
+    # 'Has the socket reached EOF?' is equivalent to 'Is the socket readable?'
+    # This is because if the other end has dropped the connection,
+    # a recv() call on the socket would return immediately (with empty bytes), and
+    # vice versa.
+    # See: https://github.com/encode/httpx/pull/143#issuecomment-515181778
+    # Typically we'd use the `select` module here, but we use the higher-level
+    # `selectors` module to improve cross-platform support.
     # See: https://github.com/encode/httpcore/issues/182
     sel = selectors.DefaultSelector()
-    for sock in socks:
-        sel.register(sock, events)
-    return [key.fileobj for key, mask in sel.select(timeout) if mask & events]
-
-
-def wait_for_read(socks: list, timeout: float = None) -> list:
-    return _wait_for_io_events(socks, events=selectors.EVENT_READ, timeout=timeout)
+    sel.register(sock_fd, selectors.EVENT_READ)
+    read_ready = [
+        key.fileobj for key, mask in sel.select(0) if mask & selectors.EVENT_READ
+    ]
+    return len(read_ready) > 0
--- a/httpcore/_utils.py
+++ b/httpcore/_utils.py
@@ -1,7 +1,7 @@
 import itertools
 import logging
 import os
-import selectors
+import select
 import sys
 import typing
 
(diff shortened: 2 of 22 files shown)

### CODE REVIEW

**Comment by florimondmanca** [comment]
From my own testing, this updated implementation seems to work fine:

Server:

```python
import uvicorn
from starlette.responses import PlainTextResponse

app = PlainTextResponse("Hello, World!")

if __name__ == "__main__":
    # Make it so that idle client connections are shut down after 3s.
    uvicorn.run(app, timeout_keep_alive=3)
```

Client script (exits successfully)

```python
import time
import httpcore

with httpcore.SyncConnectionPool() as transport:
    method = b"HEAD"
    url = (b"http", b"localhost", 8000, b"/")
    headers = [(b"host", b"localhost")]
    *_, stream = transport.request(method, url, headers)

    try:
        # During the first two seconds, the connection remains open…
        for __ in range(2):
            time.sleep(1)
            assert not stream.connection.is_connection_dropped()

        # After 3s, Uvicorn shuts down the connection, and we detect it.
        time.sleep(1)
        assert stream.connection.is_connection_dropped()

    finally:
        stream.close()

```

httpcore/_backends/sync.py
```diff
@@ -78,8 +78,8 @@ def close(self) -> None:
                 self.sock.close()
 
     def is_connection_dropped(self) -> bool:
-        rready, _wready, _xready = select.select([self.sock], [], [], 0)
-        return bool(rready)
+        rready = wait_for_read([self.sock], timeout=0)
```
cdeler, line 81: May you please add the same mechanism to the `curio` backend?

https://github.com/encode/httpcore/blob/685a9b6c968f0662b45f637ffacaf4795454fb9d/httpcore/_backends/curio.py#L135-L138

tomchristie:
Seems like a clear win yup.

We'll probably also want some follow ups, eg... if no selector exists, then our graceful-ish fallback can just be never detecting connections as expired, and just relying on the keepalive expiry.

There might be some neater behaviour we can have for detecting expiry, at the point we're *actually* making the request. (Eg. attempting a read with a minimal timeout prior to writing any data, in order to see any EOF if it exists?....)

agronholm:
I find it worrying that there is no test that verifies that dropped connections are actually detected. The test suite passes just fine if I unconditionally return `False`.

tomchristie:
It'll come.
We've had to cover a *lot* of ground to get here.

agronholm:
Anyway, I did some manual testing and found that `sock.recv(1, MSG_PEEK)` does detect disconnection just as well as `select.select()`:
- If the connection is still alive, `sock.recv(1, MSG_PEEK)` raises `BlockingIOError`
- If the connection has been dropped, `sock.recv(1, MSG_PEEK)` return `b''`
- If there is still buffer in the data, neither `sock.recv()` or `selectors.select()` can distinguish that from a ddetect if the connection has been dropped
Given that `selectors.select()` will use `select.select()` on Windows anyway, wouldn't it make sense to use `sock.recv(1, MSG_PEEK)` since it works on all platforms (I checked on Linux, Windows and macOS)?

florimondmanca:
> wouldn't it make sense to use sock.recv(1, MSG_PEEK) since it works on all platforms (I checked on Linux, Windows and macOS)?

Pretty interesting, indeed…

```python
>>> import socket
>>> alice, bob = socket.socketpair(socket.AF_UNIX, socket.SOCK_STREAM)
>>> alice.setblocking(False)
>>> bob.setblocking(False)
>>> alice.recv(1, socket.MSG_PEEK)
BlockingIOError: [Errno 35] Resource temporarily unavailable
>>> bob.send(b"hello")
>>> alice.recv(5)
b'hello'
>>> bob.close()  # Server connection lost
>>> alice.recv(1, socket.MSG_PEEK)
b''
```

florimondmanca:
Following on @tomchristie's comment, perhaps this `MSG_PEEK` trick could be used when there's no selector available…?

Also, it's not clear to me yet whether #182 would surface on Windows. Eg in https://github.com/urllib3/urllib3/issues/589 @sethmlarson wrote:

> On non-Windows platforms (or any platform with a better selector than `select.select`) this will be fixed with <PR>

I'm not sure whether this implies that the problem from #182 isn't present on Windows, or if Seth meant that it would be solved with another urllib3 PR at the time.

florimondmanca:
@agronholm Since you mentioned testing on Windows, were you / would you be able to reproduce the issue in #182 on Windows? If we can't reproduce there, then perhaps the safest bet would be to just not deal with it at the moment, issue a fix for Unix, and then deal with any upcoming issues for Windows should they arrive?

agronholm:
It just requires a lot of established connections within a single process, right? That shouldn't be hard to do.

tomchristie:
It looks like MSG_PEEK is possibly a better approach anyways, seems to be available throughout, and less likely to have platform specific differences vs. working with `selector`?

tomchristie:
Incidentally we might want to rename `is_connection_dropped` to `is_at_eof` since it better indicates exactly what we mean.

If there's data still in the buffer we *want* the interface to return False. (We'll most likely end up with a protocol error in that odd case, at least for HTTP/1.1, but the point is that reading on the socket isn't at EOF, so we should go ahead an make the request)

florimondmanca:
I guess we could roll with `MSG_PEEK` and see if it explodes in our hands? 😅

tomchristie:
I'm all for it - it really does look like the actually most sensible thing to do.

florimondmanca:
Hitting some road blocks on trying to implement this `.recv()` idea…

* When running `anyio+asyncio+uvloop`, the raw socket is a uvloop [`PseudoSocket`](https://www.google.com/search?client=firefox-b-d&q=uvloop+transport+sockets+do+not+support) that stubs out `recv()` calls (among other methods…)

```console
E   TypeError: transport sockets do not support recv() method
```

* In the sync case, the raw socket is blocking, so if the connection is alive, `sock.recv(1, MSG_PEEK)` will block waiting for the other end to send a piece of data (instead of failing immediately). Moving to a non-blocking usage here would mean implementing some infinite loops for read/write operations, which sounds pretty quirky.

Filed https://github.com/encode/httpcore/pull/193 if folks would like to see the code.

I'm going to proceed and see if the `selectors` tweak works with sync, `anyio` and `curio`… If it does, then I think we should probably just roll with it.

agronholm:
> When running anyio+asyncio+uvloop, the raw socket is a uvloop PseudoSocket that stubs out recv() calls (among other methods…)

I recall running into this myself. If memory serves, I got the `fileno()` out of it and then used `socket.fromfd()` to get a real socket object.

> I'm going to proceed and see if the selectors tweak works with sync, anyio and curio… If it does, then I think we should probably just roll with it.

I'll remind you that this does not fix the problem on Windows.

florimondmanca:
> `fileno()`

Yep, that's what I actually ended up using in this PR too… Updated #193 as well — looking _quite_ good now!

florimondmanca:
Closing in favor of #193, which I think is getting in good shape now!

florimondmanca:
I am reviving this PR based on the insights from Nathaniel in https://github.com/encode/httpcore/pull/193#issuecomment-703129316

TL;DR: this should only be a problem on Unix, so the `selectors` module is just what we need. (Windows `select()` is different than on Unix, and shouldn't be exposed to the issue reported in #182. Besides, #182 is indeed _about Linux_, not Windows.)

florimondmanca:
Added a test case based on #219.

Should be ready for re-review! cc @JayH5 @cdeler

httpcore/_utils.py
```diff
@@ -63,3 +64,27 @@ def origin_to_url_string(origin: Origin) -> str:
...
+
+    See: https://github.com/encode/httpx/pull/143#issuecomment-515181778
+    """
+    # NOTE: We prefer the `selectors` module to `select`, because of known limitations
+    # of `select` on Linux when dealing with many open file descriptors.
+    # See: https://github.com/encode/httpcore/issues/182
+    # On Windows `select` is just fine, but that's also what `DefaultSelector` uses
+    # there, so `selectors` is really the generally-appropriate solution.
+    # See: https://github.com/encode/httpcore/pull/193#issuecomment-703129316
+    sel = selectors.DefaultSelector()
```
JayH5, line 86: This will use `epoll` on Linux. From https://github.com/encode/httpcore/pull/193#issuecomment-703129316:
> [...]we're only trying to query a single socket at a time, so the fancy modern options are actually less efficient than good ol' `poll`/`select`.

But I do appreciate that just using `DefaultSelector` makes the code simpler because we don't need to add if-statements for different platforms. So don't really mind 🤷‍♂️ 

OTOH, linking to a comment that suggests _not_ using `selectors` and then going ahead and using it anyway may be a bit confusing.

florimondmanca: Yup, fair enough, I might just go ahead and steal the trio code that does "select on Windows, poll elsewhere". :)

cdeler: The [comment](https://github.com/encode/httpcore/pull/193#issuecomment-703129316) mentioned by @JayH5 made sense.

I ran local performance tests `selectors` (I'm on Mac) vs `poll`, checking "is_socket_readable" both approaches, so `poll` was faster. After that I aimed google and found some articles like [that](https://daniel.haxx.se/docs/poll-vs-select.html), which told:
> Compared to poll and select, any event-based system will give a performance boost already with a few hundred file descriptors and then the benefit just grows the more connections you add.

So that it's reasonable to copy `trio` approach to our library

httpcore/_utils.py
```diff
@@ -63,3 +64,27 @@ def origin_to_url_string(origin: Origin) -> str:
...
     port = f":{explicit_port}" if explicit_port != DEFAULT_PORTS[scheme] else ""
     return f"{scheme.decode('ascii')}://{host.decode('ascii')}{port}"
+
+
+def is_socket_readable(sock_fd: int) -> bool:
+    """
+    Return whether a socket, as identifed by its file descriptor, is readable.
+
+    "A socket is readable" means that it would return immediately with b"" if we
+    called .recv() on it.
```
tomchristie, line 74: Is this meant to instead read as...

> "A socket is NOT readable" means that it would return immediately with b"" if we called .recv() on it.

?

florimondmanca: No, "the socket _is_ readable" is the correct (although _very_ intuitive, but that's why it stuck in my head!) way to describe things, at least if we go back to the conversation we've already had about this here…  https://github.com/encode/httpx/pull/143#issuecomment-515181778 (If you look at the comment just above that one, Seth asked the exact same question than the one you're asking now. :-))

tomchristie:
Okay, so that one thing I hadn't fully groked about this is that it *shouldn't* be called `is_connection_dropped()`, because that's *not* what it's testing for. It's testing for `is_socket_readable()`, which *in the narrow and very particular sense of a socket that's in between request/response cycles would also **just-so-happen** to indicate that it must have been dropped*.

Why?

Because the only valid data that could be about to come back from the socket at that point in the connection lifecycle is `""`.
If there's no outstanding request, then we're not expecting any response data.

But in *general* the state of `is_socket_readable()` doesn't mean that the connection is dropped.

So... I think we need to:

1. Rename the method.
2. Have a brief comment at the point in the connection pool where we call into `is_socket_readable()` indicating that if the socket is readable in this state then the only valid reason it could have for being readable is a pending disconnect.

httpcore/_utils.py
florimondmanca: Yup, swiitched to the `trio` approach!

florimondmanca:
Okay, I think this is starting to look quite good? Applied latest suggestions and addressed comments, happy to take new reviews!

JayH5 (approved):
Looks great to me 🍰

cdeler (approved):
Thats perfect for me :fire:

PR #51, opened 2020-04-01
WIP: SOCKS proxy support

yeraydiazdiaz:
Adds SOCKS4 proxy support via socksio 🎉 

You can test this either locally using Dante via Docker as [we do in socksio](https://github.com/sethmlarson/socksio/tree/master/docker) or simply running an `ssh -vND <PORT> <OTHER_MACHINE>`, then running the following script, assuming the proxy is running in `localhost:1080`:

```python
import asyncio
import httpcore

async def read_body(stream):
    try:
        body = []
        async for chunk in stream:
            body.append(chunk)
        return b"".join(body)
    finally:
        await stream.aclose()

async def socks_proxy(proxy_server, proxy_mode):
    async with httpcore.AsyncHTTPProxy(proxy_server, proxy_mode=proxy_mode) as http:
        method = b"GET"
        url = (b"http", b"93.184.216.34", 80, b"/")
        headers = [(b"host", b"example.org")]
        http_version, status_code, reason, headers, stream = await http.request(
            method, url, headers
        )
        body = await read_body(stream)
        print(body)

        assert http_version == b"HTTP/1.1"
        assert status_code == 200
        assert reason == b"OK"

def main():
    proxy_server = (b'http', b'localhost', 1080)
    proxy_mode = 'SOCKS4'
    asyncio.run(socks_proxy(proxy_server, proxy_mode))

if __name__ == '__main__':
    main()
```

Before I carry on with SOCKS4A and SOCKS5 (or possibly I should open a different PR?) I'd like some feedback.

PR #188, opened 2020-09-17
Implement socks5 proxy support using socksio

cdeler:
The final draft PR for https://github.com/encode/httpx/issues/203 (I closed previous). It's based on https://github.com/encode/httpcore/pull/51, I added @yeraydiazdiaz to co-authored-by

What I did:
1. added `pproxy` as a dev. dependency to have a local socks proxy server
1. added `AsyncSOCKSConnection` and `SyncSOCKSConnection`. For now it handles only socks5-without-auth connections

### DIFF
--- a/httpcore/__init__.py
+++ b/httpcore/__init__.py
@@ -29,9 +29,9 @@ __all__ = [
     "AsyncByteStream",
     "AsyncConnectionPool",
     "AsyncHTTPProxy",
-    "AsyncSocksProxy",
     "AsyncHTTPTransport",
     "AsyncIteratorByteStream",
+    "AsyncSocksProxy",
     "CloseError",
     "ConnectError",
     "ConnectTimeout",
@@ -48,13 +48,14 @@ __all__ = [
     "SyncByteStream",
     "SyncConnectionPool",
     "SyncHTTPProxy",
-    "SyncSocksProxy",
     "SyncHTTPTransport",
+    "SyncSocksProxy",
     "TimeoutException",
     "UnsupportedProtocol",
     "WriteError",
     "WriteTimeout",
 ]
+
 __version__ = "0.11.1"
 
 __locals = locals()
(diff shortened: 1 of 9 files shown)


httpcore/_async/connection.py
```diff
@@ -162,3 +164,76 @@ async def aclose(self) -> None:
         async with self.request_lock:
             if self.connection is not None:
                 await self.connection.aclose()
+
+
+class AsyncSOCKSConnection(AsyncHTTPConnection):
```
florimondmanca, line 169: A bit wary of the subclassing dependency here...

I understand the intent that "a SOCKS connection is just an HTTP connection, but connected differently". But since "open a socket" seems to be the *only* step that changes, and considering we have a `socket` parameter on `AsyncHTTPConnection`, would it make sense for the "open a SOCKS proxy socket" operation to be performed by an `AsyncSOCKSProxy` class?

In other words, instead of a separate connection, we'd have a new proxy connection pool class that opens a socket explicitly before instantiating and saving a connection.

Dunno how feasible that is and I haven't went through actually implementing this, and perhaps you've already thought about that, but just throwing ideas out there! :)

florimondmanca (commented):
Nice start! Curious if CI passes with this setup :eyes:

httpcore/_async/connection.py
cdeler: You mean to replace inheritance by aggregation?

florimondmanca: Sure, never mind, I see that #51 took this approach too. :-)

cdeler:
Well, I mostly implemented `socks5` proxy support. Now I want to think up what to do with `socks4`

cdeler:
Converted back to draft until socks4 proxy support has been added (then we can split the PR into chunks)

cdeler:
I reworked this draft PR according to the @tomchristie 's comment: https://gitter.im/encode/community?at=5f6db8edf41f4105e4d7a9eb

So that right now `(A)syncSocksProxy`  is the wrapper over the `(A)syncConnectionPool`

The next step I want to do is to merge https://github.com/encode/httpcore/pull/199 (it's a PR about migrating `http` proxy suite from `mitmproxy` to `pproxy`)

tomchristie:
Now supported in 0.14.5 - https://www.encode.io/httpcore/proxies/#socks-proxy-support

PR #189, opened 2020-09-21
Use `.arequest` for the async version of the API.

tomchristie:
In order to support both sync and async styles on a *single* implementation class, let's move the async request method into a different namespace.

See https://github.com/encode/httpx/issues/1274

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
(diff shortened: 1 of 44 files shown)


florimondmanca (approved):
Brilliant yup, just want to confirm that this is an initial step towards eventually having a single `httpcore.HTTPTransport` interface with both `.request()` and `.arequest()` on it?

tomchristie:
Yeah, although more likely `httpx.HTTPTransport`.

PR #192, opened 2020-09-22
Drop incorrect curio classifier

tomchristie:


florimondmanca:
Ah, no PyPI trove classifier for Curio yet? Shame... 😅

PR #193, opened 2020-09-22
Detect broken connections using socket peeking

### LINKED ISSUE #182: Using `select` is causing issues when many files are open

florimondmanca:
Fixes #182 

Attempts implementing https://github.com/encode/httpcore/pull/185#issuecomment-692229274

Attempt peeking data from the socket to detect whether the connection has reached EOF ("broken connection" / "dropped connection").

Better cross-platform support than the `selectors` alternative here: #185.

Follow-ups:

- Rename `is_connection_dropped()` to `is_at_eof()`.

### DIFF
--- a/httpcore/_utils.py
+++ b/httpcore/_utils.py
@@ -66,8 +66,9 @@ def origin_to_url_string(origin: Origin) -> str:
     return f"{scheme.decode('ascii')}://{host.decode('ascii')}{port}"
 
 
-def _wait_io_events(socks: list, events: int, timeout: float = None) -> list:
-    # Better cross-platform approach than low-level `select.select()` calls.
+def _wait_for_io_events(socks: list, events: int, timeout: float = None) -> list:
+    # Prefer the `selectors` module rather than the lower-level `select` module to
+    # improve cross-platform support.
     # See: https://github.com/encode/httpcore/issues/182
     sel = selectors.DefaultSelector()
     for sock in socks:
@@ -76,4 +77,4 @@ def _wait_io_events(socks: list, events: int, timeout: float = None) -> list:
 
 
 def wait_for_read(socks: list, timeout: float = None) -> list:
-    return _wait_io_events(socks, events=selectors.EVENT_READ, timeout=timeout)
+    return _wait_for_io_events(socks, events=selectors.EVENT_READ, timeout=timeout)
--- a/httpcore/_utils.py
+++ b/httpcore/_utils.py
@@ -66,8 +66,14 @@ def origin_to_url_string(origin: Origin) -> str:
     return f"{scheme.decode('ascii')}://{host.decode('ascii')}{port}"
 
 
-def is_socket_at_eof(sock: socket.socket) -> bool:
-    assert sock.gettimeout() == 0, "sock must be non-blocking."
+def is_socket_at_eof(sock_fd: int) -> bool:
+    # Duplicate the socket from the file descriptor. (we do this so we always get a
+    # real `socket.socket` object, rather than a library-provided socket-like object,
+    # which may behave differently and/or may not have an implementation for `.recv()`.)
+    sock = socket.fromfd(sock_fd, socket.AF_INET, socket.SOCK_STREAM)
+    # Then put the copy into non-blocking mode. (We need this so that the `.recv()` call
+    # does not block.)
+    sock.setblocking(False)
 
     try:
         data = sock.recv(1, socket.MSG_PEEK)
--- a/httpcore/_utils.py
+++ b/httpcore/_utils.py
@@ -66,18 +66,34 @@ def origin_to_url_string(origin: Origin) -> str:
     return f"{scheme.decode('ascii')}://{host.decode('ascii')}{port}"
 
 
-def is_socket_at_eof(sock_fd: int) -> bool:
-    # Duplicate the socket from the file descriptor. (we do this so we always get a
-    # real `socket.socket` object, rather than a library-provided socket-like object,
-    # which may behave differently and/or may not have an implementation for `.recv()`.)
-    sock = socket.fromfd(sock_fd, socket.AF_INET, socket.SOCK_STREAM)
-    # Then put the copy into non-blocking mode. (We need this so that the `.recv()` call
-    # does not block.)
+def is_socket_at_eof(sock_fd: int, family: int, type: int) -> bool:
+    """
+    Return whether a socket, as identified by its file descriptor, has reached
+    EOF, i.e. whether its read buffer is empty.
+
+    If we're still expecting data, then the socket being at EOF most likely means
+    that the server has disconnected.
+    """
+    # Duplicate the socket, so we get a distinct throw-away copy.
+    # (We do this instead of accepting a `socket` object from the backend-specific
+    # implementation, because those may not actually be *real* `socket` objects.
+    # We can't use `socket.socket(sock_fd)` either, although it auto-populates family
+    # and type, because it would *replace* the existing socket, making the previous
+    # file descriptor obsolete.)
+    sock = socket.fromfd(sock_fd, family, type)
+
+    # Put the copy into non-blocking mode. We need this so that the `.recv()` call
+    # does not block.
     sock.setblocking(False)
 
+    # Then peek a byte of data, to see if there's someone still sending data on the
+    # other end, or if they disconnected.
     try:
         data = sock.recv(1, socket.MSG_PEEK)
     except BlockingIOError:
         return False
     else:
         return data == b""
+    finally:
+        # Dispose the copy.
+        sock.close()
--- a/httpcore/_utils.py
+++ b/httpcore/_utils.py
@@ -83,17 +83,23 @@ def is_socket_at_eof(sock_fd: int, family: int, type: int) -> bool:
     sock = socket.fromfd(sock_fd, family, type)
 
     # Put the copy into non-blocking mode. We need this so that the `.recv()` call
-    # does not block.
+    # raises a `BlockingIOError` if the socket is still active, instead of blocking
+    # and returning actual data after a while.
     sock.setblocking(False)
 
-    # Then peek a byte of data, to see if there's someone still sending data on the
-    # other end, or if they disconnected.
+    # Then peek a byte of data to check on how the other end is doing...
     try:
         data = sock.recv(1, socket.MSG_PEEK)
     except BlockingIOError:
+        # The socket would block if we attempted to read.
+        # => There's still someone on the other end.
         return False
     else:
-        return data == b""
+        # The socket returned immediately if we attempted to read.
+        # => Can only happen if the server on the other end has disconnected, in which
+        # case `data` should be empty bytes.
+        assert data == b""
+        return True
     finally:
         # Dispose the copy.
         sock.close()
--- a/httpcore/_utils.py
+++ b/httpcore/_utils.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import selectors
 import sys
 import typing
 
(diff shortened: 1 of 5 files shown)


httpcore/_utils.py
```diff
@@ -63,3 +64,20 @@ def origin_to_url_string(origin: Origin) -> str:
     scheme, host, explicit_port = origin
     port = f":{explicit_port}" if explicit_port != DEFAULT_PORTS[scheme] else ""
     return f"{scheme.decode('ascii')}://{host.decode('ascii')}{port}"
+
+
+def is_socket_at_eof(sock_fd: int) -> bool:
+    # Duplicate the socket from the file descriptor. (we do this so we always get a
+    # real `socket.socket` object, rather than a library-provided socket-like object,
+    # which may behave differently and/or may not have an implementation for `.recv()`.)
+    sock = socket.fromfd(sock_fd, socket.AF_INET, socket.SOCK_STREAM)
```
JayH5, line 73: I guess maybe an important note is that this is TCP-only, although it wouldn't be anything else (yet).

I was looking through some Trio code recently trying to learn about sockets and I saw this code to support different socket types: https://github.com/python-trio/trio/blob/b74f1d85a9c8e23950f49f72bb5fcd28534dfff2/trio/_socket.py#L290-L307

Not sure if we also need to do `sock.detach()`?

JayH5: > Not sure if we also need to do sock.detach()?

Reading more closely and I think the answer is "no". `socket.fromfd` dup's the file descriptor so we do want it to close that.

(Also I'm not really suggesting any change here, just linking to that Trio source out of interest)

florimondmanca: > although it wouldn't be anything else

Actually it could, if the user is using a Unix Domain Socket… 😕

httpcore/_utils.py
```diff
@@ -63,3 +64,20 @@ def origin_to_url_string(origin: Origin) -> str:
...
+    # Then put the copy into non-blocking mode. (We need this so that the `.recv()` call
+    # does not block.)
+    sock.setblocking(False)
+
+    try:
+        data = sock.recv(1, socket.MSG_PEEK)
```
cdeler, line 79: Hello,

I'm sorry for the question, do we really need to copy the socket? May be we can just try to
`data = original_socket.recv(1, socket. MSG_PEEK | socket. MSG_DONTWAIT)`

florimondmanca: Yes that's what I'd like to do, but as I've put in another comment in some cases this won't work.

Eg one weird case is when rubbing anyio in asyncio with uvloop: the socket object we get is a weird `PseudoSocket` instance, and when you try to call `.recv()` on it it raises an error saying it's not implemented. That's actually the only case I've seen of this for now, and maybe an okay strategy would be "recv(), and if it doesn't succeed copy the socket temporarily to get a real socket instance, and call recv() again". Then the "TCP only" limitation would only apply to the latter case, which should be a minority.

cdeler: What if you try:
```python
sock = socket(fileno=sock_fd)
data = sock.recv(1, socket. MSG_PEEK | socket. MSG_DONTWAIT)
```

According to the [official doc](https://docs.python.org/3/library/socket.html#socket.socket):
> If fileno is specified, the values for family, type, and proto are auto-detected from the specified file descriptor

and
> Unlike socket.fromfd(), fileno will return the same socket and not a duplicate

so you don't need to close it

florimondmanca: Wonderful. 💞

cdeler: Unfortunately according to the [python source](https://github.com/python/cpython/blob/deb016224cc506503fb05e821a60158c83918ed4/Lib/test/test_socket.py#L2986) `MSG_DONTWAIT` is for Linux only.

cdeler: (but probably we still need to close the `sock`)

httpcore/_utils.py
```diff
@@ -63,3 +64,20 @@ def origin_to_url_string(origin: Origin) -> str:
...
+    except BlockingIOError:
+        return False
+    else:
+        return data == b""
```
cdeler, line 83:
```suggestion
        return data == b""
    finally:
        sock.close()
    
```

florimondmanca:
@cdeler @JayH5 Okay, I've revisited the approach for `is_socket_at_eof()`, simply passing explicitly the `fd`, `family` and `type` so we can properly duplicate the socket in both the TCP and UDS cases. I've done some testing against a Uvicorn server over TCP and UDS, in a standard and "server has disconnected early" scenarios, and everything seems to work well. How is this looking?

cdeler (approved):
It's brilliant

JayH5 (approved):
LGTM. Kinda curious whether a library like Trio would want to include this kind of low-level function.

florimondmanca:
Hi @njsmith! Pinging you here just in case you'd have any thoughts on this approach. Since that's not the way trio is doing things (_I think?_), I'm curious if there's some hidden socket wizardry subtelty we're missing, or if this approach seems legit? For more context feel free to read through #185.

njsmith:
Trio sockets expose a special `is_readable` method, exactly for this use case, so on Trio you should just use that :-).

The implementation is pretty simple, and AFAIK there aren't any compatibility issues or anything:

https://github.com/python-trio/trio/blob/20ee2b1b7376db637435d80e266212a35837ddcc/trio/_socket.py#L471-L478

I skimmed #185 but I couldn't really follow what you all were worrying about. If you want to steal that code for non-Trio backends, go for it. Otherwise, I'm not sure what to say b/c I don't know what problem you're trying to solve :-)

On Trio though please do call `sock.is_readable` instead of pulling out the fileno and doing something custom, because that will allow [mock networking for testing to work](https://trio.readthedocs.io/en/stable/reference-testing.html#trio.socket.set_custom_socket_factory).

florimondmanca:
@njsmith Woops — meant to point you at #182, the original issue. Yeah it's a bit fuzzy, here's a quick run-through… (also helpful so I can validate this is all well-posed…)

The problem exposed in #182 is something that came up in urllib3 years ago, which we're vulnerable to as of today: when many files are open and we hit the 1024 fd limit on Unix (512 on Windows), `select.select()` starts failing.

In #185 I explored using the `selectors` module, which solves the Unix case by using a suitable selector based on the platform (eg poll, kqueue). But @agronholm mentioned that it wouldn't solve the Windows case because only `select.select()` is available there (as the trio `is_readable()` method also mentions).

So this PR takes a different approach, peeking onto the socket rather than using any selector. Then since we're able to get a raw socket object from async backends too, we're now using a generic cross-platform helper (the `is_socket_at_eof()` function) to do the same "has the socket reached EOF?" logic in both the sync and async cases. (This goes a bit beyond the scope of #182 which only reported an issue in the sync case, but at least we're now consistent everywhere.)

I guess I'm not asking anything here, was just curious what your thoughts would be given your experience with low-level networking via trio. If this still sounds nebulous, then, fair enough. 😄

njsmith:
> The problem exposed in #182 is something that came up in urllib3 years ago, which we're vulnerable to as of today: when many files are open and we hit the 1024 fd limit on Unix (512 on Windows), `select.select()` starts failing.

`select` is a very different beast on Unix and Windows.

On Unix, `select` uses a *bitmask* of all possible fds up to some numeric limit. So it isn't even the number of fds that you pass in that matters, what matters is the *numeric value* of your highest-numbered fd. Obviously in modern times this is terrible, so they invented `poll`, which takes an array of integer fds, so now the numeric values don't matter, just the number that you're passing in.

On Windows, `select` has always used an array of socket handles. So Windows `select` has exactly the same performance characteristics as Unix `poll`.

There are also the fancier even-more-modern syscalls like `epoll` and IOCP and all that, which the `selectors` module tries to use if it can. But those are designed to optimize for when you're watching a large number of sockets at once. In this case we're only trying to query a single socket at a time, so the fancy modern options are actually less efficient than good ol' `poll`/`select`.

So I think that six lines of code in Trio will solve all your problems everywhere.

> Then since we're able to get a raw socket object from async backends too

Again, please don't try to pull out raw socket objects on Trio, b/c it breaks testing helpers. If there's something you need that's not available on Trio socket objects, lmk and we'll figure out how to add it.

cdeler:
@florimondmanca 

Original issue fixing could be checked if you add the following test:

```python3
# conftest.py
@pytest.fixture(scope="function")
def many_files() -> typing.Iterator[None]:
    files = []
    real_file = tempfile.NamedTemporaryFile(delete=False)

    try:
        real_file.write(b"some")
        real_file.flush()
        files.append(real_file.fileno())
        for _ in range(1023):
            fd = os.dup(real_file.fileno())
            files.append(fd)

        yield
    finally:
        for fd in files:
            os.close(fd)

        os.remove(real_file.name)


# test_interface.py
@pytest.mark.anyio
async def test_dropped_connection_detection_bug(
    backend: str, server: Server, many_files
) -> None:
    async with httpcore.AsyncConnectionPool(backend=backend) as http:
        method = b"GET"
        url = (b"http", *server.netloc, b"/")
        headers = [server.host_header]
        status_code, headers, stream, ext = await http.arequest(method, url, headers)
        await read_body(stream)

        assert status_code == 200
        assert ext == {"http_version": "HTTP/1.1", "reason": "OK"}
        assert len(http._connections[url[:3]]) == 1  # type: ignore

        method = b"GET"
        url = (b"http", *server.netloc, b"/")
        headers = [server.host_header]
        status_code, headers, stream, ext = await http.arequest(method, url, headers)
        await read_body(stream)

        assert status_code == 200
        assert ext == {"http_version": "HTTP/1.1", "reason": "OK"}
        assert len(http._connections[url[:3]]) == 1  # type: ignore

```

May be you can add the test to the suite?

florimondmanca:
Closing, in favor of #185… :-)

PR #194, opened 2020-09-22
Add local test server, using Hypercorn

### LINKED ISSUE #109: Local test server

florimondmanca:
Closes #109, inspred by https://github.com/encode/httpx/pull/412, as well as the Uvicorn setup in the HTTPX repo.

* Switch almost all tests to using a local test server (using Hypercorn). Also use Hypercorn for UDS tests, so that we drop Uvicorn completely as a dev dependency. I left the HTTPS proxy test as-is, because there's some non-trivial-to-me local SSL configuration to do so that the proxy knows how to use and recognize certs of the local server.
* Hypercorn is 3.7+ only, so we'd be skipping most tests on 3.6. To make up for this I kept some of the existing tests against `example.org`, marked as "live" tests.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,10 +23,10 @@ black==20.8b1
 flake8==3.8.3
 flake8-bugbear==20.1.4
 flake8-pie==0.6.1
+hypercorn==0.10.2
 isort==5.5.2
 mitmproxy==5.2
 mypy==0.782
 pytest==6.0.2
 pytest-cov==2.10.1
 trustme==0.6.0
-uvicorn==0.11.8
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,7 +23,7 @@ black==20.8b1
 flake8==3.8.3
 flake8-bugbear==20.1.4
 flake8-pie==0.6.1
-hypercorn==0.10.2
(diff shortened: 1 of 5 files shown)


florimondmanca:
I reckon this is a pretty big change footprint…

One way forward that doesn't require switching to Hypercorn yet might be to isolate a `Server` helper, whose `netloc` returns `("example.org", 80)`, like in this PR. Then later we can add a Hypercorn server fixture, used in _some_ (one?) of the tests to begin with, and then finish that off with a full switch.

florimondmanca:
Closing for now, going for #204, then we'll be able to change the `Server` class it adds so that it's backed by Hypercorn rather than the live `example.org` website. :+1:

PR #195, opened 2020-09-24
Add some basic connection-pool-specific tests

tomchristie:
A few basic test cases for `httpcore.AsyncConnectionPool`/`httpcore.SyncConnectionPool` that work at the level of mocking out the underlying connection instances that they use.

Note that unlike `test_interfaces.py` these test cases just run against a single async backend case. Here we're not interested in the different backends, and we're not touching any backend specific code.

These tests aren't perfect but they're a great start towards us being able to test `httpcore` a bit more comprehensively by starting to add some layer-by-layer testing.

We'll also want to use them to help us start to make the contract between `ConnectionPool` and `Connection` more explicit.

### DIFF
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -2,7 +2,7 @@ import warnings
 from ssl import SSLContext
 from typing import AsyncIterator, Callable, Dict, List, Optional, Set, Tuple, cast
 
-from .._backends.auto import AsyncLock, AsyncSemaphore
+from .._backends.auto import AsyncBackend, AsyncLock, AsyncSemaphore
 from .._backends.base import lookup_async_backend
 from .._exceptions import LocalProtocolError, PoolTimeout, UnsupportedProtocol
 from .._threadlock import ThreadLock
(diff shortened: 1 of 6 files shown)


httpcore/_async/connection_pool.py
```diff
@@ -147,6 +147,24 @@ def _connection_acquiry_lock(self) -> AsyncLock:
...
+        origin: Tuple[bytes, bytes, int],
+        backend: AsyncBackend,
+        http2: bool = False,
+        uds: str = None,
+        ssl_context: SSLContext = None,
+        local_address: str = None,
+    ) -> AsyncHTTPConnection:
+        return AsyncHTTPConnection(
+            origin=origin,
+            http2=self._http2,
```
jcugat, line 161: Don't we want to use the arguments instead of the instance attributes?

```suggestion
            http2=http2,
```

tests/async_tests/test_connection_pool.py
```diff
@@ -0,0 +1,179 @@
...
+        async for chunk in stream:
+            body.append(chunk)
+        return b"".join(body)
+    finally:
+        await stream.aclose()
+
+
+@pytest.mark.trio
+async def test_sequential_requests_h11() -> None:
+    async with ConnectionPool(http_version="HTTP/1.1") as http:
```
jcugat, line 78: We could parametrize `http_version` since both tests `test_sequential_requests_h11` and `test_sequential_requests_h2` are identical otherwise.

johtso: Data driven testing :heart:

httpcore/_async/connection_pool.py
tomchristie: Ah good catch. Prompted a nice clean-up... Don't pass the arguments through at all.

We really only need to be passing the `origin` argument. Everything else is available as attributes on the class.

tests/async_tests/test_connection_pool.py
tomchristie: 👍

jcugat (approved):
Looks good!

florimondmanca (approved):
_Great_ addition! Will help us prevent obvious regressions in our connection pooling. :-)

PR #196, opened 2020-09-25
Added pproxy test fixture, which allows to test requesting through socks proxy

cdeler:
As a part of https://github.com/encode/httpx/issues/203 (draft pr is https://github.com/encode/httpcore/pull/188) I'd like to introduce the `"socks"` test fixture.

Using `pproxy` it starts a subprocess with `socks4` and `socks5` (with auth and without auth) proxies on localhost.

Also the PR contains some new types, required in the fixture  

**There is a problem**: this server doesn't support `socks4a` proxy (may be it's not so important)

### DIFF
--- a/httpcore/_types.py
+++ b/httpcore/_types.py
@@ -1,8 +1,8 @@
 """
 Type definitions for type checking purposes.
 """
-
-from typing import List, Mapping, Optional, Tuple, TypeVar, Union
+import enum
+from typing import List, Mapping, NamedTuple, Optional, Tuple, TypeVar, Union
 
 T = TypeVar("T")
 StrOrBytes = Union[str, bytes]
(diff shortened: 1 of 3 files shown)


tomchristie:
I don't think this is quite what I was expecting here.
A more minimal first step would be adding some `pproxy` tests cases for our *existing* regular HTTP/HTTPS proxy support.

cdeler:
@tomchristie 

You are talking about `proxy_server` fixture, which is used by `test_http_proxy` and `test_proxy_https_requests`?

https://github.com/encode/httpcore/blob/1f67394211e82713dfd18be9014efaec769ed05c/tests/conftest.py#L84-L85

cdeler:
Closing this PR, replacing it by https://github.com/encode/httpcore/pull/199

PR #199, opened 2020-09-28
Migrate http proxy tests to pproxy

cdeler:
In https://github.com/encode/httpcore/pull/196 comments it was mentioned that the first step of implementing `socks` proxy support might be replacing `mitmproxy` by `pproxy` (which allows us to test `http`, `socks4` and `socks5` support)

So that I rewrote the test fixtures and changed `requirement` file.

The next steps I propose to do after this PR:
1. rebase https://github.com/encode/httpcore/pull/188 draft
1. move on acroding to the comment https://gitter.im/encode/community?at=5f6db8edf41f4105e4d7a9eb

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,8 +1,5 @@
 import contextlib
 import os
-import shlex
-import socket
-import subprocess
 import threading
 import time
 import typing
@@ -12,42 +9,19 @@ import uvicorn
 
 from httpcore._types import URL
 
-from .utils import Server
+from .utils import Server, http_proxy_server
 
 SERVER_HOST = "example.org"
 HTTPS_SERVER_URL = "https://example.org"
 
 
-def wait_until_pproxy_serve_on_port(host: str, port: int):
-    while True:
-        try:
-            sock = socket.create_connection((host, port))
-        except ConnectionRefusedError:
-            time.sleep(0.25)
-        else:
-            sock.close()
-            break
-
-
 @pytest.fixture(scope="session")
 def proxy_server() -> typing.Iterator[URL]:
-    http_proxy_host = "127.0.0.1"
-    http_proxy_port = 8080
+    proxy_host = "127.0.0.1"
+    proxy_port = 8080
 
-    proc = None
-    try:
-        command_str = f"pproxy -l http://{http_proxy_host}:{http_proxy_port}/"
-        command = shlex.split(command_str)
-        proc = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-
-        wait_until_pproxy_serve_on_port(http_proxy_host, http_proxy_port)
-
-        print(f"HTTP proxy started on http://{http_proxy_host}:{http_proxy_port}/")
-
-        yield b"http", http_proxy_host.encode(), http_proxy_port, b"/"
-    finally:
-        if proc is not None:
-            proc.kill()
+    with http_proxy_server(proxy_host, proxy_port) as proxy_server_fixture:
+        yield proxy_server_fixture
 
 
 class UvicornServer(uvicorn.Server):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,9 @@
-import asyncio
 import contextlib
 import os
+import shlex
+import socket
 import ssl
+import subprocess
 import threading
 import time
 import typing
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -70,36 +27,36 @@ def ca_ssl_context(cert_authority: trustme.CA) -> ssl.SSLContext:
...
-    with example_org_cert.private_key_and_cert_chain_pem.tempfile() as tmp:
-        yield tmp
+def proxy_server() -> typing.Iterator[URL]:
+    http_proxy_host = "127.0.0.1"
+    http_proxy_port = 8080
 
+    proc = None
+    try:
+        command_str = f"pproxy -l http://{http_proxy_host}:{http_proxy_port}/"
+        command = shlex.split(command_str)
```
florimondmanca, line 49: Don't think we need `shlex` here, if we build the list of arguments directly…?

```suggestion
        command = ["pproxy", "-l", f"http://{http_proxy_host}:{http_proxy_port}/"]
```

tests/conftest.py
```diff
@@ -70,36 +27,36 @@ def ca_ssl_context(cert_authority: trustme.CA) -> ssl.SSLContext:
...
+        else:
+            sock.close()
+            break
 
 
 @pytest.fixture(scope="session")
-def example_org_cert_path(example_org_cert: trustme.LeafCert) -> typing.Iterator[str]:
-    with example_org_cert.private_key_and_cert_chain_pem.tempfile() as tmp:
-        yield tmp
+def proxy_server() -> typing.Iterator[URL]:
```
florimondmanca, line 42: Instead of overriding the existing code, which makes things _a bit_ hard to review (nothing blocking though), should we take the opportunity here and define any code related to the test proxy in `conftest/utils.py`? I think it's a good pattern to isolate anything that doesn't _need_ to be defined in `conftest.py` to somewhere else, for readability. Bonus point here is the diff should be easier to review. :-)

tests/conftest.py
```diff
@@ -70,36 +27,36 @@ def ca_ssl_context(cert_authority: trustme.CA) -> ssl.SSLContext:
...
+    http_proxy_host = "127.0.0.1"
+    http_proxy_port = 8080
```
florimondmanca, line 44: (Nit) This is a tad verbose, maybe…?

```suggestion
    proxy_host = "127.0.0.1"
    proxy_port = 8080
```

tests/conftest.py
```diff
@@ -70,36 +27,36 @@ def ca_ssl_context(cert_authority: trustme.CA) -> ssl.SSLContext:
...
 
-@pytest.fixture()
-def proxy_server(example_org_cert_path: str) -> typing.Iterator[URL]:
-    """Starts a proxy server on a different thread and yields its origin tuple.
+        wait_until_pproxy_serve_on_port(http_proxy_host, http_proxy_port)
 
-    The server is configured to use a trustme CA and key, this will allow our
-    test client to make HTTPS requests when using the ca_ssl_context fixture
-    above.
+        print(f"HTTP proxy started on http://{http_proxy_host}:{http_proxy_port}/")
```
florimondmanca, line 54: Should we instead let the stdout/stderr of the pproxy process pass through?

Not sure which flag we'd need to pass to `Popen(stdout=..., stderr=...)`, but that way we wouldn't need this print statement, and we'd be able to see the full output using controls built into pytest (such as the `-s` flag).

florimondmanca (commented):
Looking good!

How come the HTTPS proxy test passes w/o further configuration or tweaks, though, whereas there are a few fixtures such as `ca_ssl_context` that relate to our previous HTTPS proxy setup? We ought to clean up any fixtures we don't use anymore, and see if there's anything we need to update in HTTPS proxies tests.

tests/conftest.py
cdeler: I hope I understood you correctly. I've moved all the code apart from fixture initialization into the `tests/utils.py`

tests/conftest.py
cdeler: You are right, I was not consistent. Removed the `print` statement and the pipes

cdeler:
@florimondmanca 
I've removed `trustme` from requirements, but found that you are using it in the https://github.com/encode/httpcore/pull/205

florimondmanca:
@cdeler Sounds good, feel free to drop anything you don't need as of this PR, I can add stuff back in the future. :)

tests/conftest.py
```diff
@@ -1,109 +1,27 @@
...
-    integrate in our tests.
-    """
-    try:
-        thread = ProxyWrapper(PROXY_HOST, PROXY_PORT, certs=[example_org_cert_path])
-        thread.start()
-        thread.notify.started.wait()
-        yield (b"http", PROXY_HOST.encode(), PROXY_PORT, b"/")
-    finally:
-        thread.join()
+    with http_proxy_server(proxy_host, proxy_port) as proxy_server_fixture:
```
florimondmanca, line 23: (Nit) How about `as proxy_url`, since it seems the context manager returns an URL tuple?

tests/utils.py
```diff
@@ -11,6 +15,17 @@ def lookup_sync_backend():
     return "sync"
 
 
+def wait_until_pproxy_serve_on_port(host: str, port: int):
```
florimondmanca, line 18: (Nit) Looks like there's nothing particular to pproxy here, so rather should we name this `_wait_can_connect(host, port)`, or something more generic like that? (With `_` indicating this is not a helper we're meaning to import from conftest or our tests.)

florimondmanca (approved):
Great work! Two nits, but this is mergeable to me. Feel free to address whatever feels appropriate, and then we can get this merged. :)

tests/conftest.py
cdeler: Renamed

tests/utils.py
cdeler: Good point, renamed

PR #201, opened 2020-10-01
Treat asyncio `resume_reading` error as a closed connection

cdeler:
Closes https://github.com/encode/httpx/issues/1213

There is a gist in the issue, which can help you reproduce the error: [link](https://gist.github.com/cdeler/f43704721829e6ed3c799ddc7d443d5c) . You can also switch the async backend and check, that the error appears only when we `asyncio` backend.

The introduced fix mends the error, so that the error message caused by the exception looks like:

> httpcore.RemoteProtocolError: peer closed connection without sending complete message body (received 917504 bytes, expected 1000000)

instead of the original stacktrace from the issue

**There is a problem:**
I cannot figure out how to make a unit test for this error

### DIFF
--- a/httpcore/_backends/asyncio.py
+++ b/httpcore/_backends/asyncio.py
@@ -131,9 +131,20 @@ class SocketStream(AsyncSocketStream):
         exc_map = {asyncio.TimeoutError: ReadTimeout, OSError: ReadError}
         async with self.read_lock:
             with map_exceptions(exc_map):
-                return await asyncio.wait_for(
-                    self.stream_reader.read(n), timeout.get("read")
-                )
+                try:
+                    return await asyncio.wait_for(
+                        self.stream_reader.read(n), timeout.get("read")
+                    )
+                except AttributeError:  # pragma: nocover
+                    """
+                    We have to return empty string there due to one ugly bug in asyncio
+                    More information you can find in this issue:
+                        https://github.com/encode/httpx/issues/1213
+
+                    The empty byte-string returned from there caused
+                    httpcore.RemoteProtocolError at the top level
+                    """
+                    return b""
 
     async def write(self, data: bytes, timeout: TimeoutDict) -> None:
         if not data:
--- a/httpcore/_backends/asyncio.py
+++ b/httpcore/_backends/asyncio.py
@@ -135,16 +135,15 @@ class SocketStream(AsyncSocketStream):
                     return await asyncio.wait_for(
                         self.stream_reader.read(n), timeout.get("read")
                     )
-                except AttributeError as ex:  # pragma: nocover
-                    if "resume_reading" in str(ex):
-                        # We have to return empty string there,
-                        # due to one ugly bug in asyncio
-                        # More information you can find in this issue:
-                        # https://github.com/encode/httpx/issues/1213
+                except AttributeError as exc:  # pragma: nocover
+                    if "resume_reading" in str(exc):
+                        # Python's asyncio has a bug that can occur when a
+                        # connection has been closed, while it is paused.
+                        # See: https://github.com/encode/httpx/issues/1213
                         #
-                        # Returning an empty byte-string will eventually raise
-                        # an httpcore.RemoteProtocolError to the user when this
-                        # goes through our HTTP parsing layer.
+                        # Returning an empty byte-string to indicate connection
+                        # close will eventually raise an httpcore.RemoteProtocolError
+                        # to the user when this goes through our HTTP parsing layer.
                         return b""
                     raise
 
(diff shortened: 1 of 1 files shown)


httpcore/_backends/asyncio.py
```diff
@@ -131,9 +131,20 @@ async def read(self, n: int, timeout: TimeoutDict) -> bytes:
...
         async with self.read_lock:
             with map_exceptions(exc_map):
-                return await asyncio.wait_for(
-                    self.stream_reader.read(n), timeout.get("read")
-                )
+                try:
+                    return await asyncio.wait_for(
+                        self.stream_reader.read(n), timeout.get("read")
+                    )
+                except AttributeError:  # pragma: nocover
```
florimondmanca, line 138: Should we check that the attribute lookup that caused the error is indeed due to `resume_reading`?

```python
except AttributeError as exc:
    if "resume_reading" in str(exc):
        # Explanation...
        return b""
    raise
```

florimondmanca (commented):
Hm, okay, sounds like a sensible ugly fix to an asyncio weakness.

Note: have we tried to see if this reproduces on anyio (rubbing in asyncio)?

cdeler:
@florimondmanca 
Having updated the gist, I tried to reproduce the error versus "anyio" backend, but couldn't. I'll try to figure out if we can do something with it

PS
Thats miracle, thank you @agronholm for `"anyio"`

httpcore/_backends/asyncio.py
```diff
@@ -131,9 +131,21 @@ async def read(self, n: int, timeout: TimeoutDict) -> bytes:
...
+                    )
+                except AttributeError as ex:  # pragma: nocover
+                    if "resume_reading" in str(ex):
+                        # We have to return empty string there,
+                        #     due to one ugly bug in asyncio
+                        # More information you can find in this issue:
+                        #     https://github.com/encode/httpx/issues/1213
+                        #
+                        # The empty byte-string returned from there caused
+                        #     httpcore.RemoteProtocolError at the top level
```
florimondmanca, line 146: (Nit) Tweaking the leading spaces here…

```suggestion
                        # We have to return empty string there,
                        # due to one ugly bug in asyncio
                        # More information you can find in this issue:
                        # https://github.com/encode/httpx/issues/1213
                        #
                        # The empty byte-string returned from there caused
                        # httpcore.RemoteProtocolError at the top level
```

florimondmanca, line 146: What do we mean by this? Is is that returning an empty byte string _will_ cause a `RemoteProtocolError` to be raised later (by `h11` / `h2`)?

```suggestion
                        # Returning an empty byte-string will eventually raise
                        # an httpcore.RemoteProtocolError to the user when this
                        # goes through our HTTP parsing layer.
```

florimondmanca (approved):
Hello! I've got some nitpicks on the code comment… Other than that, I'm not sure how confident we are merging this change? Can you point here to a comment with a test setup that allows reproducing this 100% of the time, to verify this is fixed with this PR?

cdeler:
@florimondmanca 
> Can you point here to a comment with a test setup that allows reproducing this 100% of the time, to verify this is fixed with this PR?

I created a gist with the test file

https://gist.github.com/cdeler/86d99a42c5bfb522bc15cac8d27824aa

Just run the script using python3, also you can check that the bug is not reproducible under other backends (just run `path_to_script --help` to check other switches)

httpcore/_backends/asyncio.py
cdeler: Fixed

cdeler:
@florimondmanca 

Probably I can add the test from the gist to our CI

httpcore/_backends/asyncio.py
```diff
@@ -131,9 +131,22 @@ async def read(self, n: int, timeout: TimeoutDict) -> bytes:
...
+                except AttributeError as ex:  # pragma: nocover
+                    if "resume_reading" in str(ex):
+                        # We have to return empty string there,
+                        # due to one ugly bug in asyncio
+                        # More information you can find in this issue:
+                        # https://github.com/encode/httpx/issues/1213
+                        #
+                        # Returning an empty byte-string will eventually raise
+                        # an httpcore.RemoteProtocolError to the user when this
+                        # goes through our HTTP parsing layer.
```
tomchristie, line 147:
```suggestion
                except AttributeError as exc:  # pragma: nocover
                    if "resume_reading" in str(exc):
                        # Python's asyncio has a bug that can occur when a
                        # connection has been closed, while it is paused.
                        # See: https://github.com/encode/httpx/issues/1213
                        #
                        # Returning an empty byte-string to indicate connection
                        # close will eventually raise an httpcore.RemoteProtocolError
                        # to the user when this goes through our HTTP parsing layer.
```

tomchristie (approved):
Lovely, thanks!

PR #203, opened 2020-10-02
Update examples after 0.11.0

sandrotosi:
in 0.11.0 the return tuple of `request()` changed: https://github.com/encode/httpcore/blob/master/CHANGELOG.md#0110-september-22nd-2020

examples needed to be adapted to it

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Here's an example of making an HTTP GET request using `httpcore`...
 
 ```python
 with httpcore.SyncConnectionPool() as http:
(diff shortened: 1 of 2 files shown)


florimondmanca (approved):
Great catch!

Do we need to update the docs home page as well?

sandrotosi:
> Do we need to update the docs home page as well?

yeah, i didnt realize the same code was duplicated -- i've updated the doc/index.md file as well

tomchristie (approved):
Great, thanks!

PR #204, opened 2020-10-02
Abstract away example.org in interface tests

florimondmanca:
Attempt at reducing the scope of #194: as a first step, make our tests agnostic to which "live" endpoint we're hitting, by adding a small `Server` class and using it in `test_interfaces.py` (i.e. where we're actually hitting `example.org`).

Next step would be to add Hypercorn, switch `Server`'s properties so that it points to `localhost:<port>`, instead of `example.org`, and use that updated version in fixtures.

Diff size is still not negligible, but hopefully this is a non-controversial and easily reviewable change.

### DIFF
--- a/tests/async_tests/test_interfaces.py
+++ b/tests/async_tests/test_interfaces.py
@@ -5,8 +5,8 @@ import pytest
 
 import httpcore
 from httpcore._types import URL
-from tests.conftest import Server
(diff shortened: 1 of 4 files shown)


tomchristie (approved):
Great approach, yup! 👍

PR #205, opened 2020-10-02
Introduce Hypercorn as a local test server

### LINKED ISSUE #109: Local test server

florimondmanca:
Re-ignite of #194:

Closes #109 

100% of our tests now run locally (and switching to `pproxy` (via #196) made this possible with much less pain :-)).

Note: Hypercorn only runs on 3.7, so on 3.6 we keep testing against the `example.org` website, like we do currently. There's a bit of overhead to that, but nothing too blocking I hope, and it's actually nice that we keep some coverage against real-world websites.

Follow-ups include:

- Use Hypercorn for the UDS tests as well, so we remove can Uvicorn entirely (instead of keeping both Hypercorn and Uvicorn as this PR does for now, for ease of reviewing).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,7 +22,7 @@ coverage==5.3
 flake8==3.8.4
 flake8-bugbear==20.1.4
 flake8-pie==0.6.1
-hypercorn==0.10.2; python_version >= '3.7'
+hypercorn==0.11.1; python_version >= '3.7'
 isort==5.5.4
 mypy==0.782
 pproxy==2.3.7
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,9 +22,11 @@ coverage==5.3
 flake8==3.8.4
 flake8-bugbear==20.1.4
 flake8-pie==0.6.1
+hypercorn==0.10.2; python_version >= '3.7'
 isort==5.5.4
 mypy==0.782
 pproxy==2.3.7
 pytest==6.1.1
 pytest-trio==0.6.0
+trustme==0.6.0
 uvicorn==0.12.1
--- a/tests/async_tests/test_interfaces.py
+++ b/tests/async_tests/test_interfaces.py
@@ -33,7 +33,8 @@ async def test_http_request(backend: str, server: Server) -> None:
         await read_body(stream)
 
         assert status_code == 200
-        assert ext == {"http_version": "HTTP/1.1", "reason": "OK"}
+        reason = "OK" if server.sends_reason else ""
+        assert ext == {"http_version": "HTTP/1.1", "reason": reason}
         assert len(http._connections[url[:3]]) == 1  # type: ignore
 
 
@@ -47,7 +48,8 @@ async def test_https_request(backend: str, https_server: Server) -> None:
         await read_body(stream)
 
         assert status_code == 200
-        assert ext == {"http_version": "HTTP/1.1", "reason": "OK"}
+        reason = "OK" if https_server.sends_reason else ""
+        assert ext == {"http_version": "HTTP/1.1", "reason": reason}
         assert len(http._connections[url[:3]]) == 1  # type: ignore
 
 
@@ -85,7 +87,8 @@ async def test_closing_http_request(backend: str, server: Server) -> None:
         await read_body(stream)
 
         assert status_code == 200
-        assert ext == {"http_version": "HTTP/1.1", "reason": "OK"}
+        reason = "OK" if server.sends_reason else ""
+        assert ext == {"http_version": "HTTP/1.1", "reason": reason}
         assert url[:3] not in http._connections  # type: ignore
 
 
@@ -99,7 +102,8 @@ async def test_http_request_reuse_connection(backend: str, server: Server) -> No
         await read_body(stream)
 
         assert status_code == 200
-        assert ext == {"http_version": "HTTP/1.1", "reason": "OK"}
+        reason = "OK" if server.sends_reason else ""
+        assert ext == {"http_version": "HTTP/1.1", "reason": reason}
         assert len(http._connections[url[:3]]) == 1  # type: ignore
 
         method = b"GET"
@@ -109,7 +113,8 @@ async def test_http_request_reuse_connection(backend: str, server: Server) -> No
         await read_body(stream)
 
         assert status_code == 200
-        assert ext == {"http_version": "HTTP/1.1", "reason": "OK"}
+        reason = "OK" if server.sends_reason else ""
+        assert ext == {"http_version": "HTTP/1.1", "reason": reason}
         assert len(http._connections[url[:3]]) == 1  # type: ignore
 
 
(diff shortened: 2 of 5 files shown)


tests/async_tests/test_interfaces.py
```diff
@@ -34,7 +34,7 @@ async def test_http_request(backend: str, server: Server) -> None:
         await read_body(stream)
 
         assert status_code == 200
-        assert ext == {"http_version": "HTTP/1.1", "reason": "OK"}
+        assert ext == {"http_version": "HTTP/1.1", "reason": ""}
```
florimondmanca, line 37: Interestingly, Hypercorn sends the status line as `200`, rather than a full `200 OK`, so the reason sent by the server is empty now.

Re: the follow-up to keep testing against `example.org` on 3.6, I'm not sure exactly how we'd abstract this out, but this is something we'd need to deal with.

A super simple approach would be:

```suggestion
        reason = "OK" if server.sends_reason else ""
        assert ext == {"http_version": "HTTP/1.1", "reason": reason}
```

But again, that's going to be treated as a follow-up. :)

tests/async_tests/test_interfaces.py
```diff
@@ -223,8 +223,8 @@ async def test_proxy_https_requests(
     https_server: Server,
 ) -> None:
     method = b"GET"
-    url = (b"https", *https_server.netloc, b"/")
-    headers = [https_server.host_header]
+    url = (b"https", b"example.org", 443, b"/")
+    headers = [(b"host", b"example.org")]
```
florimondmanca, line 227: Switching these back to using `example.org`, since configuring certs for going through mitm-proxy was quite iffy.

tests/utils.py
```diff
@@ -16,14 +28,58 @@ class Server:
...
+        self._should_exit = False
 
     @property
     def netloc(self) -> Tuple[bytes, int]:
         return (self._host.encode("ascii"), self._port)
 
     @property
     def host_header(self) -> Tuple[bytes, bytes]:
-        return (b"host", self._host.encode("utf-8"))
+        return (b"host", self._host.encode("ascii"))
```
florimondmanca, line 57: Small unrelated change as a cleanup for #204 - for correctness we ought to decode this as `ascii`, I think?

tests/async_tests/test_interfaces.py
florimondmanca: > But again, that's going to be treated as a follow-up. :)

Actually, given that I have to skip coverage enforcement on 3.6 so that CI keeps passing, I'm also open to doing the "fallback to a live server on 3.6" thing right in this PR, although I think doing things in 2 steps is probably easier to review, even if that creates a temporary weakness in the 3.6 test suite.

tests/async_tests/test_interfaces.py
cdeler: https://github.com/encode/httpcore/pull/199 has been merged, so probably it can be changed back?

tests/utils.py
cdeler: @florimondmanca 

Once I've got a strange error with the `Server` fixture:
https://gist.github.com/cdeler/d4b32031687c3a81f5b904c580cf56f0

_Just for your information_

florimondmanca: @cdeler Yup, I can see where that's coming from: we are replacing the `headers` request argument (that contains the host header) with the response `headers` (same variable name).

Is the code for the stack trace in master already, or is it in this PR? If it's in master, we can go for a fix. :)

cdeler: The code is from `master`

cdeler: I can create an issue

scripts/coverage
```diff
@@ -7,4 +7,14 @@ fi
 
 set -x
 
+PYTHON_VERSION=$(${PREFIX}python -V)
+set +e
+echo $PYTHON_VERSION | grep "3.6" > /dev/null
+if [ $? -eq 0 ]; then
```
cdeler, line 13: May we rewrite it to something like

```suggestion
set +e
IS_PYTHON_LESS_37="$(${PREFIX}python -c 'import sys; print(sys.version_info[:2] < (3, 7))')"
if [[ ${IS_PYTHON_LESS_37} == "True" ]]; then
```

Benefits:
1. we control login using `python`, not `bash`
1. I'd rather use `[[` instead `[` ([reason](https://unix.stackexchange.com/questions/306111/what-is-the-difference-between-the-bash-operators-vs-vs-vs))

cdeler (approved):
So I've tested your branch (with `mitmproxy`)

So on my local machine the average test suite run takes 27.63s , whereas the latest `master`'s test run takes more than 90s.

Apart from some questions and a merge conflict, it's a great PR, thank you.

I'm anxious to have these changes in `master`.

scripts/coverage
florimondmanca: :-) Looks like CI doesn't support `[[`

```console
Run scripts/coverage
+ python -c import sys; print(sys.version_info < (3, 7))
+ IS_BELOW_PY_3_7=True
+ [[ True == True ]]
scripts/coverage: 11: scripts/coverage: [[: not found
```

florimondmanca:
I'm happy with the state of this PR, and happy enough with the fact that we're testing against a different location on 3.6 and elsewhere, as Hypercorn only supports 3.7+, but pinging @tomchristie for some of his thoughts as I see it was a possible space of concern for him (https://github.com/encode/httpcore/issues/109#issuecomment-680040109).

requirements.txt
```diff
@@ -22,9 +22,11 @@ coverage==5.3
 flake8==3.8.4
 flake8-bugbear==20.1.4
 flake8-pie==0.6.1
+hypercorn==0.10.2; python_version >= '3.7'
```
JayH5, line 25: hypercorn is up to v0.11.1 now

florimondmanca:
Happy to take some final reviews, otherwise I think I'll just move forward with this one. :-)

cdeler (approved):
Having reviewed that again, I've found it perfect! It would allow us to emulate a lot of things to increase coverage

tests/utils.py
```diff
@@ -42,7 +71,72 @@ def netloc(self) -> Tuple[bytes, int]:
...
+
+
+class HypercornServer(Server):  # pragma: no cover  # Python 3.7+ only
```
florimondmanca, line 77: For the sake of simplicity I'm `nocover`-ing all these classes that are only used either on 3.6 (where `hypercorn` isn't available) or on 3.7+ (where `hypercorn` is available), but I'm also aware that something like [conditional coverage](https://sobolevn.me/2020/02/conditional-coverage) could be used to define rules allowing us to only skip coverage on specific versions. For now, keeping it simple. :-)

florimondmanca:
> Successful in 1m

So delighted how fast our test suite is now ✨

PR #206, opened 2020-10-03
Turn transport.request() into a context manager

### LINKED ISSUE #145: Should `Transport.request()` return a context manager?
Prompted by https://github.com/encode/httpx/pull/998#issuecomment-671070588…

cc @tomchristie 

### Prior art

Right now our transport API looks like this... I'm using a slightly edited version of the README, where we only inspect the response headers:

```python
async with httpcore.AsyncConnectionPool() as http:
    http_version, status_code, reason_phrase, headers, stream = await http.request(
        method=b'GET',
        url=(b'https', b'example.org', 443, b'/'),
        headers=[(b'host', b'example.org'), (b'user-agent', 'httpcore')]
    )

    print(status_code, body)
```

Although it's definitely not obvious to the un-trained eye, this code could be problematic in some situations: the `stream` object is a wrapper over an open socket, and we're not enforcing that users close it properly before exiting the context of the associated request.

This example runs fine, but it's mostly by chance. `AsyncConnectionPool` may open and close connections on a per-request basis, but this doesn't pose a problem because the lifespan of a connection typically outranges that of a request (eg because of `Keep-Alive`). So the overall machinery occurs at the transport level. In other words, when the `AsyncConnectionPool` block exits, all underlying connections are closed, and so the underlying socket is closed, and nothing leaks.

But…

### Case study: per-request resourceful transport case

Consider the case of `ASGITransport` in https://github.com/encode/httpx/pull/998, or any other example that requires a per-request I/O or concurrency resource (in the `ASGITransport` case, a task group for running the ASGI app in the background).

Then currently we _must_ do some unclassy `__aenter__()`/`__aexit__()` dance with that resource, so that `request()` returns a `stream` containing some kind of reference to that resource, with an `aclose_func=...` that properly closes said resource. (This is tricky. I _think_ we need to always pass `sys.exc_info()` to `__aexit__()` so that the resource is aware that it might be being closed in response to an exception within the block, _but I'm not sure_).

For example...

```python
from somewhere import create_async_resource

class MyTransport:
    async def request(self, ...):
        resource = create_async_resource()

        await resource.__aenter__()  # "Acquire" resource.

        async def aclose():
            await resource.__aexit__(*sys.exc_info())  # "Cleanup" resource.

        stream = httpcore.AsyncByteStream(..., aclose_func=aclose)

        return (*..., stream)
```

And now my variant of the README example will be broken for this particular transport:

```python
async with MyTransport() as http:
    http_version, status_code, reason_phrase, headers, stream = await http.request(
        method=b'GET',
        url=(b'https', b'example.org', 443, b'/'),
        headers=[(b'host', b'example.org'), (b'user-agent', 'httpcore')]
    )

    print(status_code, body)

# OOPS! `resource` has leaked...
```

### Leverage transport-level context manager syntax (not satisfactory)

One way to fix it would be to track instances of `resource` on the transport class, and then on `MyTransport.__aexit__()` make sure all `resource` instances are closed. It might not generally be very practical to do this, eg `resource` might fail if you're trying to close it again. And overall it feels like late patching, instead of fixing the problem as close to its source as possible (the request "context").

### Proposal: request-level context manager syntax

On the other hand, what if `.request()` was returning a context manager, instead of a plain awaitable…?

Then we could do this:

```python
from somewhere import create_async_resource

class MyTransport:
    @asynccontextmanager
    async def request(self, ...):
        async with create_async_resource() as resource:.
            stream = httpcore.AsyncByteStream(...)
            yield (*..., stream)
```

The usage syntax will change slightly, to become:

```python
async with MyTransport() as http:
    async with http.request(
        method=b'GET',
        url=(b'https', b'example.org', 443, b'/'),
        headers=[(b'host', b'example.org'), (b'user-agent', 'httpcore')]
    ) as (
        http_version,
        status_code,
        reason_phrase,
        headers,
        stream,
    ):
        print(status_code, body)

    # Regardless of whether `stream.aclose()` was called, the stream is now properly closed,
    # and `resource` was cleaned up.
```

Now _the syntax_ guarantees us that `resource` will be closed whenever we exit the request context.

Benefits:

- Stricter following of principles of structured concurrency (which, as we're starting to observe, is _generally a good thing_).
- Leads to simpler (as in: "more concise, and less likely to shoot oneself in the foot") implementations for transports that need to hold onto resources for each request.

florimondmanca:
Closes #145 

**Note**: recommend reviewing with "Hide whitespace changes" turned on.

This PR introduces a breaking change to the transport API:

Before:

```python
async with Transport() as http:
    response = await http.request(...)
    ...
```

After:

```python
async with Transport() as http:
    async with http.request(...) as response:
        ...
```

From my point of view, motivation is two-fold:

- Move towards more strictly observing principles of structured concurrency. One of the main manifestations of this is preferring to rely on a `with` syntax block, rather than using e.g. wrapper classes that hold a `close` callback. This should make the transport API _more correct_ to use, meaning less chances to shoot oneself in the foot by forgetting a callback somewhere.
- Facilitate per-request holding of I/O state, making things like supporting streaming ASGI responses not only possible, but also very easy to implement. (See https://github.com/encode/httpx/pull/998)

There are trade-offs to consider on the user side of things:

- Callback-driven clients and APIs will have to interface with HTTPCore's preference for `with` blocks. This can be done _correctly_ and _reliably_ with the help of exit stacks, which convert the syntactical scope of a `with` block into a programmatic scope (via `.enter_context()` and `.close()`), but I'd reckon this could introduce friction.
- For HTTPX, embracing the `with` syntax seems like the most promising option: https://github.com/encode/httpx/pull/1355. (For the record, keeping its callback-driven nature was also attempted: https://github.com/encode/httpx/pull/1341.)

### DIFF
--- a/httpcore/__init__.py
+++ b/httpcore/__init__.py
@@ -1,7 +1,7 @@
 from ._async.base import AsyncByteStream, AsyncHTTPTransport
 from ._async.connection_pool import AsyncConnectionPool
 from ._async.http_proxy import AsyncHTTPProxy
-from ._bytestreams import AsyncIteratorByteStream, IteratorByteStream, PlainByteStream
+from ._bytestreams import PlainByteStream
 from ._exceptions import (
     CloseError,
     ConnectError,
@@ -28,11 +28,9 @@ __all__ = [
     "AsyncConnectionPool",
     "AsyncHTTPProxy",
     "AsyncHTTPTransport",
-    "AsyncIteratorByteStream",
     "CloseError",
     "ConnectError",
     "ConnectTimeout",
-    "IteratorByteStream",
     "LocalProtocolError",
     "NetworkError",
     "PlainByteStream",
--- a/httpcore/__init__.py
+++ b/httpcore/__init__.py
@@ -1,4 +1,4 @@
-from ._async.base import AsyncByteStream, AsyncHTTPTransport
+from ._async.base import AsyncHTTPTransport
 from ._async.connection_pool import AsyncConnectionPool
 from ._async.http_proxy import AsyncHTTPProxy
 from ._bytestreams import PlainByteStream
@@ -19,12 +19,11 @@ from ._exceptions import (
     WriteError,
     WriteTimeout,
 )
-from ._sync.base import SyncByteStream, SyncHTTPTransport
+from ._sync.base import SyncHTTPTransport
 from ._sync.connection_pool import SyncConnectionPool
 from ._sync.http_proxy import SyncHTTPProxy
 
 __all__ = [
-    "AsyncByteStream",
     "AsyncConnectionPool",
     "AsyncHTTPProxy",
     "AsyncHTTPTransport",
@@ -40,7 +39,6 @@ __all__ = [
     "ReadError",
     "ReadTimeout",
     "RemoteProtocolError",
-    "SyncByteStream",
     "SyncConnectionPool",
     "SyncHTTPProxy",
     "SyncHTTPTransport",
--- a/httpcore/_async/http11.py
+++ b/httpcore/_async/http11.py
@@ -5,6 +5,7 @@ import h11
 
 from .._backends.auto import AsyncSocketStream
 from .._bytestreams import AsyncIteratorByteStream, PlainByteStream
+from .._compat import asynccontextmanager
 from .._exceptions import LocalProtocolError, RemoteProtocolError, map_exceptions
 from .._types import URL, Headers, TimeoutDict
 from .._utils import get_logger
@@ -47,6 +48,7 @@ class AsyncHTTP11Connection(AsyncBaseHTTPConnection):
         if self.state == ConnectionState.IDLE:
             self.state = ConnectionState.READY
 
+    @asynccontextmanager
     async def arequest(
         self,
         method: bytes,
@@ -54,7 +56,7 @@ class AsyncHTTP11Connection(AsyncBaseHTTPConnection):
         headers: Headers = None,
         stream: AsyncByteStream = None,
         ext: dict = None,
-    ) -> Tuple[int, Headers, AsyncByteStream, dict]:
+    ) -> AsyncIterator[Tuple[int, Headers, AsyncByteStream, dict]]:
         headers = [] if headers is None else headers
         stream = PlainByteStream(b"") if stream is None else stream
         ext = {} if ext is None else ext
@@ -78,7 +80,10 @@ class AsyncHTTP11Connection(AsyncBaseHTTPConnection):
             "http_version": http_version.decode("ascii", errors="ignore"),
             "reason": reason_phrase.decode("ascii", errors="ignore"),
         }
-        return (status_code, headers, response_stream, ext)
+        try:
+            yield (status_code, headers, response_stream, ext)
+        finally:
+            await response_stream.aclose()
 
     async def start_tls(
         self, hostname: bytes, timeout: TimeoutDict = None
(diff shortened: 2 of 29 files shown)


florimondmanca:
@tomchristie It's a big one, but at least this is where I think we want to go, and we can now find ways to possibly get this in incrementally (not sure how, but you always have better ideas than me on that front :-)).

florimondmanca:
There's a follow up to this related to point 3/ in https://github.com/encode/httpx/issues/1274#issuecomment-694884553, being that byte streams now don't really need a `.close()`/`.aclose()` method, since we're able to call any close callback in a `finally: ...` block around the `yield` statement. But I think we should treat it as a follow-up, so that this PR remains focused on the core API change?

httpcore/_async/http11.py
```diff
@@ -72,13 +74,13 @@ async def arequest(
...
             aiterator=self._receive_response_data(timeout),
-            aclose_func=self._response_closed,
         )
         ext = {
             "http_version": http_version.decode("ascii", errors="ignore"),
             "reason": reason_phrase.decode("ascii", errors="ignore"),
         }
-        return (status_code, headers, response_stream, ext)
+        yield (status_code, headers, response_stream, ext)
+        await self._response_closed()
```
florimondmanca, line 83: @tomchristie Not sure this comes from you, do we need a `try/finally` here?

```suggestion
        try:
            yield (status_code, headers, response_stream, ext)
        finally:
            await self._response_closed()
```

tomchristie: I'd thought it wasn't required, but since your comment I've convinced myself that I was wrong, using this...

```python
from contextlib import contextmanager


@contextmanager
def request(method, url):
    print('request() started')
    try:
        yield 200, {}, [b'Hello, world',]
    finally:
        print('request() cleaning up')


def main():
    print('main')
    with request("GET", "https://example.org/") as response:
        raise ValueError('Oops')


main()
```

Versus this...

```python
from contextlib import contextmanager


@contextmanager
def request(method, url):
    print('request() started')
    yield 200, {}, [b'Hello, world',]
    print('request() cleaning up')  # Never runs


def main():
    print('main')
    with request("GET", "https://example.org/") as response:
        raise ValueError('Oops')


main()
```

florimondmanca (commented):
@tomchristie I've gone over the PR again and I'm pretty happy with how this looks. At least implementation-wise this is clean.

httpcore/__init__.py
```diff
@@ -1,7 +1,7 @@
-from ._async.base import AsyncByteStream, AsyncHTTPTransport
+from ._async.base import AsyncHTTPTransport
```
florimondmanca, line 1: I guess we just need to be careful that this drops `SyncByteStream` and `AsyncByteStream`, which we rely on in a few places in HTTPX. Some other implementations out there would also still be relying on these interfaces.

Should we perhaps keep some aliases in place until 1.0…?

```python
from typing import Iterable, AsyncIterable

SyncByteStream = Iterable[bytes]
AsyncByteStream = AsyncIterable[bytes]
```

tomchristie: Yeah I don't think we're sufficiently confident yet about exactly where we're going with this interface,
so I reckon rolling the other stuff (where we *are* confident) and keeping this hanging a little longer.

tomchristie:
Closing in line with https://github.com/encode/httpx/pull/1491

Goodness me but it's been a long road.

I think we can be very happy with the Transport API we're actually going to end up with https://github.com/encode/httpx/pull/1522 - going all the way with context managed resources here just feels like too much of a heavy trade off.

We *could* potentially still choose to use context managed approaches internally within `httpcore` if we wanted, since we can see how to tackle bridging between the two neatly enough. (Eg. see https://github.com/encode/httpx/pull/998)

PR #210, opened 2020-10-04
Add remove-all-unused-imports to autoflake invocation

cdeler:
`flake8` during `script/test` invocation raises an error if we have unused imports.

As `script/lint` already calls `autoflake`, we can add `--remove-all-unused-imports --ignore-init-module-imports` to the call, which might make development process easier

### DIFF
--- a/scripts/lint
+++ b/scripts/lint
@@ -8,7 +8,7 @@ export SOURCE_FILES="httpcore tests"
 
 set -x
 
-${PREFIX}autoflake --in-place --recursive $SOURCE_FILES
+${PREFIX}autoflake --in-place --recursive --remove-all-unused-imports --ignore-init-module-imports $SOURCE_FILES
 ${PREFIX}isort --project=httpcore $SOURCE_FILES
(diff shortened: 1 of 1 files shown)


scripts/lint
```diff
@@ -8,7 +8,7 @@ export SOURCE_FILES="httpcore tests"
 
 set -x
 
-${PREFIX}autoflake --in-place --recursive $SOURCE_FILES
+${PREFIX}autoflake --in-place --recursive --remove-all-unused-imports --ignore-init-module-imports $SOURCE_FILES
```
jcugat, line 11: We can drop this option since all the imports in `__init__.py` files are used.

```suggestion
${PREFIX}autoflake --in-place --recursive --remove-all-unused-imports $SOURCE_FILES
```

cdeler: Sounds reasonable, thank you!

jcugat (approved):

jcugat:
As a followup we might want to add it to httpx too if we find it useful here 👍

PR #215, opened 2020-10-06
Version 0.12

tomchristie:
### Changed

- HTTP header casing is now preserved, rather than always sent in lowercase. (#216 and python-hyper/h11#104)

### Added

- Add Python 3.9 to officially supported versions.

### Fixed

- Gracefully handle a stdlib asyncio bug when a connection is closed while it is in a paused-for-reading state. (#201)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,7 @@ setup(
     zip_safe=False,
     install_requires=["h11==0.*", "sniffio==1.*"],
     extras_require={
-        "http2": ["h2==4.*"],
+        "http2": ["h2>=3,<5"],
     },
     classifiers=[
         "Development Status :: 3 - Alpha",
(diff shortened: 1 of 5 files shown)


florimondmanca:
@tomchristie Shall we have a separate PR for the changes to our `h11` integration? 😇 For the sake of searchability on GitHub PRs, in case we need to dig up things…

setup.py
```diff
@@ -70,6 +70,7 @@ def get_packages(package):
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
```
florimondmanca, line 73: Also this seems to warrant an "Add official support for Python 3.9" PR and changelog entry?

tomchristie:
> Shall we have a separate PR for the changes to our h11 integration? 😇

Keeping me honest. Thanks! 😃

https://github.com/encode/httpcore/pull/216

setup.py
tomchristie: Good call, yup.

PR #216, opened 2020-10-06
Preserve HTTP header casing

tomchristie:
Closes https://github.com/encode/httpx/issues/538

Update `h11` and expose raw HTTP header casing, rather than forced-lowercase.


cdeler (approved):
I was thrilled reading https://github.com/python-hyper/h11/pull/103 and https://github.com/python-hyper/h11/pull/104

Approved

PR #218, opened 2020-10-08
Replacing pytest-cov by coverage

cdeler:
After https://github.com/encode/uvicorn/pull/809 it would be great to introduce the same changes in `httpcore`

Pros:

1. `pytest-cov` makes work in IDE harder, suppressing breakpoints ([link](https://pytest-cov.readthedocs.io/en/latest/debuggers.html))
1. `pytest-cov` overwrites some `coverage` flags (probably to be compatible with other plugins e.g. `pytest-xdist`)

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -11,7 +11,7 @@ if [ -z $GITHUB_ACTIONS ]; then
     scripts/check
 fi
 
-${PREFIX}pytest $@
+${PREFIX}coverage run --debug config -m pytest
 
 if [ -z $GITHUB_ACTIONS ]; then
     scripts/coverage
--- a/scripts/test
+++ b/scripts/test
@@ -11,7 +11,7 @@ if [ -z $GITHUB_ACTIONS ]; then
     scripts/check
 fi
 
-${PREFIX}coverage run --debug config -m pytest
+${PREFIX}coverage run -m pytest
 
 if [ -z $GITHUB_ACTIONS ]; then
     scripts/coverage
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,6 +20,10 @@ known_third_party = brotli,certifi,chardet,cryptography,h11,h2,hstspreload,pytes
 skip = httpcore/_sync/,tests/sync_tests/
 
 [tool:pytest]
-addopts = --cov-report= --cov=httpcore --cov=tests -rxXs
+addopts = -rxXs
 markers =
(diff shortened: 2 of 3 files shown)


setup.cfg
```diff
@@ -20,6 +20,10 @@ known_third_party = brotli,certifi,chardet,cryptography,h11,h2,hstspreload,pytes
 skip = httpcore/_sync/,tests/sync_tests/
 
 [tool:pytest]
-addopts = --cov-report= --cov=httpcore --cov=tests -rxXs
+addopts = -rxXs
```
tomchristie, line 23: I'd love for us to be using long form flags here at some point. I mean, what the heck does *this do*?

(Tho, that's nothing to do with this pull request, just makes me notice it more is all.)

johtso: Magic incantation ✨

scripts/test
```diff
@@ -11,7 +11,7 @@ if [ -z $GITHUB_ACTIONS ]; then
     scripts/check
 fi
 
-${PREFIX}pytest $@
+${PREFIX}coverage run --debug config -m pytest
```
tomchristie, line 14: I don't think we want the `--debug config` here?...

```suggestion
${PREFIX}coverage run -m pytest
```

setup.cfg
cdeler: Let's make the flags wide (wait a second please)

scripts/test
cdeler: Well, yes, it's a bit verbose

setup.cfg
euri10: afaik you cant make reporting flags wide, 
```
  -r chars              show extra test summary info as specified by chars: (f)ailed, (E)rror, (s)kipped, (x)failed, (X)passed, (p)assed, (P)assed with output, (a)ll except passed (p/P), or (A)ll. (w)arnings are enabled by default (see --disable-warnings), 'N' can
                        be used to reset the list. (default: 'fE').
```

cdeler: Ha :-) There is no long option for `-r`. The flag combination means:
`-r`: show extra test summary info as specified by chars: (s)kipped, (x)failed, (X)passed

tomchristie: Well, we oughta treat it separately in any case. Perhaps we don't really want it at all.

tomchristie (approved):
Yup I'm good with this. Thanks!
Would welcome an equivalent PR in `httpx`, too.

PR #219, opened 2020-10-08
[Sandbox] Add failing too-many-files test

florimondmanca:
Sandbox PR to prove that something like https://github.com/encode/httpcore/pull/193#issuecomment-703256971 to #193 or #185 would work to prevent a regression for #182.

### DIFF
--- a/tests/async_tests/test_interfaces.py
+++ b/tests/async_tests/test_interfaces.py
@@ -365,6 +365,10 @@ async def test_explicit_backend_name(server: Server) -> None:
 
 @pytest.mark.anyio
 @pytest.mark.usefixtures("too_many_open_files_minus_one")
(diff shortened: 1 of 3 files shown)


florimondmanca:
Yup 😄 

https://github.com/encode/httpcore/pull/219/checks?check_run_id=1227636776#step:7:300

florimondmanca:
:+1:

https://github.com/encode/httpcore/pull/219/checks?check_run_id=1227750207#step:7:208

```console
    def is_connection_dropped(self) -> bool:
        raw_socket = self.stream.extra(SocketAttribute.raw_socket)
>       rready, _wready, _xready = select.select([raw_socket], [], [], 0)
E       ValueError: filedescriptor out of range in select()
```

PR #221, opened 2020-10-10
Add connect retries

florimondmanca:
Alternative to https://github.com/encode/httpx/pull/1196, prompted by https://github.com/encode/httpx/pull/1196#issuecomment-690148589

Closes https://github.com/encode/httpx/issues/1141

This PR adds off-by-default "retry on connection failure" capability to the connection pool, using a `retries=<int>` option. Also includes an exponential backoff mechanism, with a fixed backoff factor of 0.5 for now (so, 0s, 0.5s, 1.0s, 2.0s, 4.0s, etc).

We add it at the transport layer, as close to opening sockets as possible, so that connect retries don't interact with the connection pooling logic, so that we do "acquire a connection, open a socket, open a socket, …, release the connection".

(We're in a better position now to add this feature now, because we have a `backend` option on the connection pool transport. This PR tweaks it a bit so we can pass a full-fledged backend instance, rather than only a backend name. Then we build and provide a mock backend that raises exception on-demand, for testing purposes.)

(The diff looks quite big, but everything is unasynced, so duplicated, and test code is a bit large too, so…)

Once merged, we'll be able to document this on the HTTPX side, in a new "Retries" section in "Advanced Usage". For now usage will be a bit involved, but once `httpx.HTTPTransport()` lands (https://github.com/encode/httpx/issues/1302) things will be easier UX-wise.

### DIFF
--- a/tests/async_tests/test_retries.py
+++ b/tests/async_tests/test_retries.py
@@ -34,10 +34,7 @@ class AsyncMockBackend(AutoBackend):
 
 async def read_body(stream: httpcore.AsyncByteStream) -> bytes:
     try:
-        body = []
-        async for chunk in stream:
-            body.append(chunk)
-        return b"".join(body)
+        return b"".join([chunk async for chunk in stream])
     finally:
         await stream.aclose()
 
--- a/tests/async_tests/test_retries.py
+++ b/tests/async_tests/test_retries.py
@@ -1,0 +1,152 @@
+import queue
+import time
+from typing import Any, List, Optional
+
+import pytest
+
+import httpcore
+from httpcore._backends.auto import AsyncSocketStream, AutoBackend
+from tests.utils import Server
+
+
+class AsyncMockBackend(AutoBackend):
+    def __init__(self) -> None:
+        super().__init__()
+        self._exceptions: queue.Queue[Optional[Exception]] = queue.Queue()
+        self._timestamps: List[float] = []
+
+    def push(self, *exceptions: Optional[Exception]) -> None:
+        for exc in exceptions:
+            self._exceptions.put(exc)
+
+    def pop_open_tcp_stream_intervals(self) -> list:
+        intervals = [b - a for a, b in zip(self._timestamps, self._timestamps[1:])]
+        self._timestamps.clear()
+        return intervals
+
+    async def open_tcp_stream(self, *args: Any, **kwargs: Any) -> AsyncSocketStream:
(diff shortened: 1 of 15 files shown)


tests/async_tests/test_retries.py
```diff
@@ -0,0 +1,152 @@
...
+            raise exc
+        return await super().open_tcp_stream(*args, **kwargs)
+
+
+async def read_body(stream: httpcore.AsyncByteStream) -> bytes:
+    try:
+        body = []
+        async for chunk in stream:
+            body.append(chunk)
+        return b"".join(body)
```
JayH5, line 40: Nit: I see this `body.append(chunk)` a lot when I think the below is much cleaner:
```suggestion
        return b"".join([chunk async for chunk in stream])
```

JayH5 (commented):
I like this as a most basic (but still useful) implementation of retries 🎉

Only question is whether the parameter should be `retries` or `connect_retries`? Not sure if that discussion has already been had somewhere?

cdeler:
@florimondmanca 

Sorry for the question. You added the retries mechanism only for connection establishing yes?

It's talking about (probably) the possibility, if we may be introducing some sort of `Retry`'s dict as it's done for timeouts' dict?

After that we can carefully handle retry for, e.g. "HTTP POST" requests (allow retries for connect, but do not allow after that)

florimondmanca:
> Only question is whether the parameter should be retries or connect_retries? Not sure if that discussion has already been had somewhere?

I think we've had it already on the HTTPX repo. TLDR was "let's use `retries` so we can expand its signature if needed w/o having to add an extra option". So eg in top of `retries=<int>`, we can easily expand to support `retries=(int, int, ...)` (or a dict) for any kinds of retries we'd like to support. Same if we want to make the exponential back off configurable at some point. And then we wrap it up in a handy `httpx.Retries` class on the HTTPX side, like we do for `httpx.Timeout`. :)

tests/async_tests/test_retries.py
florimondmanca: Yes, fair enough, just copy-pasted some code from other modules. :-)

JayH5 (approved):
I'm happy with this 😄

cdeler (approved):

tomchristie:
Fantastic, lovely piece of work!

And yup, once we've had a good look at encode/httpx#1302 we ought to be in a better position wrt. usability for this too.

PR #223, opened 2020-10-14
Fixed the problem with socket, which might leak when tunnel proxy connection couldn't be establish

cdeler:
Fixes https://github.com/encode/httpx/issues/1360

I extracted a tunnel proxy connection establishing into the separated method, and wrapped it by `try`-`except` statement, closing the `proxy_connection` if it's necessary.

Also I added the unit test and a test fixture from https://github.com/encode/httpx/issues/1360 comments

### DIFF
--- a/httpcore/_async/http_proxy.py
+++ b/httpcore/_async/http_proxy.py
@@ -237,21 +237,21 @@ class AsyncHTTPProxy(AsyncConnectionPool):
                 # We assume the target speaks TLS on the specified port
                 if scheme == b"https":
                     await proxy_connection.start_tls(host, timeout)
-
-                # The CONNECT request is successful, so we have now SWITCHED PROTOCOLS.
-                # This means the proxy connection is now unusable, and we must create
-                # a new one for regular requests, making sure to use the same socket to
-                # retain the tunnel.
-                connection = AsyncHTTPConnection(
-                    origin=origin,
-                    http2=self._http2,
-                    ssl_context=self._ssl_context,
-                    socket=proxy_connection.socket,
-                )
-                await self._add_to_pool(connection, timeout)
             except Exception as exc:
                 await proxy_connection.aclose()
                 raise ProxyError(exc)
+
+            # The CONNECT request is successful, so we have now SWITCHED PROTOCOLS.
+            # This means the proxy connection is now unusable, and we must create
+            # a new one for regular requests, making sure to use the same socket to
+            # retain the tunnel.
+            connection = AsyncHTTPConnection(
+                origin=origin,
+                http2=self._http2,
+                ssl_context=self._ssl_context,
+                socket=proxy_connection.socket,
+            )
+            await self._add_to_pool(connection, timeout)
 
         # Once the connection has been established we can send requests on
         # it as normal.
--- a/httpcore/_async/http_proxy.py
+++ b/httpcore/_async/http_proxy.py
@@ -207,41 +207,51 @@ class AsyncHTTPProxy(AsyncConnectionPool):
             connect_url = self.proxy_origin + (target,)
             connect_headers = [(b"Host", target), (b"Accept", b"*/*")]
             connect_headers = merge_headers(connect_headers, self.proxy_headers)
-            (proxy_status_code, _, proxy_stream, _) = await proxy_connection.arequest(
-                b"CONNECT", connect_url, headers=connect_headers, ext=ext
-            )
 
-            proxy_reason = get_reason_phrase(proxy_status_code)
-            logger.trace(
-                "tunnel_response proxy_status_code=%r proxy_reason=%r ",
-                proxy_status_code,
-                proxy_reason,
-            )
-            # Read the response data without closing the socket
-            async for _ in proxy_stream:
-                pass
+            try:
+                (
+                    proxy_status_code,
+                    _,
+                    proxy_stream,
+                    _,
+                ) = await proxy_connection.arequest(
+                    b"CONNECT", connect_url, headers=connect_headers, ext=ext
+                )
 
-            # See if the tunnel was successfully established.
-            if proxy_status_code < 200 or proxy_status_code > 299:
-                msg = "%d %s" % (proxy_status_code, proxy_reason)
-                raise ProxyError(msg)
+                proxy_reason = get_reason_phrase(proxy_status_code)
+                logger.trace(
+                    "tunnel_response proxy_status_code=%r proxy_reason=%r ",
+                    proxy_status_code,
+                    proxy_reason,
+                )
+                # Read the response data without closing the socket
+                async for _ in proxy_stream:
+                    pass
 
-            # Upgrade to TLS if required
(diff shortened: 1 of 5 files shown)


cdeler:
In the test case I had to filter out any warnings from https://github.com/encode/httpx/issues/825

The couldn't appear in this test case (as it's run under `trio`), so I felt free to do that

httpcore/_async/http_proxy.py
```diff
@@ -172,33 +172,22 @@ async def _forward_request(
 
         return status_code, headers, wrapped_stream, ext
 
-    async def _tunnel_request(
```
tomchristie, line 175: Great stuff! So...

I think the first thing would be to try to get the change footprint of this PR down a bit.

* I'd suggest we *keep* a single `_tunnel_request` method, rather than splitting out into two methods here?
* Can we catch the exception more narrowly? Perhaps we could have a top-level `TransportError` class, in the same way as `httpx`. Then, in cases like this we could *just* catch `TransportError`.

tomchristie: Actually thinking about it let's treat that second line item *independently* of this pull request.

cdeler: Do I understand correctly the plan:
1. to introduce the root exception for all `httpcore` exceptions, to name the class `TransportError`. This change should go in the separate PR
2. as soon as (1) has been merged, to merge `_establish_tunnel_proxy_connection` back into `_tunnel_request`, catching `TransportError` instead of `Exception`?

Talking about (2) we can try. But also it means that there are cases (some error conditions) when we do not clean up resources. The conditions are exceptions not inherited by `TransportError`...

tomchristie: No don't worry about (2) at all. We can think about that some other day.

cdeler: Therefore I'm to just eliminate `_establish_tunnel_proxy_connection`, moving this code back to `_tunnel_request` to reduce the PR footprint, yes? _(sorry for repeating questions, I want to fully understand you)_

cdeler: @tomchristie 
I did that, so the PR became significantly smaller (but it's required to compare it ignoring whitespace changes)

tests/async_tests/test_interfaces.py
```diff
@@ -191,6 +191,36 @@ async def test_http_proxy(
...
+async def test_proxy_socket_does_not_leak_when_the_connection_hasnt_been_added_to_pool(
+    proxy_server: URL,
+    server: Server,
+    proxy_mode: str,
+    protocol: bytes,
+    port: int,
+):
+    method = b"GET"
+    url = (protocol, b"example.com", port, b"/")
+    headers = [(b"host", b"example.org")]
```
florimondmanca, line 206: Is the host mismatch between url / headers expected? I think this is somewhat related to the _block file_ passed to `pproxy`. If so, ad assuming the block file is intended to be a list of URL hosts that _cannot_ be requested through the proxy…

* Should we set the blocked hostname to something more explicit like `blockedhost.org`?
* If the different host in `url` and `headers` is on purpose and required, add a small comment to explain why?

tests/async_tests/test_interfaces.py
```diff
@@ -191,6 +191,36 @@ async def test_http_proxy(
...
+                    _ = await http.arequest(method, url, headers)
+                except httpcore.RemoteProtocolError:
+                    pass
+
+    # have to filter out https://github.com/encode/httpx/issues/825 from other tests
+    warnings_list = [
+        *filter(lambda warn: "asyncio" not in warn.filename, recorded_warnings.list)
+    ]
+
+    assert len(warnings_list) == 0
```
florimondmanca, line 221: (Nit) (Assuming this doesn't conflict with a top-level `import warnings`?)

```suggestion
    warnings = [
        *filter(lambda warn: "asyncio" not in warn.filename, recorded_warnings.list)
    ]

    assert len(warnings) == 0
```

httpcore/_async/http_proxy.py
```diff
@@ -207,41 +207,51 @@ async def _tunnel_request(
...
+                # This means the proxy connection is now unusable, and we must create
+                # a new one for regular requests, making sure to use the same socket to
+                # retain the tunnel.
+                connection = AsyncHTTPConnection(
+                    origin=origin,
+                    http2=self._http2,
+                    ssl_context=self._ssl_context,
+                    socket=proxy_connection.socket,
+                )
+                await self._add_to_pool(connection, timeout)
```
florimondmanca, line 251: Do we want to explicitly put this operation _outside_ the `try`/`except` block? I assume the only operations that we want to fail upon is socket-related stuff, so `proxy_connection.arequest()` and `proxy_connection.start_tls()`. Anything that's not related to it, such as adding to the pool, should be left outside the block, correct?

tests/utils.py
```diff
@@ -48,13 +49,34 @@ def host_header(self) -> Tuple[bytes, bytes]:
...
+            _wait_can_connect(proxy_host, proxy_port)
+
+            yield b"http", proxy_host.encode(), proxy_port, b"/"
+        finally:
+            if proc is not None:
+                proc.kill()
+
+
+@contextlib.contextmanager
+def create_proxy_block_file(blocked_domains: List[str]):
```
florimondmanca, line 73: I'd suggest we add a small docstring here with any relevant docs material from `pproxy` (eg a link to relevant docs) describing what a "block file" is and what effect it has?

florimondmanca (commented):
Interesting, and a bit fiddly, so I'm being a bit careful :-)

httpcore/_async/http_proxy.py
```diff
@@ -207,41 +207,51 @@ async def _tunnel_request(
...
+            except Exception:
+                await proxy_connection.aclose()
+                raise
```
florimondmanca, line 254: Also, as I think @tomchristie mentioned, should we raise an `httpcore`-specific exception in all cases here? For example wrapping around `ProxyError`:

```suggestion
            except Exception as exc:
                await proxy_connection.aclose()
                raise ProxyError(exc)
```

tests/async_tests/test_interfaces.py
cdeler: You are right! I missed that the url and header are different, and it might confuse. Moreover the server name is not so clear. I'm renaming this host to "blockedhost.example.com" (and changing the fixture the same way)

tests/utils.py
cdeler: I added some docs to `http_proxy_server` and `create_proxy_block_file`

httpcore/_async/http_proxy.py
cdeler: Yes, you are right

cdeler:
@florimondmanca 
Thank you for the review, I fixed all you mentioned

florimondmanca (approved):
Rad :) Great one!

PR #233, opened 2020-11-07
Use Hypercorn for UDS tests

florimondmanca:
Follow-up to #205

Use `hypercorn` for UDS tests. Had to refactor `HypercornServer` a bit to remove the `host/port` options, and use `bind` in all cases.

We can't ditch `uvicorn` entirely because `hypercorn` is 3.7+ only, so we only install and use `uvicorn` on 3.6.

The great part here is that we now use the `Server` interface across all tests.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,4 +29,4 @@ pproxy==2.3.7
 pytest==6.1.2
 pytest-trio==0.7.0
 trustme==0.6.0
-uvicorn==0.12.1
+uvicorn==0.12.1; python_version < '3.7'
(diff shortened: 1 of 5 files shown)


florimondmanca:
We've got a flake.

```console
 >           assert backend.pop_open_tcp_stream_intervals() == [
                pytest.approx(0, abs=1e-3),  # Retry immediately.
                pytest.approx(0.5, rel=0.1),  # First backoff.
                pytest.approx(1.0, rel=0.1),  # Second (increased) backoff.
            ]
E           assert [0.0010938644...1297950744629] == [0 ± 1.0e-03,...1.0 ± 1.0e-01]
E             At index 0 diff: 0.0010938644409179688 != 0 ± 1.0e-03
E             Use -v to get the full diff
```

JayH5 (approved):
Nice!

PR #234, opened 2020-11-07
Fix flaky test

florimondmanca:
Prompted by https://github.com/encode/httpcore/pull/233#issuecomment-723445718

It's _probably_ due to Hypercorn (#205), that may be a tiny bit slower, and so this is most likely affecting master too.


JayH5 (approved):
Wondering what about Hypercorn may be making this slightly slower 🤔

florimondmanca:
Okay well, actually `master` is still looking good now. I'm thinking it might have been a temporary slow-down of GitHub Actions' machines, or something… But still, we can merge this so we avoid any new failures in the future.

PR #235, opened 2020-11-13
Properly wrap asyncio connect errors

cdeler:
Closes https://github.com/encode/httpx/issues/1387

Fixed the problem with `asyncio` backend, when it incorrectly handles connection timeouts.

### DIFF
--- a/httpcore/_backends/asyncio.py
+++ b/httpcore/_backends/asyncio.py
@@ -107,15 +107,18 @@ class SocketStream(AsyncSocketStream):
 
         loop_start_tls = getattr(loop, "start_tls", backport_start_tls)
 
-        transport = await asyncio.wait_for(
-            loop_start_tls(
-                transport,
-                protocol,
-                ssl_context,
-                server_hostname=hostname.decode("ascii"),
-            ),
-            timeout=timeout.get("connect"),
-        )
+        exc_map = {asyncio.TimeoutError: ConnectTimeout, OSError: ConnectError}
+
+        with map_exceptions(exc_map):
+            transport = await asyncio.wait_for(
+                loop_start_tls(
+                    transport,
+                    protocol,
+                    ssl_context,
+                    server_hostname=hostname.decode("ascii"),
+                ),
+                timeout=timeout.get("connect"),
+            )
 
         stream_reader.set_transport(transport)
         stream_writer = asyncio.StreamWriter(


elonzh:
If we need an ordered dict, why not use `OrderedDict`?

cdeler:
I forgot that dicts are not ordered in 'python3.6'

tests/async_tests/test_interfaces.py
```diff
@@ -228,6 +228,10 @@ async def test_proxy_socket_does_not_leak_when_the_connection_hasnt_been_added_t
         *filter(lambda warn: "asyncio" not in warn.filename, recorded_warnings.list)
     ]
 
+    if warnings:
```
elonzh, line 231: I'm not familiar with the project code, but this code looks meaningless

tests/sync_tests/test_interfaces.py
```diff
@@ -228,6 +228,10 @@ def test_proxy_socket_does_not_leak_when_the_connection_hasnt_been_added_to_pool
         *filter(lambda warn: "asyncio" not in warn.filename, recorded_warnings.list)
     ]
 
+    if warnings:
```
elonzh, line 231: and here

tests/async_tests/test_interfaces.py
cdeler: It's for CI debugging. You can check the last failed build, this line prints a warning. It means that I forgot to close a connection in the newly added test

elonzh: the pytest will show variables when assertion fail.

httpcore/_backends/anyio.py
```diff
@@ -167,11 +170,13 @@ async def open_uds_stream(
...
-            TimeoutError: ConnectTimeout,
-            BrokenResourceError: ConnectError,
-        }
+        exc_map = OrderedDict(
+            [
+                (TimeoutError, ConnectTimeout),
+                (OSError, ConnectError),
+                (BrokenResourceError, ConnectError),
+            ]
+        )
```
florimondmanca, line 179: Dicts are officially ordered since 3.7, and have been in CPython as an implementation detail since 3.5.

So… Could we do without `OrderedDict`, at least _for now_*?

```suggestion
        exc_map = {
            TimeoutError: ConnectTimeout,
            OSError: ConnectError,
            BrokenResourceError: ConnectError,
        }
```

(*I understand this wouldn't be guaranteed to work on eg Pypy, and this may not convey the idea that "order matters" particularly well. There's a case to be made to perhaps switch `exc_map` to accept a _sequence_ of 2-tuples, but we'd need to treat it as a follow-up to keep this PR focused on fixing the bug, rather than any extra refactoring.)

florimondmanca (changes requested):
Thanks!

tests/utils.py
```diff
@@ -197,3 +197,40 @@ def create_proxy_block_file(blocked_domains: List[str]):
         file.flush()
 
         yield file.name
+
+
+@contextlib.contextmanager
+def slow_app(host: str, port: int):
```
florimondmanca, line 203: There's probably a way we could achieve this as an ASGI app?

It's not super clear to me what in this code makes the app "slow". I assume it's that that we `sleep(0.5)` and the test uses a `connect` timeout of 0.1?

tests/conftest.py
```diff
@@ -25,6 +25,9 @@
     SERVER_HTTPS_PORT = 8003
     HTTPS_SERVER_URL = f"https://localhost:{SERVER_HTTPS_PORT}"
 
+SLOW_SERVER_HOST = "localhost"
+SLOW_SERVER_HTTPS_PORT = 8004
```
florimondmanca, line 29: The `slow_server` fixture isn't actually using HTTPS:

```suggestion
SLOW_SERVER_PORT = 8004
```

tests/conftest.py
```diff
@@ -161,6 +164,15 @@ def https_server(
         yield server
 
 
+@pytest.fixture(scope="session")
+def slow_https_server(
+    localhost_cert_pem_file: str, localhost_cert_private_key_file: str
+):
+
+    with slow_app(SLOW_SERVER_HOST, SLOW_SERVER_HTTPS_PORT):
+        yield SLOW_SERVER_HOST.encode(), SLOW_SERVER_HTTPS_PORT
```
florimondmanca, line 173:
```suggestion
def slow_https_server():
    with slow_app(SLOW_SERVER_HOST, SLOW_SERVER_PORT):
        yield SLOW_SERVER_HOST.encode(), SLOW_SERVER_PORT
```

florimondmanca (commented):
Some more thoughts…

tests/utils.py
florimondmanca: > There's probably a way we could achieve this as an ASGI app?

…But `hypercorn` only allows us to test that on 3.7+. Hmm, that's actually a real bummer that we can't use Hypercorn to test against arbitrary apps on 3.6.

tests/async_tests/test_interfaces.py
```diff
@@ -466,3 +466,21 @@ async def test_cannot_connect_uds(backend: str) -> None:
...
+    method = b"GET"
+    host, port = slow_https_server
+    url = (b"https", host, port, b"/")
+    timeouts = {"connect": 0.1}
+
+    async with httpcore.AsyncHTTPProxy(proxy_server, backend=backend) as http:
+        with pytest.raises(httpcore.ProxyError) as cm:
+            await http.arequest(method, url, ext={"timeout": timeouts})
+
+        print(cm.value)
```
florimondmanca, line 486:
```suggestion
```

tests/utils.py
```diff
@@ -197,3 +197,40 @@ def create_proxy_block_file(blocked_domains: List[str]):
...
+            sock.close()
+
+    thread = threading.Thread(target=run)
+
+    try:
+        thread.start()
+        yield
+        stop = True
+    finally:
+        thread.join()
```
florimondmanca, line 236: Should this be…?

```suggestion
    finally:
        stop = True
        thread.join()
```

(If an exception occurs around the `yield`, we _do_ want `stop` to be set, otherwise we'd have a deadlock as the thread would never stop.)

florimondmanca (commented):
I do appreciate the extra E2E test here, but to be honest this is a fairly big change footprint — so for the record, I would be very much okay with us merging just the fix in `_backends`.

cdeler:
@florimondmanca you are right. I'm changing it as soon as I reach my laptop

httpcore/_backends/anyio.py
cdeler: Yes, I'd get rid of `OrderedDict`.

Anyway I throw away evetyhing not related with the `asyncio` changes

florimondmanca:
@cdeler Thanks for the update!

> But the problem with the exceptions order is still actual for anyio backend in case of tcp & uds connection timeouts

So, do we still need to keep the change to the `anyio` backend?

cdeler:
@florimondmanca I'd extract it into separate PR with the test

cdeler:
@florimondmanca I added another PR with the `anyio` changeset and tests: https://github.com/encode/httpcore/pull/236

Working on this PR I **didn't find** another tests verified timeouts on I/O routines

florimondmanca (approved):

florimondmanca:
@cdeler Thanks. I retitled this PR according to what this seems to be doing. :-)

PR #236, opened 2020-11-16
Fix connect timeouts for anyio backend

cdeler:
### Why to add changes
In https://github.com/encode/httpx/issues/1387#issuecomment-726775441 I found that `anyio` backend has problem with connection timeouts [here](https://github.com/encode/httpcore/blob/db81ed09592bac8ce740b3be74cf99c86a68c104/httpcore/_backends/anyio.py#L170-L186):

```python
        exc_map = {
            OSError: ConnectError,
            TimeoutError: ConnectTimeout,
            BrokenResourceError: ConnectError,
        }

        with map_exceptions(exc_map):
            async with anyio.fail_after(connect_timeout):
                stream: anyio.abc.ByteStream
                stream = await anyio.connect_tcp(
                    unicode_host, port, local_host=local_address
                )
                if ssl_context:
                    stream = await TLSStream.wrap(
                        stream,
                        hostname=unicode_host,
                        ssl_context=ssl_context,
                        standard_compatible=False,
                    )

```

As `assert isinstance(TimeoutError(), OSError)`, the above `map_exceptions` throws `ConnectError` in case of any timeout (at least in `await TLSStream.wrap`)

### What has been done
I added 2 tests (`test_connection_timeout_tcp` and `test_connection_timeout_uds`) and fixed the error

#### Note

`test_connection_timeout_uds` has been marked as `python >= 3.7` to reduce changeset footprint in the `slow_uds_server` fixture (otherwise there should be a fallback Uvicorn server)

### DIFF
--- a/tests/async_tests/test_interfaces.py
+++ b/tests/async_tests/test_interfaces.py
@@ -1,4 +1,5 @@
 import platform
+import sys
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -47,6 +51,14 @@ async def app(scope: dict, receive: typing.Callable, send: typing.Callable) -> N
     await send({"type": "http.response.body", "body": b"Hello, world!"})
 
 
+async def slow_app(
```
florimondmanca, line 54: Actually, let's consider updating `app` so it just switches on the requested `path`:

```python
async def app(scope, receive, send):
    path = scope["raw_path"]
    if path.startswith(b"/slow"):
        await slow(scope, receive, send)
    else:
        await hello_world(scope, receive, send)


async def hello_world(scope, receive, send):
    ...


async def slow(scope, receive, send):
    path = scope["raw_path"]
    seconds = int(path.split(b"/")[1].decode())
    await trio.sleep(seconds)
    await hello_world(scope, receive, send)
```

=> Then we can just reuse the existing server fixtures — no need to setup anything new.

tests/async_tests/test_interfaces.py
```diff
@@ -466,3 +467,34 @@ async def test_cannot_connect_uds(backend: str) -> None:
...
+
+
+@pytest.mark.skipif(
+    sys.version_info[:2] < (3, 7), reason="Hypercorn doesn't support python < 3.7"
+)
+@pytest.mark.anyio
+async def test_connection_timeout_uds(backend: str, slow_uds_server: Server) -> None:
+    uds = slow_uds_server.uds
+    method = b"GET"
+    url = (b"https", b"localhost", None, b"/delay/5")
```
florimondmanca, line 494: Currently this `/delay/5` path isn't doing anything special, but I agree it's good to be able to specify it from the test. In my other comment I suggest parsing the amount of seconds from the requested URL path, so that requesting this `/delay/5` would effectively make the test server respond after 5 seconds.

florimondmanca (commented):
Looking great! Left a comment to simplify things on the test setup side. :+1:

tests/conftest.py
cdeler: Good idea. I've done with it

cdeler:
@florimondmanca 
As at `hypercorn` level I cannot manage at socket level (e.g. I cannot "accept a connection after small delay"), I had to do a work-around. I'm trying to access http server using https url. It causes a timeout inside 
```python
stream = await TLSStream.wrap(
    stream,
    hostname=unicode_host,
    ssl_context=ssl_context,
    standard_compatible=False,
)
```
what is expected by me. 

On other hand it means that all the work around fixtures was redundant. Should I remove all the `conftest.py` changes from this PR?

florimondmanca:
@cdeler Yes, if just trying to connect via `https` to the HTTP server is enough to trigger a timeout, then let's drop the rest and anything we don't need. I'm all for reducing the footprint of PRs whenever possible. :-)

cdeler:
@florimondmanca I removed all unnecessary fixtures. But I still have to skip the test on 3.6 since testing against example.com behaves different

florimondmanca (approved):
Great, thanks!

PR #237, opened 2020-11-20
Fix import of HTTP/1.1 connection causing thread-safety bug on 3.9

florimondmanca:
Fixes https://github.com/encode/httpx/issues/1393

Moves the local import of the HTTP/1.1 connection class to a global one. It wasn't necessary (`h11` is always installed), and it's a non-threadsafe trick which started causing `ImportError` on Python 3.9 (that it worked in 3.8 and before was actually considered as a bug: https://bugs.python.org/issue35943).

Also adds a basic test case to verify that `httpcore.SyncConnectionPool` can indeed be used in thread-concurrent environments.

Local testing showed that this is not a problem for `h2`, so keeping that local import as-is for now.

cc @Senpos

### DIFF
--- a/tests/test_threadsafety.py
+++ b/tests/test_threadsafety.py
@@ -1,0 +1,46 @@
+import concurrent.futures
+
+import pytest
+
+import httpcore
+
+from .utils import Server
+
+
+def read_body(stream: httpcore.SyncByteStream) -> bytes:
+    try:
+        return b"".join(chunk for chunk in stream)
+    finally:
+        stream.close()
+
+
+@pytest.mark.parametrize(
+    "http2", [pytest.param(False, id="h11"), pytest.param(True, id="h2")]
+)
+def test_threadsafe_basic(server: Server, http2: bool) -> None:
+    """
+    The sync connection pool can be made to perform requests concurrently using
+    threads.
+
+    Also a regression test for: https://github.com/encode/httpx/issues/1393
+    """
+    with httpcore.SyncConnectionPool(http2=http2) as http:
+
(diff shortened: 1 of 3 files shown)


euri10 (approved):
very interesting, wasn't aware of this local / non-thread safe thing

tests/test_threadsafety.py
```diff
@@ -0,0 +1,46 @@
...
+        with concurrent.futures.ThreadPoolExecutor(max_workers=4) as executor:
+            futures = [executor.submit(request, http) for _ in range(10)]
+            num_results = 0
+
+            for future in concurrent.futures.as_completed(futures):
+                status_code = future.result()
+                assert status_code == 200
+                num_results += 1
+
+            assert num_results == 10
```
cdeler, line 46: Should we close the executor before running `concurrent.futures.as_completed` ? (I don't exactly remember why, but I did it every time I used executors. Probably to not allow new submissions before `as_completed`
```suggestion
        num_results = 0

        for future in concurrent.futures.as_completed(futures):
            status_code = future.result()
            assert status_code == 200
            num_results += 1

        assert num_results == 10
```

cdeler (approved):
Great PR, you rock 🚀

tests/test_threadsafety.py
Senpos: They don't do that in [this example](https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor-example) in `concurrent.futures` docs, is that necessary?

florimondmanca: Looks like these operations should be executed within the context manager like we do currently; according to the docs on exit threads are "cleaned up".

Issue #240, opened 2020-11-25
Host header always overwritten in HTTP/2


chripede:
I have a use-case where I want to use an http proxy and resolve the IP address of the target on my side of the proxy.

```
url = "https://1.2.3.4/"
client.headers.update({"Host": "domain.com"})
client.get(url, verify=False)
```

However in HTTP/2 mode, the host header is always removed. See https://github.com/encode/httpcore/blob/84c7738de7fd23ec20187964a33b5b9a0de01493/httpcore/_sync/http2.py#L327

It works as expected in HTTP/1.1

tomchristie:
Right, yup. We drop the host header, because HTTP/2 doesn't use it. Instead it has the `:authority` pseudo header...

https://tools.ietf.org/html/rfc7540

> Clients that generate HTTP/2 requests directly SHOULD use the ":authority" pseudo-header field instead of the Host header field.

However, we probably ought to give precedence to the value of `headers["Host"]` when setting the `:authority` pseudo header. Eg...

https://github.com/encode/httpcore/blob/84c7738de7fd23ec20187964a33b5b9a0de01493/httpcore/_sync/http2.py#L317-L320

Should instead perhaps be, this?...

```python
if port is None or port == default_port:
    authority = hostname
else:
    authority = b"%s:%d" % (hostname, port)

for k, v in headers:
    if k == b"host":
        authority = v
````

chripede:
That would be great. That would make it possible to always use Host and not worry about 1.1 or 2.

Btw I tried setting the ":authority" header as well, but ended up with two of them.

Do you need me to submit a pull request?

tomchristie:
> Do you need me to submit a pull request?

It'd be great if you could, yup.

chripede:
I submitted a pull request, but wanted some input before fixing for async as well.

My problem with this is that Host is always set, so it will always override the hostname. I don't know if that is intended.

florimondmanca:
@chripede I just saw your comment here:

> My problem with this is that Host is always set, so it will always override the hostname. I don't know if that is intended.

Do you mean that HTTPX always passes a `Host` header, and so with the changes from #241 we will actually end up always using that instead of the URL `hostname`? I _think_ that's actually beneficial, since I see `Host` and `:authority` as closer in goal than the URL `hostname` and `:authority`. But reviewing the definition of `:authority` might help…

florimondmanca:
[...]

PR #241, opened 2020-11-25
Allow Host to override :authority in HTTP2

chripede:
#240

### DIFF
--- a/httpcore/_async/http2.py
+++ b/httpcore/_async/http2.py
@@ -325,10 +325,10 @@ class AsyncHTTP2Stream:
                 authority = v
                 break
 
-        if not authority:
+        if authority is None:
             authority = hostname
 
-        if port and port != default_port:
+        if port is not None and port != default_port:
             authority = b"%s:%d" % (authority, port)
 
         headers = [
--- a/httpcore/_async/http2.py
+++ b/httpcore/_async/http2.py
@@ -336,7 +336,19 @@ class AsyncHTTP2Stream:
             (b":authority", authority),
             (b":scheme", scheme),
             (b":path", path),
-        ] + [(k, v) for k, v in headers if k not in (b"host", b"transfer-encoding", b":method", b":authority", b":scheme", b":path")]
+        ] + [
+            (k, v)
+            for k, v in headers
+            if k
+            not in (
+                b"host",
+                b"transfer-encoding",
+                b":method",
+                b":authority",
+                b":scheme",
+                b":path",
+            )
+        ]
         end_stream = not has_body
 
         await self.connection.send_headers(self.stream_id, headers, end_stream, timeout)
--- a/httpcore/_async/http2.py
+++ b/httpcore/_async/http2.py
@@ -314,17 +314,29 @@ class AsyncHTTP2Stream:
     ) -> None:
         scheme, hostname, port, path = url
         default_port = {b"http": 80, b"https": 443}.get(scheme)
-        if port is None or port == default_port:
+
+        # In HTTP/2 the ':authority' pseudo-header is used instead of 'Host'.
+        # For convenience we automatically treat any 'Host' header value
+        # as a value for ':authority', allowing users to pass 'Host' whenever they
+        # want to customize the host header regardless of the HTTP version in use.
+        authority = None
+        for k, v in headers:
+            if k == b"host":
+                authority = v
+                break
+
+        if not authority:
             authority = hostname
-        else:
-            authority = b"%s:%d" % (hostname, port)
+
+        if port and port != default_port:
+            authority = b"%s:%d" % (authority, port)
 
         headers = [
             (b":method", method),
             (b":authority", authority),
             (b":scheme", scheme),
             (b":path", path),
-        ] + [(k, v) for k, v in headers if k not in (b"host", b"transfer-encoding")]
+        ] + [(k, v) for k, v in headers if k not in (b"host", b"transfer-encoding", b":method", b":authority", b":scheme", b":path")]
         end_stream = not has_body
 
         await self.connection.send_headers(self.stream_id, headers, end_stream, timeout)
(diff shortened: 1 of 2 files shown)


cdeler:
Hello @chripede 

You changes are discarded by CI. It happens because some of our files are autogenerated, for example `httpcore/_sync/http2.py` is generated from `httpcore/_async/http2.py`. You can do it locally using the script named `lint` (it's in the project repository [scripts/lint](https://github.com/encode/httpcore/blob/master/scripts/lint))

You can make the requested in the async file, and the edit sync using the `lint` script

Also there is the 2 script named `script/test`, which can help you to run CI locally (just to check if CI is ok locally)

httpcore/_sync/http2.py
```diff
@@ -314,17 +314,25 @@ def send_headers(
     ) -> None:
         scheme, hostname, port, path = url
         default_port = {b"http": 80, b"https": 443}.get(scheme)
-        if port is None or port == default_port:
+
+        authority = None
```
florimondmanca, line 318: Let's consider adding a code comment to explain what's going on here — I think it can be helpful to remind the reader that HTTP/2 doesn't use `Host` but has `:authority`.

```suggestion
        # In HTTP/2 the ':authority' pseudo-header is used instead of 'Host'.
        # For convenience we automatically treat any 'Host' header value
        # as a value for ':authority', allowing users to pass 'Host' whenever they
        # want to customize the host header regardless of the HTTP version in use.
        authority = None
```

florimondmanca (commented):
Looking good! Left a suggestion to add an explanatory comment.

As @cdeler hinted, the workflow in this repo is to A/ edit files in the `_async/` directory, then B/ run `scripts/lint` to generate the sync version. If following this workflow you should be able to get CI back in check. :-)

chripede:
For some reason Black formats the async file different than the sync file. Fixing one linting problem introduces another. Any help would be appreciated ;)

chripede:
Nevermind the last comment. Fixing by hand worked and Black doesn't return it to a one-liner.

cdeler (approved):
Looks good for me 🍦

httpcore/_async/http2.py
```diff
@@ -314,17 +314,41 @@ async def send_headers(
...
+        # For convenience we automatically treat any 'Host' header value
+        # as a value for ':authority', allowing users to pass 'Host' whenever they
+        # want to customize the host header regardless of the HTTP version in use.
+        authority = None
+        for k, v in headers:
+            if k == b"host":
+                authority = v
+                break
+
+        if not authority:
```
florimondmanca, line 328: Let's try to stick to `is None` / `is not None` checks as closely as we can; in the past we've had issues related to eg passing or not passing empty values or zeroes, so checking specifically for None if that's the intent should be preferable.

```suggestion
        if authority is None:
```

httpcore/_async/http2.py
```diff
@@ -314,17 +314,41 @@ async def send_headers(
...
             authority = hostname
-        else:
-            authority = b"%s:%d" % (hostname, port)
+
+        if port and port != default_port:
```
florimondmanca, line 331:
```suggestion
        if port is not None and port != default_port:
```

florimondmanca (changes requested):
Small tweak (almost a nitpick, I'm aware), but this is looking good! Any chance we could add a unit test?

chripede:
I would create a unit test, but I am unable to run the current set of tests

image

florimondmanca:
@chripede That's hum, interesting. It looks like you're showing the stdout output (`pytest -s`). What does the pytest final output look like — which tests are failing?

florimondmanca (approved):
A unit test would be fab but I'm good with merging this as-is too, so, approving.

cdeler:
@florimondmanca 
I'm merging this ticket as it's approved.
I think it's ok to add new tests in an extra PR linked to https://github.com/encode/httpcore/issues/240

httpcore/_async/http2.py
```diff
@@ -314,17 +314,41 @@ async def send_headers(
...
+            if k
+            not in (
+                b"host",
+                b"transfer-encoding",
+                b":method",
+                b":authority",
+                b":scheme",
+                b":path",
+            )
+        ]
```
tomchristie, line 351: I'd prefer if we didn't change the pseudo-header behaviour as part of this PR.

There's some complexities here that'd need further discussion. Since this change isn't necessary for the rest of the pull request to make sense, let's treat anything there independantly.

httpcore/_async/http2.py
```diff
@@ -314,17 +314,41 @@ async def send_headers(
...
+                authority = v
+                break
+
+        if authority is None:
             authority = hostname
-        else:
-            authority = b"%s:%d" % (hostname, port)
+
+        if port is not None and port != default_port:
+            authority = b"%s:%d" % (authority, port)
```
tomchristie, line 332: This case shouldn't override an explicit `Host` header, so I think there's a bit of tweaking needed here.

httpcore/_async/http2.py
chripede: I agree that it isn't part of this fix, but I changed it because I tried adding ":authority" to the headers and got an exception because of duplicate headers.

httpcore/_async/http2.py
chripede: It doesn't. At this point, authority has the value of an explicit Host header.

httpcore/_async/http2.py
tomchristie: Sure thing. If we want to tackle that let's do so separately tho.

For instance, if we are going to strip pseudo headers like this, then we'd probably also want to do the same for HTTP/1.1 codepaths.
 
It's part of a more general conversation about "which header values should be treated as valid".

httpcore/_async/http2.py
tomchristie: Right, tho we don't want to append the port here.

Issue #239, opened 2020-11-24
AttributeError 'NoneType' object has no attribute 'fileno'.


povilasb:
We somewhat discussed the issue at https://gitter.im/encode/community .

TL;DR when connection is closed/reset `httpcore` assumes `socket` is still available for transport and crashes with 

> AttributeError 'NoneType' object has no attribute 'fileno'.

## Environment

Docker image: python:3.8.6-buster
Python: 3.8.6
httpcore: 0.12.2

## Details

* The exact line of the code is here: https://github.com/encode/httpcore/blob/0.12.2/httpcore/_backends/asyncio.py#L181
```python
    def is_readable(self) -> bool:
        transport = self.stream_reader._transport  # type: ignore
        sock: socket.socket = transport.get_extra_info("socket")
        return is_socket_readable(sock.fileno())
```
* If `transport.get_extra_info("socket")` returns `None`, `sock.fileno()` crashes.
* First they type of `sock` should probably be `Optional[socket.socket]`.
* The python docs are not very detailed about the guarantees of `extra["socket"]` existence but in the example code they do test it: https://docs.python.org/3/library/asyncio-protocol.html#asyncio.BaseTransport.get_extra_info
* So as other code e.g. in [uvicorn](https://github.com/encode/uvicorn/blob/0b941abfbb527d102d11e950c2fea1f9ecf1988f/uvicorn/protocols/utils.py#L6) or [aiohttp](https://github.com/aio-libs/aiohttp/blob/6ab76b084bf5012b7185046162ed92bedcf073b5/aiohttp/tcp_helpers.py#L28)
* I wish I could provide a working example, but this only occasionally fails in my system. 
* However I do have a stacktrace capture:
```
AttributeError 'NoneType' object has no attribute 'fileno'.

  File "httpx/_client.py", line 1371, in request
    response = await self.send(
  File "httpx/_client.py", line 1406, in send
    response = await self._send_handling_auth(
  File "httpx/_client.py", line 1444, in _send_handling_auth
    response = await self._send_handling_redirects(
  File "httpx/_client.py", line 1476, in _send_handling_redirects
    response = await self._send_single_request(request, timeout)
  File "httpx/_client.py", line 1502, in _send_single_request
    (status_code, headers, stream, ext,) = await transport.arequest(
  File "httpcore/_async/http_proxy.py", line 124, in arequest
    return await self._tunnel_request(
  File "httpcore/_async/http_proxy.py", line 190, in _tunnel_request
[...]

PR #242, opened 2020-11-26
Fix SocketStream.is_readable()

povilasb:
transport.get_extra_info('socket') may return None when connection is dropped.
This commit makes sure we expect such case.
Also, that means socket is readable so that the stream user would be notified/awaken and act accordingly.

Fixes https://github.com/encode/httpcore/issues/239

### DIFF
--- a/httpcore/_backends/asyncio.py
+++ b/httpcore/_backends/asyncio.py
@@ -3,6 +3,7 @@ import socket
 from ssl import SSLContext
 from typing import Optional
 
+from .. import _utils
 from .._exceptions import (
     CloseError,
     ConnectError,
@@ -14,7 +15,6 @@ from .._exceptions import (
     map_exceptions,
 )
 from .._types import TimeoutDict
-from .._utils import is_socket_readable
 from .base import AsyncBackend, AsyncLock, AsyncSemaphore, AsyncSocketStream
 
 SSL_MONKEY_PATCH_APPLIED = False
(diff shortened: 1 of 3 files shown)


povilasb:
Rebased off latest `master`.

tests/backend_tests/test_asyncio.py
```diff
@@ -0,0 +1,25 @@
...
+            stream_reader._transport.get_extra_info.return_value = None
+            sock_stream = SocketStream(stream_reader, MagicMock())
+
+            assert sock_stream.is_readable()
+
+        @patch("httpcore._utils.is_socket_readable", MagicMock(return_value=True))
+        @pytest.mark.asyncio
+        async def test_returns_true_when_socket_is_readable(self):
+            stream_reader = MagicMock()
+            stream_reader._transport.get_extra_info.return_value = None
```
florimondmanca, line 22: Hmm. Should this be returning something truthy? Right now the two tests do the same thing, and the patched is_socket_readable helper isn't actually called.

florimondmanca (commented):
Looking good! We haven't been using mock-based unit tests very much yet. Usually we use a "fakes" approach, where we pass actual objects implementing the interfaces we're testing against. But in this case I agree monkey patching seems like the simplest approach.

httpcore/_backends/asyncio.py
```diff
@@ -195,8 +195,10 @@ async def aclose(self) -> None:
 
     def is_readable(self) -> bool:
         transport = self.stream_reader._transport  # type: ignore
-        sock: socket.socket = transport.get_extra_info("socket")
-        return is_socket_readable(sock.fileno())
+        sock: Optional[socket.socket] = transport.get_extra_info("socket")
+        # If socket was detatched from the transport, most likely connection was reset.
+        # Hence make it readable to notify users to poll the socket.
+        return not sock or is_socket_readable(sock.fileno())
```
florimondmanca, line 201: (Nit) Using explicit `None` checks:

```suggestion
        return sock is None or is_socket_readable(sock.fileno())
```

florimondmanca (approved):
Left a few nits, but I'm happy that this successfully fixes the issue, so approving in advance. Let me know if my comments make sense, and we can merge when ready!

tests/backend_tests/test_asyncio.py
povilasb: you're right, let me fix this :)

povilasb:
Addressed your comments and amended appropriate commits ;)

florimondmanca (approved):
Thanks!

PR #243, opened 2020-11-27
Follow up tweaks to Host header precedence for :authority in HTTP/2

tomchristie:
Follow up to #241

* Drop pseudo-headers being stripped. Let's treat any behaviour there separately.
* If a Host header is included, then don't auto-append port info onto it.
* Minor tweak of comment.


florimondmanca (approved):
Makes sense!

PR #244, opened 2020-11-30
Tweak `is_readable` comment.

tomchristie:
What do we think about this to help users understand what an `is_readable` means for a disconnected socket?


florimondmanca (approved):
Yep!

PR #249, opened 2020-12-07
Fix asyncio SSL warning when using proxy tunneling

florimondmanca:
Closes https://github.com/encode/httpx/issues/859

Turns out we need to call `protocol.connection_made()`. It calls `stream_reader.set_transport(...)`, as well as a bunch of other stuff, _including_ marking the reader as being "over SSL", which fixes the warning saw in the issue.

See https://github.com/encode/httpx/issues/859#issuecomment-740004212 for the detailed analysis.

### DIFF
--- a/httpcore/_backends/asyncio.py
+++ b/httpcore/_backends/asyncio.py
@@ -120,7 +120,7 @@ class SocketStream(AsyncSocketStream):
                 timeout=timeout.get("connect"),
             )
 
-        # Initialize the protocol, so the reader is made aware of being tied to
+        # Initialize the protocol, so it is made aware of being tied to
         # a TLS connection.
         # See: https://github.com/encode/httpx/issues/859
         protocol.connection_made(transport)
--- a/httpcore/_backends/asyncio.py
+++ b/httpcore/_backends/asyncio.py
@@ -120,7 +120,11 @@ class SocketStream(AsyncSocketStream):
                 timeout=timeout.get("connect"),
             )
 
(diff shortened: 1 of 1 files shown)


httpcore/_backends/asyncio.py
```diff
@@ -120,7 +120,11 @@ async def start_tls(
                 timeout=timeout.get("connect"),
             )
 
-        stream_reader.set_transport(transport)
+        # Initialize the protocol, so the reader is made aware of being tied to
+        # a TLS connection.
```
JayH5, line 124: I always find the asyncio protocol/transport stuff _very_ confusing, but this comment seems the wrong way around? This change is to make the _protocol_ aware of the TLS connection.

JayH5 (approved):
Good work figuring this out 👍

httpcore/_backends/asyncio.py
florimondmanca: Ah, well yeah you're right. I'm confused too 😆

httpcore/_backends/asyncio.py
```diff
@@ -120,7 +120,11 @@ async def start_tls(
                 timeout=timeout.get("connect"),
             )
 
-        stream_reader.set_transport(transport)
+        # Initialize the protocol, so the reader is made aware of being tied to
```
florimondmanca, line 123:
```suggestion
        # Initialize the protocol, so it is made aware of being tied to
```

PR #252, opened 2020-12-10
Skip CONNECT when making HTTPS requests through an HTTPS proxy

florimondmanca:
Closes https://github.com/encode/httpx/issues/1424

_Not sure_ about this, so happy to take any reviews / proof-thinking.

When an HTTP proxy is used, if a request to an HTTPS server is made then we issue a CONNECT to the proxy to establish a tunnel, then upgrade to TLS. This is standard, and fine.

However if using an HTTPS proxy, then the TCP connection returned by CONNECT will be a TLS connection between the client and the proxy. Right now we still attempt to do `.start_tls()` on top of it, which fails, obviously.

From investigation, it seems that instead of "re-upgrading" to TLS, `curl` drops the proxy connection at this point and proceeds to requesting the HTTPS server directly, while Requests doesn't even bother and does the HTTPS request directly, without a CONNECT request. (Both assumptions need confirming through source code, but that's the behavior I'm guessing from investigation and observations in https://github.com/encode/httpx/issues/1424.)

This PR switches HTTPCore to doing to same thing as Requests. The net result is that everything is as before, except any HTTPS proxy is ignored when making HTTPS requests. (If using an HTTP proxy then we still do `.start_tls()`.)

TODO: tests:

* [ ] Update `pproxy` test setup so it listens on an HTTPS port as well.
* [ ] Add `test_https_proxy_http_request()` and `test_https_proxy_https_request()`.

### DIFF
--- a/httpcore/_async/http_proxy.py
+++ b/httpcore/_async/http_proxy.py
@@ -188,8 +188,9 @@ class AsyncHTTPProxy(AsyncConnectionPool):
         timeout = cast(TimeoutDict, ext.get("timeout", {}))
         origin = url_to_origin(url)
         connection = await self._get_connection_from_pool(origin)
+        proxy_scheme, _, _ = self.proxy_origin
 
-        if connection is None:
+        if connection is None and proxy_scheme == b"http":
             scheme, host, port = origin
 
             # First, create a connection to the proxy server
@@ -250,6 +251,16 @@ class AsyncHTTPProxy(AsyncConnectionPool):
                 http2=self._http2,
                 ssl_context=self._ssl_context,
                 socket=proxy_connection.socket,
+            )
+            await self._add_to_pool(connection, timeout)
+        elif connection is None:
+            # HTTPS requests issued through an HTTPS proxy don't need proxying
+            # at all. Send them through directly.
+            assert proxy_scheme == b"https"
+            connection = AsyncHTTPConnection(
+                origin=origin,
+                http2=self._http2,
+                ssl_context=self._ssl_context,
             )
             await self._add_to_pool(connection, timeout)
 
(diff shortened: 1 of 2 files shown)


florimondmanca:
@elliottmoos Hey — Could you perhaps try installing `httpcore` from this branch, and looking at if it resolves the problems you reported in https://github.com/encode/httpx/issues/1424?

```python
pip install git+https://github.com/encode/httpcore.git@fm/https-proxy-bypass#egg=httpcore
```

sethmlarson:
(Leaving a note here that I wouldn't mind reviewing this if you're okay not merging asap, just low on time atm)

florimondmanca:
@sethmlarson Hey! — Absolutely, would be super nice to get some TLS expert eyes on this. I basically opened this so we have something to look at for a start, but it could just as well been a draft PR. :)

ElliottMoos:
> ```shell
> pip install 'git+https://github.com/encode/httpcore.git@fm/https-proxy-bypass#egg=httpcore'
> ```

Just tried that branch, works for me. Thanks for letting me know about this!

ElliottMoos:
So @florimondmanca after looking at this a little more, it seems like it'll break my (and I imagine others) use case. I don't have any other route to the internet other than this proxy, and bypassing it will not work. 

I tested curl version 7.64.1 on macOS and it does not bypass the proxy (the ip I get back from ifconfig is the one I want). 

It looks like the functionality that requests and httpcore implements when it comes to https proxy urls is different than curl's. At least for the version I used.

florimondmanca:
@ElliottMoos Yeah, your comment about a proxy often being the only meeting point before the wider Internet makes sense.

From the cURL announcement here https://daniel.haxx.se/blog/2016/11/26/https-proxy-with-curl/ as well as this issue it does seem like cURL tunnels HTTPS requests as well when using an HTTPS proxy.

My main question there then is "how". Is it even possible to negotiate TLS with a server using a TCP tunnel that's managed by a proxy and encrypted on the client-proxy portion? Half-nested TLS? 🤯

florimondmanca:
I reckon this is probably not the proper way to go about this…? Will close it for now.

PR #253, opened 2020-12-12
Ensure h2 connection gets closed in case of exception

florimondmanca:
May close encode/httpx#1414

Tweaks our HTTP/2 connection implementation so that the underlying `h2_state` is properly closed before bubbling up any exceptions.

Right now it seems possible for the underlying state to leak eg if the connection gets lost unexpectedly — see: https://github.com/encode/httpx/issues/1414#issuecomment-742127990


frankwalter1301:
Doesn't work with the new version of httpcore.

tomchristie:
@frankwalter1301 Could you give some more background about why you're bumping into this?

frankwalter1301:
I simple encountered the same error as the issue linked to this talks
about. Same circumstances: too many failed connections in the same http
client.

Il mar 16 nov 2021, 12:15 Tom Christie ***@***.***> ha
scritto:

tomchristie:
Believe that #440 is what we need to close encode/httpx#1414, but *yes!* this is a good and sensible bit of defensive programming, and we should have it in.

PR #270, opened 2021-02-07
Rework connection timeout tests

JayH5:
These tests are currently broken on master. They also don't really seem to be tests for connection timeouts, rather for the behaviour when trying to connect using SSL to a server that wasn't using SSL. This behaviour used to be a connection timeout, but now (at least on Python 3.7+) is an SSL error. I'm not sure what changed to cause this to happen (possibly different OpenSSL version?) but it would be good to know.

Change the `test_connection_timeout_tcp` test to attempt to connect to a server port that is bound but not listening/accepting connections.

Remove the `test_connection_timeout_uds` test because as far as I can tell there isn't really a way to cause a timeout here (connect returns immediately with either success or failure on any bound Unix socket).

Not super sure about the naming of things here...

### DIFF
--- a/tests/async_tests/test_interfaces.py
+++ b/tests/async_tests/test_interfaces.py
@@ -1,5 +1,4 @@
 import platform
-import sys
 
 import pytest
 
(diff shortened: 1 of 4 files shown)


JayH5:
😞 having trouble reproducing test failures locally.. may be a macOS vs Linux thing.

JayH5:
Yep, this approach works on macOS but not Linux.

PR #271, opened 2021-02-12
Improve interoperability when closing down asyncio TLS connections

florimondmanca:
Fixes https://github.com/encode/httpx/issues/1459, follow-up to #167.

Catch any `SSLError` that may occur when waiting for servers to acknowledge that we're shutting down a TLS connection. As we saw in encode/httpx#1459, this may happen due to protocol violations on the server side, which I think are okay enough to ignore since they don't prevent the connection from being shut down.

cc @lazka

### DIFF
--- a/httpcore/_backends/asyncio.py
+++ b/httpcore/_backends/asyncio.py
@@ -1,6 +1,6 @@
 import asyncio
 import socket
-from ssl import SSLContext
+from ssl import SSLContext, SSLError
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


lazka:
I can confirm that this fixes things in my case. As for if it is the right thing to do, I don't know enough about this :)

tomchristie:
Superseeded by https://github.com/encode/httpcore/pull/310

Issue #273, opened 2021-02-22
Expose socket connection interface in connection


jborean93:
I've working on an custom httpx transport which uses Kerberos/NTLM/Negotiate authentication when talking to a peer. These authentication methods have a custom extension when talking over TLS called channel binding tokens (CBT) where you embed a structure that is based on information from the TLS context which in essence binds the authentication against the TLS context making it harder to achieve MitM attacks.

A version of CBT that I'm currently using is documented in [RFC 5929](https://tools.ietf.org/html/rfc5929#section-4) which is essentially `tls-server-end-point:<cert hash>`. In the past I've just send a request to the host, gotten the SSL object from the socket in the response but this has a few problems:

* You waste a round trip sending a request you know will fail
  * Somewhat more annoying because in TLS there are more than 1 roundtrips to set up the TLS context
* You need to have set up the connection before you can build the authentication context and generate the first auth token
* If the connection is closed in the auth request then the CBT data you used might not be correct anymore

The last point is problematic because in my scenario I send a blank request which sets up a connection and the response I get back is a 401. Typically I've seen a `Connection: close` header which has the `*ConnectionPool` classes drop the connection from the pool. This means the next request is a brand new connection with a brand new negotiated TLS context.

For the `tls-server-end-point` I could still get the certificate hash as that should be static between the connections but with new protocols like TLS 1.3 the [draft proposal for CBT](https://tools.ietf.org/html/draft-whited-tls-channel-bindings-for-tls13-00#section-2) seems to indicate it's based on the unique connection set up Breaking authentication.

What I'm hoping is for is to expose the [_open_socket](https://github.com/encode/httpcore/blob/ed76a30277529c5687cbdb6e203d6dbd0d56833f/httpcore/_async/connection.py#L108-L139) method publically in some shape or form. This would allow me to create the socket connection and set up the TLS request before I send a request through `(s)request` on the connection. I feel like this could potentially be done by setting up the socket myself and passing it into the `__init__()` for `*HTTPConnection` or by just making that method more public by removing the `_`. This also hinges on my other question around whether connections are actually meant to be publicly exposed in https://github.com/encode/httpcore/issues/272.

Very interested to hear your thoughts on this as I do have a working prototype for all this but it relies on being able to access `AsyncHTTPConnection` and calling `_open_socket` manually before sending the request through `(a)request`.

jborean93:
[...]

PR #274, opened 2021-02-23
Added open method to open just the connection

jborean93:
This adds a `(a)open()` method to the transport base class that defaults to a no-op operation. It is used by the HTTP Connection classes to solely open the socket connection and create the underlying HTTP connection object without sending an actual HTTP request.

Just wanted to get your thoughts on this before I start spending some time on adding some tests to see if this approach is ok. Right now the default open method is a no-op and the classes that do have an actual implementation aren't publicly exposed so it shouldn't affect users of this library. Hopefully it's a first step to enable some of my use cases without re implementing the entire class myself.

Fixes https://github.com/encode/httpcore/issues/273

### DIFF
--- a/httpcore/_async/base.py
+++ b/httpcore/_async/base.py
@@ -100,7 +100,13 @@ class AsyncHTTPTransport:
         and any keep alive connections.
         """
 
+    async def aopen(self) -> None:
+        """
+        Opens the implementation connection to the target.
+        """
+
     async def __aenter__(self: T) -> T:
+        await self.aopen()
         return self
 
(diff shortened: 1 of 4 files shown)


florimondmanca:
@jborean93 :wave:

What might help in assessing whether this is going in a sensible direction is an example usage code snippet. How would we expect users to benefit from this extra hook, and what would code for that look like? It often helps to start from the desired public API and then figure out the inner details.

tomchristie:
Closing this off since we need more prior discussion before moving forward with API changes like this.

PR #275, opened 2021-02-23
Open sockets outside of connection objects

florimondmanca:
See https://github.com/encode/httpcore/issues/273#issuecomment-784225770

Experimenting with a different style for socket management in HTTP connections:

Instead of opening sockets as part of the first `request()` call, open the socket beforehand as part of the connection pool's `request()`, and pass it down to the new connection object.

### DIFF
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -1,53 +1,50 @@
 from ssl import SSLContext
-from typing import Optional, Tuple, cast
+from typing import Optional, Tuple
 
-from .._backends.auto import AsyncBackend, AsyncLock, AsyncSocketStream, AutoBackend
-from .._exceptions import ConnectError, ConnectTimeout
+from .._backends.auto import AsyncBackend, AsyncSocketStream, AutoBackend
 from .._types import URL, Headers, Origin, TimeoutDict
-from .._utils import exponential_backoff, get_logger, url_to_origin
-from .base import (
-    AsyncByteStream,
-    AsyncHTTPTransport,
-    ConnectionState,
-    NewConnectionRequired,
-)
+from .._utils import get_logger, url_to_origin
+from .base import AsyncByteStream, AsyncHTTPTransport, ConnectionState
 from .http import AsyncBaseHTTPConnection
 from .http11 import AsyncHTTP11Connection
 
 logger = get_logger(__name__)
-
-RETRIES_BACKOFF_FACTOR = 0.5  # 0s, 0.5s, 1s, 2s, 4s, etc.
 
 
 class AsyncHTTPConnection(AsyncHTTPTransport):
     def __init__(
         self,
         origin: Origin,
-        http2: bool = False,
-        uds: str = None,
+        socket: AsyncSocketStream,
         ssl_context: SSLContext = None,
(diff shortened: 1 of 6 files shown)


httpcore/_async/connection.py
```diff
@@ -81,103 +68,25 @@ async def arequest(
...
-                        "open_socket origin=%r timeout=%r", self.origin, timeout
-                    )
-                    self.socket = await self._open_socket(timeout)
-                self._create_connection(self.socket)
-            elif self.state in (ConnectionState.READY, ConnectionState.IDLE):
-                pass
-            elif self.state == ConnectionState.ACTIVE and self.is_http2:
-                pass
-            else:
-                raise NewConnectionRequired()
```
florimondmanca, line 100: I'm not sure about why this `NewConnectionRequired()` case was introduced, so it's possible the new impl is missing something in the HTTP/1.1 case.

httpcore/_async/connection_pool.py
```diff
@@ -167,17 +182,50 @@ def _connection_acquiry_lock(self) -> AsyncLock:
...
-            http2=self._http2,
-            uds=self._uds,
+            socket=socket,
             ssl_context=self._ssl_context,
-            local_address=self._local_address,
-            retries=self._retries,
             backend=self._backend,
         )
 
+    async def _open_socket(
```
florimondmanca, line 194: Custom pool implementations would be able to override this method (along with a strict package pin) if they wanted to customize how sockets are opened.

Standalone implementations (single connection w/o pooling) could copy this code and use them in their own `open_custom_connection()` context manager…

httpcore/_async/connection.py
```diff
@@ -1,54 +1,51 @@
...
-RETRIES_BACKOFF_FACTOR = 0.5  # 0s, 0.5s, 1s, 2s, 4s, etc.
-
 
 class AsyncHTTPConnection(AsyncHTTPTransport):
     def __init__(
         self,
         origin: Origin,
-        http2: bool = False,
-        uds: str = None,
+        socket: AsyncSocketStream,
```
florimondmanca, line 18: This is the change from which all others derive: `socket` becomes required, and will be passed by whoever wants to interact with an HTTP connection (in our case, the connection pool).

httpcore/_async/connection.py
jborean93: I was also unsure what this was for, a quick glance at the repo and I couldn't find any location where it was caught to be handled.

httpcore/_async/connection_pool.py
jborean93: While you can always have your own implementation you also need to implement the underlying connection code which is not impossible but it would be nice to utilise the existing connection codes in httpcore. By reimplementing all the code you

* Duplicate work, it's IMO a non-trivial amount of code to get a basic HTTP connection going
* Missing out on new functionality/bugfixes that is added by httpcore who know what they are doing when it comes to issues
* Makes the implementers code easier to maintain, they don't need to worry about h11, h2 and how to interact with that

I totally understand this project is in it's infancy so you may not be prepared to expose the existing classes, I just wanted to point out something I wish to see in the future :)

tomchristie:
Closing as out of date.

PR #277, opened 2021-02-27
Ensured AnyIO 3.x compatibility

agronholm:


tomchristie:
For clarity, what's the necessary change wrt. AnyIO 3.0?
What *is* `maybe_async_cm`?
Might(?) be worth a comment in the test code.

agronholm:
Several functions and methods were changed from async to sync and the dummy awaitables they return will raise deprecation warnings. This PR avoids those. I wasn't going to mark this ready for testing until at least a release candidate of v3.0 has been published.

The `maybe_async_cm()` function wraps a potentially asynchronous context manager and lets one use it as an asynchronous context manager. See the docs: https://anyio.readthedocs.io/en/stable/api.html#anyio.maybe_async_cm

tomchristie:
Okay. Just to help me out a bit more here, let's take one particular example...

Have I got things right in thinking that this case: `anyio.fail_after(connect_timeout)` was `async` in 2.x and is now sync in 3.x?

(If so, I'm a bit confused about how that particular case can be a sync function, given that it's potentially loop-blocking?)

agronholm:
It doesn't really block the loop. It was async in 2.x only because of curio which does not let anything interface with the kernel without async/await. Now with curio dropped as a backend, the goal is to bring the API closer to that of trio, where the same things are synchronous.

tomchristie:
Okay, right, makes more sense. Perhaps a single-line comment on the `import` switch to note that some async API becomes sync API from 2.x -> 3.x. Then I think we're good on this. Thanks!

(Also, currently blocked on #282)

florimondmanca:
Interesting… https://github.com/agronholm/anyio/issues/185

florimondmanca:
@agronholm Are there other changes you plan to add to this PR? It looks approvable and mergeable to me now, just wondering given it's currently in "draft" state.

agronholm:
> @agronholm Are there other changes you plan to add to this PR? It looks approvable and mergeable to me now, just wondering given it's currently in "draft" state.

It was waiting for the first 3.0 prerelease. I didn't want to submit this PR before I knew the full extent of the API changes I would make. Now 3.0.0rc1 is out so we can move ahead with this.

agronholm:
It says "check failed" but does not provide any reason why.

tomchristie:
True, lemme hit that "re-run jobs" button.

agronholm:
Now that AnyIO 3.0 final is out, I think it's time to get this moving. I'll make one last pass to update this against the final API and then mark it as ready for review.

agronholm:
Ok, it's all good now.

tomchristie:
Great, let's get it in then.

Welcome to issue a pull request bumping the version number and changelog if you'd like to roll a release from this.

Any thoughts on how long you'd like to maintain 2.0 + 3.0 support for?

agronholm:
I'm not sure really. At least when 4.0 is released will I drop 2.x support. That's ways off at the moment though (no plans for that yet).

I think it would be prudent to add an extra for `anyio` which would install a compatible version. I'll make a PR of that.

Issue #276, opened 2021-02-23
Broken `master` build: SSL_WRONG_VERSION_NUMBER


florimondmanca:
We have a CI build error that I saw in the latest `master` build as well as some other branches, eg #275.

* Not able to reproduce on my local macOS machine. CI uses Linux -> Can someone on Linux reproduce by running the test suite?
* It happens in `test_connection_timeout_tcp`, which does something hacky: try to connect via HTTPS to generate a `ConnectTimeout`. In CI we get a `ConnectError` instead. Probably related to the underlying `openssl` version?

https://github.com/encode/httpcore/runs/1785395873?check_suite_focus=true#step:7:79

```
__________________ test_connection_timeout_tcp[asyncio-auto] ___________________

backend = 'auto'
server = <tests.utils.HypercornServer object at 0x7f7f41716810>

    @pytest.mark.skipif(
        sys.version_info[:2] < (3, 7),
        reason="Hypercorn doesn't support python < 3.7 (this test is local-only)",
    )
    @pytest.mark.anyio
    async def test_connection_timeout_tcp(backend: str, server: Server) -> None:
        # we try to access http server using https. It caused some SSL timeouts
        # in TLSStream.wrap inside inside AnyIOBackend.open_tcp_stream
        method = b"GET"
        url = (b"https", *server.netloc, b"/")
        headers = [server.host_header]
        ext = {"timeout": {"connect": 0.1}}
    
        async with httpcore.AsyncConnectionPool(backend=backend) as http:
            with pytest.raises(httpcore.ConnectTimeout):
>               await http.arequest(method, url, headers, ext=ext)

tests/async_tests/test_interfaces.py:487: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
httpcore/_async/connection_pool.py:259: in arequest
    socket = await self._open_socket(origin, timeout=timeout)
httpcore/_async/connection_pool.py:214: in _open_socket
    local_address=self._local_address,
httpcore/_backends/auto.py:45: in open_tcp_stream
    hostname, port, ssl_context, timeout, local_address=local_address
httpcore/_backends/asyncio.py:268: in open_tcp_stream
    connect_timeout,
/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/contextlib.py:130: in __exit__
    self.gen.throw(type, value, traceback)
[...]

PR #281, opened 2021-03-12
Drop pyproject.toml

tomchristie:
Personally I'm not keen on us having a `pyproject.toml` exclusively for some bits of black configuration. We don't have this on our other projects, and I don't particularly appreciate IDE design choices forcing us into particular tooling conventions.

I'm okay with us having a conversation around this, but also am very strongly in favour of us being able to stick with similar conventions on our projects across `encode`, and striving towards low-config, low-tooling approaches wherever possible. To me this also include narrowing down the number of magical configuration files we have at the top-level.

Also, I needed a PR excuse to have a look at [the latest build failure nonsense](https://github.com/encode/httpcore/runs/1785395873).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,8 +1,0 @@
-[tool.black]
-exclude = '''
-/(
-    httpcore/_sync
-  | tests/sync_tests
-)/
-'''
-target-version = ['py36']
--- a/scripts/check
+++ b/scripts/check
@@ -9,7 +9,7 @@ export SOURCE_FILES="httpcore tests"
 set -x
 
 ${PREFIX}isort --check --diff --project=httpcore $SOURCE_FILES
-${PREFIX}black --check --diff --target-version=py36 --exclude 'httpcore/_sync/.*' --exclude 'tests/sync_tests/.*' $SOURCE_FILES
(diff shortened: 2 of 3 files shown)


tomchristie:
I don't understand black's behaviour here.

Eg...

```shell
+ venv/bin/black --force-exclude 'httpcore/_sync/.*' --force-exclude 'tests/sync_tests/.*' --check --diff --target-version=py36 httpcore tests
would reformat httpcore/_sync/http11.py
```

But, I just asked you *not* to include that file?

tomchristie:
Okay got it, `--exclude` (and `--force-exclude`) only accept a single argument, so the flag is being silently overwritten.

tomchristie:
Okay, so presumably the SSL errors are something to do with the very latest python releases having some newer SSL config.

florimondmanca:
> Okay, so presumably the SSL errors are something to do with the very latest python releases having some newer SSL config.

Yup, tracked this under https://github.com/encode/httpcore/issues/276 — we've got a test that needs updating to unblock CI.

florimondmanca:
Referencing https://github.com/encode/httpcore/pull/156/ and https://github.com/encode/httpcore/pull/156/#discussion_r469156830 which is where we moved to this `pyproject.toml` file. We used to not have it before, right. :-)

florimondmanca (approved):

PR #282, opened 2021-03-12
Drop flaky connection tests

tomchristie:
Giving this a quick tyre-kick based on #276, to double check it it is these test cases that are causing the current failures on master.

If so, then my assessment would be that flaky tests are worse than no tests, and that we should drop any flaky cases, but raise an issue to re-introduce them. That'd get us moving again, while still keeping an open issue for addressing this issue more comprehensively.


florimondmanca (approved):
Yes, this confirms this test was the culprit.

PR #283, opened 2021-03-15
[WIP] Add HTTP/3 support

jlaine:
This PR is still a very rough draft, it:

- lacks any kind of unittest
- ~~only implements asyncio~~
- doesn't yet do packet retransmission
- doesn't handle timeouts
- doesn't respect the SSL configuration

Still I am able to use the example from the README to perform an HTTP/3 requests against `quic.aiortc.org` port 443.

### DIFF
--- a/httpcore/_async/http3.py
+++ b/httpcore/_async/http3.py
@@ -1,0 +1,354 @@
+from ssl import SSLContext
+from typing import AsyncIterator, Dict, List, Tuple, cast
+
+from aioquic.h3.connection import H3_ALPN, H3Connection
+from aioquic.h3.events import DataReceived, H3Event, HeadersReceived
+from aioquic.quic.configuration import QuicConfiguration
+from aioquic.quic.connection import QuicConnection
+
+from .._backends.auto import (
+    AsyncBackend,
+    AsyncDatagramSocket,
+    AsyncLock,
+    AsyncSocketStream,
+)
+from .._bytestreams import AsyncIteratorByteStream, PlainByteStream
+from .._exceptions import RemoteProtocolError
+from .._types import URL, Headers, TimeoutDict
+from .._utils import get_logger
+from .base import AsyncByteStream, ConnectionState
+from .http import AsyncBaseHTTPConnection
+
+logger = get_logger(__name__)
+
+
+class AsyncHTTP3Connection(AsyncBaseHTTPConnection):
+    def __init__(
+        self,
+        socket: AsyncDatagramSocket,
+        backend: AsyncBackend,
+        ssl_context: SSLContext = None,
+    ):
+        self.socket = socket
+        self.ssl_context = SSLContext() if ssl_context is None else ssl_context
+
+        self.backend = backend
+        self._quic = QuicConnection(
+            configuration=QuicConfiguration(is_client=True, alpn_protocols=H3_ALPN)
+        )
+        self._http = H3Connection(self._quic)
+
+        self.sent_connection_init = False
+        self.streams = {}  # type: Dict[int, AsyncHTTP3Stream]
+        self.events = {}  # type: Dict[int, List[H3Event]]
+
+        self.state = ConnectionState.ACTIVE
+
+    def __repr__(self) -> str:
+        return f"<AsyncHTTP3Connection state={self.state}>"
+
+    def info(self) -> str:
+        return f"HTTP/3, {self.state.name}, {len(self.streams)} streams"
+
+    @property
+    def init_lock(self) -> AsyncLock:
+        # We do this lazily, to make sure backend autodetection always
+        # runs within an async context.
+        if not hasattr(self, "_initialization_lock"):
+            self._initialization_lock = self.backend.create_lock()
+        return self._initialization_lock
+
+    @property
+    def read_lock(self) -> AsyncLock:
+        # We do this lazily, to make sure backend autodetection always
+        # runs within an async context.
+        if not hasattr(self, "_read_lock"):
+            self._read_lock = self.backend.create_lock()
+        return self._read_lock
+
+    async def start_tls(
+        self, hostname: bytes, timeout: TimeoutDict = None
+    ) -> AsyncSocketStream:
+        raise NotImplementedError("TLS upgrade not supported on HTTP/3 connections.")
+
+    def get_state(self) -> ConnectionState:
+        return self.state
+
+    def mark_as_ready(self) -> None:
+        if self.state == ConnectionState.IDLE:
+            self.state = ConnectionState.READY
+
+    async def arequest(
+        self,
+        method: bytes,
+        url: URL,
+        headers: Headers = None,
+        stream: AsyncByteStream = None,
+        ext: dict = None,
+    ) -> Tuple[int, Headers, AsyncByteStream, dict]:
+        ext = {} if ext is None else ext
+        timeout = cast(TimeoutDict, ext.get("timeout", {}))
+
+        async with self.init_lock:
+            if not self.sent_connection_init:
+                # The very first stream is responsible for initiating the connection.
+                self.state = ConnectionState.ACTIVE
+                await self.send_connection_init(timeout)
+                self.sent_connection_init = True
+
+        # await self.max_streams_semaphore.acquire()
+        try:
+            stream_id = self._quic.get_next_available_stream_id()
+
+            h3_stream = AsyncHTTP3Stream(stream_id=stream_id, connection=self)
+            self.streams[stream_id] = h3_stream
+            self.events[stream_id] = []
+            return await h3_stream.arequest(method, url, headers, stream, ext)
+        except Exception:  # noqa: PIE786
+            # await self.max_streams_semaphore.release()
+            raise
+
+    async def send_connection_init(self, timeout: TimeoutDict) -> None:
+        """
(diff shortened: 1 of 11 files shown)


httpcore/_async/http3.py
```diff
@@ -0,0 +1,354 @@
...
+        async with self.read_lock:
+            while not self.events[stream_id]:
+                await self.receive_events(timeout)
+        return self.events[stream_id].pop(0)
+
+    async def receive_events(self, timeout: TimeoutDict) -> None:
+        """
+        Read some data from the network, and update the H2 state.
+        """
+        data, addr = await self.socket.receive()
```
florimondmanca, line 152: Do we need to apply the `read` timeout here?

httpcore/_async/http3.py
```diff
@@ -0,0 +1,354 @@
...
+    async def _transmit(self) -> None:
+        """
+        Send pending datagrams to the peer and arm the timer if needed.
+        """
+        self._transmit_task = None
+
+        # send datagrams
+        now = await self.backend.time()
+        for data, addr in self._quic.datagrams_to_send(now=now):
+            await self.socket.send(data, addr)
```
florimondmanca, line 217: Do we need to apply the `write` timeout here?

httpcore/_backends/asyncio.py
```diff
@@ -78,6 +85,50 @@ async def backport_start_tls(
...
+            await self._protocol.read_event.wait()
+
+        data, addr = self._protocol.read_queue.popleft()
+
+        if not self._protocol.read_queue:
+            self._protocol.read_event.clear()
+
+        return data, addr
+
+    async def send(self, data: bytes, addr: SocketAddr) -> None:
```
florimondmanca, line 128: Is there something we should be "waiting" for? In other backends this calls an async operation, IDK if asyncio exposes any kind of equivalent "actually wait for something to happen" API, instead of just scheduling something for later?

httpcore/_backends/base.py
```diff
@@ -39,6 +39,28 @@ def lookup_sync_backend(name: str) -> "SyncBackend":
...
+
+    remote_addr: SocketAddr
+
+    def get_http_version(self) -> str:
+        return "HTTP/3"
+
+    async def aclose(self) -> None:
+        raise NotImplementedError()  # pragma: no cover
+
+    async def receive(self) -> Tuple[bytes, SocketAddr]:
```
florimondmanca, line 57: For consistency, should these be named read() and write()? Otherwise we might consider to update the doc string for this class to use the same receive/send lingo.

httpcore/_types.py
```diff
@@ -9,4 +9,5 @@
 Origin = Tuple[bytes, bytes, int]
 URL = Tuple[bytes, bytes, Optional[int], bytes]
 Headers = List[Tuple[bytes, bytes]]
+SocketAddr = Tuple
```
florimondmanca, line 12: Anything more specific?

httpcore/_backends/asyncio.py
```diff
@@ -246,6 +297,29 @@ def __init__(self) -> None:
...
+        hostname: bytes,
+        port: int,
+        timeout: TimeoutDict,
+        *,
+        local_address: Optional[str],
+    ) -> AsyncDatagramSocket:
+        loop = asyncio.get_event_loop()
+
+        infos = await loop.getaddrinfo(hostname, port, type=socket.SOCK_DGRAM)
+        _, local_addr, remote_addr = _utils.udp_socket_parameters(infos[0][4])
```
florimondmanca, line 311: Might be worth an intermediate variable for this obscure `infos[0][4]` thingy?

florimondmanca (commented):
Pretty exciting. ❤️

Leaving some comments on the code for you.

florimondmanca:
As @tomchristie mentioned in the issue, I'm not sure if we should be adding this as experimental to HTTPCore itself right away, or host the code separately?

Perhaps we could start with a gist (although somewhat lengthy) that completely ignores any connection pooling, and just provides a fully functional standalone `H3Connection` transport for people to play with? Could even be hosted on a proper repo in the interim. Then we get some feedback and clean up some rough edges, and work from there? I'm sure there are ways to either copy paste some stable code or strip out things we don't actually need for a barebones standalone connection implementation, so the end result is realistic enough yet feasible as a separate experimental project at first.

I'd also be curious about some docs draft, see how we'd frame this for users?

httpcore/_types.py
jlaine: It varies.. you can receive Tuple[str, int] or Tuple[str, int, int, int]

httpcore/_backends/base.py
jlaine: I don't think read / write make sense here as we are also conveying another information: the remote address. The "classical" corresponding APIs are `recvfrom` and `sendto` but as I understand we're avoid abbreviations, I went for "receive" and "send". 

`anyio` seems to have settled on the same names: https://github.com/agronholm/anyio/blob/3a0b805a3f4410009f2b6610d329fb9cc2b90e5a/src/anyio/_backends/_asyncio.py#L1258

httpcore/_backends/asyncio.py
jlaine: To my knowledge UDP APIs don't provide any kind of backpressure.

jlaine:
I'm not sure "hosting it externally" makes much sense, there has been an `httpx` demo in aioquic's example for over a year and a half, during which time it bit-rotted due to API refactoring in httpx. The same is likely to happen again with httpcore. I'm also not very keen on duplicating httpcore's mechanisms for cross-framework abstractions in an outside project. 

I'm not sure I understand the rationale behind wanting to remove support for connection pooling, the behaviour is almost identical to HTTP/2?

florimondmanca:
My suggestion to set pooling aside for now was to make separate hosting easier. :)

The difference with your aioquic proof of concept would be that we'd be documenting HTTP/3 experimental support anyway (so even if that's as a separate thing we'd be targeting traffic at it).

Honestly though, given how self-contained and additive this pull request is (and will likely remain), I'm personally pretty open to having this as experimental in core. We accepted the AnyIO backend as a similar low-complexity additive change. And although I share the same reservations as Tom in terms getting the framing right for users (ie "you probably don't need the complexity that comes with HTTP/3"), I'm quite excited at the idea of demonstrating functional HTTP/3 support in HTTPCore/HTTPX. Would kind of prove the point that this "transport" architecture allows the project to evolve with industry trends.

tomchristie:
> I'm not sure I understand the rationale behind wanting to remove support for connection pooling, the behaviour is almost identical to HTTP/2?

So, the behaviour with this pull request between `http2=True` and `http3=True` is actually pretty different.

With `http2=True` enabled, you'll get *optional* HTTP/2 support, with either HTTP/1.1 or HTTP/2 connections being opened and connection pooled. The determination will vary per-server connection, depending on the ALPS negotiation. (IE. `http2=True` enables HTTP/2 support, rather than mandating it.)

In contrast, setting `http3=True` here *always* attempts an HTTP/3 connection, and will fail on requests against HTTP/1.1 or HTTP/2 servers.

Really what we'd *like* to do with HTTP/3 support is determine if support is available or not based on the AltSrv header, and only switch up to HTTP/3 where available. But that's a whole other kettle of fish.

What I would like to understand *first*, is what are the blockers to us instead *starting* with an incremental approach of http3-as-a-third-party-transport-to-start-with, and being able to document something like...

```python
import httpx, aioquic

# Use HTTP/3 support.
async with httpx.AsyncClient(transport=aioquic.HTTP3Transport()) as client:
    ...
```

Working from that perspective is really important to me, because there's always a *huge* volume of unsurfaced complexities that come along with pulling in something mega like HTTP/3 support, and the incremental approach almost always does a better job of helping thrash those out in a manageable step-by-step way.

There's a big difference between us advertising that support on https://www.python-httpx.org/ vs. example code in the aioquic GitHub repo.

The obvious blockers I can see would be...

* It'd potentially be asyncio-only to start with (because replicating our backend switching would be an unwanted pain.)
* There'd need to be a bit of connection pooling work put directly in to `HTTP3Transport`.

jlaine:
OK in essence we'll just maintain a variant of the status quo: the "transport" is a demo provided in the aioquic repo. It might graduate to being shipped with aioquic in an `asyncio.httpx.H3Transport`.

However note that it brings us no nearer to an eventual integration in httpcore or httpx, because the two are implemented very differently. This PR demonstrates my point, compare this code to the demo code in aioquic.  

For for the demo I don't have any incentive to use anything looking like the eventual "UDP socket abstraction" or "timer abstraction". If it's going to be async-io only, I can keep using the higher-level `QuicConnectionProtocol` which ships with aioquic.

tomchristie:
> For for the demo I don't have any incentive to use anything looking like the eventual "UDP socket abstraction" or "timer abstraction". If it's going to be async-io only, I can keep using the higher-level QuicConnectionProtocol which ships with aioquic.

Right, which from my POV is actually okay, because we can start with thrashing out any remaining issues in an asyncio-only context to begin with. Later once we're super happy with the state of all that and we're getting real world feedback, we can look at the next steps of a built-in integration.

I'm a bit absolute on the "take the smallest steps forward possible at each point" incremental approach. Anything else always seems to come back and bite ya. Hard.

PR #284, opened 2021-03-15
Fix README to use bytes for user-agent header

jlaine:
Headers are defined as a list of `Tuple[bytes, bytes]`, so the user-agent
header in the example needs to be `bytes`.


florimondmanca (approved):
Yes, thanks. :)

PR #302, opened 2021-04-10
Add docs i18n infrastructure, with partial FR translations

florimondmanca:
Follow-up to #285, prompted by https://github.com/encode/httpcore/pull/285#pullrequestreview-631335488

This pull request expands `scripts/docs` so that it supports generating translation files, and building/serving the docs for a given language. Currently only `fr` (French) is supported, since that's the language I'd be happy to contribute translations for. 😄 

Here's how things work:

* `scripts/docs maketranslations` - Sync translation files: this updates `.*po` files under `docs/locale/...` with any changes detected by Sphinx. Should be run to update translations after documentation has been edited.
* `SPHINX_LANGUAGE=<lang> scripts/docs build` - Build the `<lang>` version of the docs.
* `SPHINX_LANGUAGE=<lang> scripts/docs` - Serve (w/ reload, as for the English docs) the `<lang>` version of the docs. Any changes to `.*po` files are autodetected and trigger site reloads.

I wrote up translations for the home page to give an idea of what things look like. I'll do `api.md` and `contributing.md` as a follow-up.

Hosting the translated docs is a different matter and is not included in this PR (hence "Draft"), I'll discuss below.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ venv*/
 .python-version
 build/
 dist/
+*.mo
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -37,6 +37,12 @@ autodoc_typehints = "description"
 
 html_theme = "furo"
 
+# -- Internationalization --
+# https://www.sphinx-doc.org/en/master/usage/advanced/intl.html
+
+locale_dirs = ['locale/']
+gettext_compact = False
+
 # -- App setup --
 
 
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -130,6 +130,28 @@ To run the documentation site locally (useful for previewing changes), use:
 $ scripts/docs
 ```
 
+### Working with translations
+
+The documentation has multi-language (internationalization, aka i18n) support. The workflow is built on top of [Sphinx internationalization](https://www.sphinx-doc.org/en/master/usage/advanced/intl.html), which relies on the `gettext` standard.
+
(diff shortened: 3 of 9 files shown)


florimondmanca:
**Hosting**:

Easiest seems to be switching to ReadTheDocs (which would also give us easy versioning support, see https://github.com/encode/httpcore/pull/285#issuecomment-816009563).

[Localization of Documentation](https://docs.readthedocs.io/en/stable/localization.html) says that for multi-language sites, we need to create one project per translation there, so eg `httpcore` and `httpcore-fr`, indicating that `httpcore-fr` is a French translation for `httpcore`.

I'm not exactly sure how/if RTD will be able to use `scripts/docs`. In this PR I make `scripts/docs` accept a `SPHINX_LANGUAGE` environment variable to configure the target language. Not sure if RTD allows setting env vars, if it sets its own thing, calls Sphinx itself without us allowing to pass a custom script, etc. That's got to be investigated but unfortunately I don't have permissions to link with `httpcore` on RTD.

pradyunsg:
> I'm not exactly sure how/if RTD will be able to use `scripts/docs`.

No it won't. They run their own commands. [Here's a recent run for pip's documentation](https://readthedocs.org/projects/pip/builds/13477774/).

You can configure those commands somewhat using `.readthedocs.yml` in the repository root or via their Web UI -- I'd recommend using the file whenever possible though, since that's under version control. :)

florimondmanca:
Closing since I think we moved back to MkDocs, so this is stale. :)

Issue #149, opened 2020-08-11
AsyncConnectionPool leaks connection on CancelledError on 3.8


lazitski-aliaksei:
From python 3.8 `CancelledError` a is subclass of `BaseException` (https://docs.python.org/3.8/library/asyncio-exceptions.html#asyncio.CancelledError)

If coroutine is cancelled after connection was acquired and during request, connection is not removed from the pool:
https://github.com/encode/httpcore/blob/master/httpcore/_async/connection_pool.py#L179
```
try:
    response = await connection.request(
        method, url, headers=headers, stream=stream, timeout=timeout
    )
except NewConnectionRequired:
    connection = None
except Exception:
    logger.trace("remove from pool connection=%r", connection)
    await self._remove_from_pool(connection)
    raise
```

tomchristie:
Related #116

PR #305, opened 2021-04-19
asyncio connection_pool: catch CanceledError exception fix connection…

aviramha:
… leak in case of cacncellations
closes https://github.com/encode/httpcore/issues/149
closes https://github.com/encode/httpx/issues/1461

### DIFF
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -226,7 +226,7 @@ class AsyncConnectionPool(AsyncHTTPTransport):
                 connection = None
             except BaseException:  # noqa: PIE786
                 # See https://github.com/encode/httpcore/pull/305 for motivation
-                # behind catching 'BaseException' rather than 'Exception' here.
+                # behind catching 'BaseException' rather than 'Exception' here.
                 logger.trace("remove from pool connection=%r", connection)
                 await self._remove_from_pool(connection)
                 raise
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -225,6 +225,8 @@ class AsyncConnectionPool(AsyncHTTPTransport):
             except NewConnectionRequired:
                 connection = None
             except BaseException:  # noqa: PIE786
+                # See https://github.com/encode/httpcore/pull/305 for motivation
+                # behind catching 'BaseException' rather than 'Exception' here.
                 logger.trace("remove from pool connection=%r", connection)
                 await self._remove_from_pool(connection)
                 raise
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -224,7 +224,7 @@ class AsyncConnectionPool(AsyncHTTPTransport):
                 )
             except NewConnectionRequired:
                 connection = None
-            except Exception:  # noqa: PIE786
+            except BaseException:  # noqa: PIE786
                 logger.trace("remove from pool connection=%r", connection)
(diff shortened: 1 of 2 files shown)


florimondmanca:
Thanks @aviramha 

It seems we'd need to fit this into our async backends layer, so that it also works with eg trio (though I'm not sure if we can reproduce the issue there?).

Essentially I think we want the usage to be `except self.backend.cancelled_exc_class` instead, with that exception being `None` in the sync case (?).

aviramha:
I'm not sure what's the functional/design benefit from such behavior - this is the async implementation of connection pool, and catching "CanceledError" in trio won't really do anything anyway so why does it matter?

tomchristie:
So, Trio's `Cancelled` exception *also* inherits from `BaseException`, yup... https://github.com/python-trio/trio/blob/4e8e19cff8fd5c1b5875d36c77564a6ec021330a/trio/_core/_exceptions.py#L31

I don't think it'd be a *terrible* idea for us to just catch `BaseException` here. We *are* always re-raising the exception.

Either...

```python
            except NewConnectionRequired:
                 connection = None
             except BaseException:  # noqa: PIE786
                 # Note that we catch BaseException here in order to also support
                 # removing connections from the pool on task cancellations.
                 logger.trace("remove from pool connection=%r", connection)
                 await self._remove_from_pool(connection)
                 raise
```

Or perhaps...

```python
            except NewConnectionRequired:
                 connection = None
             except BaseException:  # noqa: PIE786
                 try:
                     logger.trace("remove from pool connection=%r", connection)
                     await self._remove_from_pool(connection)
                 finally:
                     raise
```

httpcore/_sync/connection_pool.py
```diff
@@ -1,4 +1,5 @@
 import warnings
+from asyncio import CancelledError
```
tomchristie, line 2:
```suggestion
```

httpcore/_sync/connection_pool.py
```diff
@@ -224,7 +225,7 @@ def request(
                 )
             except NewConnectionRequired:
                 connection = None
-            except Exception:  # noqa: PIE786
+            except (Exception, CancelledError):  # noqa: PIE786
```
tomchristie, line 228:
```suggestion
            except BaseException:  # noqa: PIE786
```

httpcore/_async/connection_pool.py
```diff
@@ -224,7 +224,7 @@ async def arequest(
                 )
             except NewConnectionRequired:
                 connection = None
-            except Exception:  # noqa: PIE786
+            except BaseException:  # noqa: PIE786
```
tomchristie, line 227:
```suggestion
            except BaseException:  # noqa: PIE786
                # See https://github.com/encode/httpcore/pull/305 for motivation
                # behind catching 'BaseException' rather than 'Exception' here.
```

httpcore/_sync/connection_pool.py
```diff
@@ -224,7 +224,7 @@ def request(
                 )
             except NewConnectionRequired:
                 connection = None
-            except Exception:  # noqa: PIE786
+            except BaseException:  # noqa: PIE786
```
tomchristie, line 227:
```suggestion
            except BaseException:  # noqa: PIE786
                # See https://github.com/encode/httpcore/pull/305 for motivation
                # behind catching 'BaseException' rather than 'Exception' here.
```

tomchristie:
Thanks @aviramha!

aviramha:
Thank you for this high maintenance library!

PR #313, opened 2021-04-28
Raise ConnectError when server disconnects without sending a response

tomchristie:
Refs: https://github.com/encode/httpx/issues/1478

Currently if the server is in a listening state, then disconnects without sending a response, then users will see the following error...

```python
httpcore.RemoteProtocolError: can't handle event type ConnectionClosed when role=SERVER and state=SEND_RESPONSE
```

Which really isn't informative as an end-user. We probably would prefer to special case this particular transition as a `ConnectError`. It's not quite the same as a *network* `ConnectError`, since it occurs in cases where the connection was already established and hanging around, but is closed once it an incoming request is sent.

It's not super obvious if we want to treat this case as a `ConnectError` or something else. Here's our existing hierarchy in this area...

* NetworkError
  * ConnectError
  * ReadError
  * WriteError
  * CloseError
* ProtocolError
  * LocalProtocolError
  * RemoteProtocolError

### DIFF
--- a/httpcore/_async/http11.py
+++ b/httpcore/_async/http11.py
@@ -5,7 +5,7 @@ import h11
 
 from .._backends.auto import AsyncSocketStream
 from .._bytestreams import AsyncIteratorByteStream
-from .._exceptions import ConnectError, LocalProtocolError, RemoteProtocolError, map_exceptions
+from .._exceptions import LocalProtocolError, RemoteProtocolError, map_exceptions
 from .._types import URL, Headers, TimeoutDict
 from .._utils import get_logger
 from .base import AsyncByteStream, ConnectionState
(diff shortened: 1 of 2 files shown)


tomchristie:
Requests treats these as `ConnectionError`. URLLib3 treats them as `ProtocolError`.

See: https://github.com/psf/requests/blob/c45a4dfe6bfc6017d4ea7e9f051d6cc30972b310/requests/adapters.py#L497

Having reviewed this it makes more sense to me for use to keep the exception class the same as it otherwise would be, but issue a clearer message. So now...

```python
httpcore.RemoteProtocolError: Server disconnected without sending a response.
```

tomchristie:
Though we could also introduce a specific new class, eg...

* NetworkError
  * ConnectError
  * ReadError
  * WriteError
  * CloseError
* ProtocolError
  * LocalProtocolError
  * RemoteProtocolError
  * ServerDisconnected

The one place where our behaviour is important here is wrt. retry strategies.

PR #345, opened 2021-05-24
Use AnyIO by default on asyncio

### LINKED ISSUE #344: Switching to `anyio` as the default backend when running under `asyncio`.
Having been bitten by [some rough edges in `asyncio`'s SSL support](https://github.com/encode/httpx/discussions/1511), and having taken a look over `anyio`'s TLSStream implementation. (Which just plain *makes sense* \*) I'm now very warm on the idea of us switching over to `anyio` as the default backend for the `asyncio` case.

We *could* consider deprecating and later removing the native `curio.py` and `asyncio.py` modules, but we don't necessarily need to do that to start with.

First step here would be switching the default, and issuing a release, in order to make sure we hit any unexpected issues that might crop up as a result.

Paging @agronholm. 😀

---

\*: In contrast to this, what the heck is `loop.start_tls`? Why is that a property of the *event loop*? Etc.

agronholm:
Closes #344.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ setup(
     packages=get_packages("httpcore"),
     include_package_data=True,
     zip_safe=False,
-    install_requires=["h11>=0.11,<0.13", "sniffio==1.*"],
+    install_requires=["h11>=0.11,<0.13", "sniffio==1.*", "anyio>=3.1"],
     extras_require={
         "http2": ["h2>=3,<5"],
     },
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ setup(
     packages=get_packages("httpcore"),
     include_package_data=True,
     zip_safe=False,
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -53,7 +53,7 @@ def get_packages(package):
     packages=get_packages("httpcore"),
     include_package_data=True,
     zip_safe=False,
-    install_requires=["h11>=0.11,<0.13", "sniffio==1.*"],
+    install_requires=["h11>=0.11,<0.13", "sniffio==1.*", "anyio>=3.1"],
```
tomchristie, line 56: Let's roll with `anyio==3.*` right?

tomchristie (approved):

tomchristie:
Fantastic, yup.

I'd suggest we'll want to roll a new release with this.

PR #353, opened 2021-06-08
Better encapsulation for connection instances

tomchristie:
This pull request introduces a refactoring to more tightly encapsulate Connection instances, so that the connection pool isn't reaching inside implementation details, but instead has a neat interface to work against. This'll be a precursor to further work on tightening up the implementation, ensuring we're more obviously guarded against potential race conditions, etc...

In addition to the existing `handle_request` and `close` methods, the interface the connection pool uses here are...

* `should_close` - The connection has expired its keep alive, or been closed by the remote end.
* `is_idle` - The connection is currently idle. It may optionally be removed from the pool, if needed.
* `is_closed` - The connection is closed. It should be removed from the pool.
* `is_available` - The connection is in a state where it can current accept another request.

An obvious stand-alone follow up to this would be to make other methods and attributes on the connection private.


tomchristie:
Reading through this, one possible tweak that stands out... perhaps `may_close` should be `is_idle`?

florimondmanca (commented):
Gave it a quick look-over, really like the "hide the 'state' implementation details" idea. :+1:

tomchristie:
Alrighty. I'm going to push ahead with this, and roll a new release.

We need to get this all into the wild particularly since we've switched to `anyio` as our default backend for `asyncio`.

There's a bunch of follow up I want to do from this one onwards, really tightening up any possible race condition areas, getting coverage all the way up, sorting out tls-over-tls support etc... and generally getting `httpcore` up to the same standard of quality that I feel the `httpx` repo is now in.

PR #368, opened 2021-06-22
Test under 3.10.0-beta3

j178:
- `str(ConnectionState.ACTIVE)` returns `"ACTIVE"` under 3.10. [link](https://docs.python.org/3/whatsnew/3.10.html#enum)


tomchristie (approved):
Great bit of work, here!

PR #364, opened 2021-06-15
configure strict pytest

graingert:

PR #378, opened 2021-07-01
Treat warnings as errors

j178:
Fixes https://github.com/encode/httpcore/issues/384

Follow up #368

- Use `ssl.create_default_context()` instead `ssl.SSLContext()` to avoid some ssl deprecation warnings
- Add a `ssl_context` fixture that trusts our self issued certificates for localhost
- Minor fix for the enum repr change in 3.10
- Close the UDS socket if connection error

### DIFF
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -1,4 +1,4 @@
-from ssl import SSLContext
+from ssl import SSLContext, PROTOCOL_TLS_CLIENT
 from typing import List, Optional, Tuple, cast
 
 from .._backends.auto import AsyncBackend, AsyncLock, AsyncSocketStream, AutoBackend
@@ -33,7 +33,7 @@ class AsyncHTTPConnection(AsyncHTTPTransport):
         self._http2_enabled = http2
         self._keepalive_expiry = keepalive_expiry
         self._uds = uds
-        self._ssl_context = SSLContext() if ssl_context is None else ssl_context
+        self._ssl_context = SSLContext(PROTOCOL_TLS_CLIENT) if ssl_context is None else ssl_context
         self.socket = socket
         self._local_address = local_address
         self._retries = retries
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -1,4 +1,4 @@
-from ssl import SSLContext, PROTOCOL_TLS_CLIENT
+from ssl import PROTOCOL_TLS_CLIENT, SSLContext
 from typing import List, Optional, Tuple, cast
 
 from .._backends.auto import AsyncBackend, AsyncLock, AsyncSocketStream, AutoBackend
@@ -33,7 +33,9 @@ class AsyncHTTPConnection(AsyncHTTPTransport):
         self._http2_enabled = http2
         self._keepalive_expiry = keepalive_expiry
         self._uds = uds
-        self._ssl_context = SSLContext(PROTOCOL_TLS_CLIENT) if ssl_context is None else ssl_context
+        self._ssl_context = (
+            SSLContext(PROTOCOL_TLS_CLIENT) if ssl_context is None else ssl_context
+        )
         self.socket = socket
         self._local_address = local_address
         self._retries = retries
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -1,3 +1,4 @@
+import ssl
 from ssl import SSLContext
 from typing import List, Optional, Tuple, cast
 
@@ -33,7 +34,9 @@ class AsyncHTTPConnection(AsyncHTTPTransport):
         self._http2_enabled = http2
         self._keepalive_expiry = keepalive_expiry
         self._uds = uds
-        self._ssl_context = SSLContext() if ssl_context is None else ssl_context
+        self._ssl_context = (
+            ssl.create_default_context() if ssl_context is None else ssl_context
+        )
         self.socket = socket
         self._local_address = local_address
         self._retries = retries
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -1,9 +1,9 @@
-import ssl
 from ssl import SSLContext
 from typing import List, Optional, Tuple, cast
 
 from .._backends.auto import AsyncBackend, AsyncLock, AsyncSocketStream, AutoBackend
 from .._exceptions import ConnectError, ConnectTimeout
+from .._ssl import default_ssl_context
 from .._types import URL, Headers, Origin, TimeoutDict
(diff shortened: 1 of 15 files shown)


httpcore/_async/http2.py
```diff
@@ -51,7 +51,7 @@ def __init__(
         self._exhausted_available_stream_ids = False
 
     def __repr__(self) -> str:
-        return f"<AsyncHTTP2Connection [{self._state}]>"
+        return f"<AsyncHTTP2Connection [{self._state.name}]>"
```
j178, line 54:
```DeprecationWarning: in 3.12 format() will use the enum member, not the enum member's value;
use a format specifier, such as :d for an IntEnum member, to maintain the current display.
```

httpcore/_backends/sync.py
```diff
@@ -155,13 +155,17 @@ def open_uds_stream(
...
+            try:
+                sock.settimeout(connect_timeout)
+                sock.connect(path)
+
+                if ssl_context is not None:
+                    sock = ssl_context.wrap_socket(
+                        sock, server_hostname=hostname.decode("ascii")
+                    )
+            except OSError:
+                sock.close()
```
j178, line 167:
```
ResourceWarning: unclosed <socket.socket fd=11, family=AF_UNIX, type=SOCK_STREAM, proto=0>
```

graingert:
There's a PR here btw https://github.com/encode/httpcore/pull/364

httpcore/_async/connection.py
```diff
@@ -33,7 +34,9 @@ def __init__(
         self._http2_enabled = http2
         self._keepalive_expiry = keepalive_expiry
         self._uds = uds
-        self._ssl_context = SSLContext() if ssl_context is None else ssl_context
+        self._ssl_context = (
+            ssl.create_default_context() if ssl_context is None else ssl_context
```
graingert, line 38: create_default_context creates a secure context but SSLContext() creates an insecure one

graingert: is it work moving https://github.com/encode/httpx/blob/b43af721cd7804214396b5511e4f68d144a9b6a6/httpx/_config.py#L44 into httpcore?

httpcore/_async/connection.py
```diff
@@ -1,3 +1,4 @@
+import ssl
 from ssl import SSLContext
```
graingert, line 2: would be nice to pick one import style and stick with it

httpcore/_backends/sync.py
```diff
@@ -156,12 +156,16 @@ def open_uds_stream(
         with map_exceptions(exc_map):
             sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
             sock.settimeout(connect_timeout)
-            sock.connect(path)
```
graingert, line 159: you can still get a ResourceWarning if this socket is GCd after a KeyboardInterrupt

```
            try:
                sock.connect(path)
            except BaseException:
                sock.close()
                raise
```

tests/conftest.py
```diff
@@ -140,6 +142,21 @@ def localhost_cert_private_key_file(
...
 
 
+@pytest.fixture(scope="session")
+def ssl_context(
+    cert_authority: trustme.CA, localhost_cert: trustme.LeafCert
+) -> ssl.SSLContext:
+    ssl_context = ssl.create_default_context()
+
+    if hypercorn is None:
+        ssl_context.load_verify_locations(certifi.where())
```
graingert, line 152: create_default_context should already load the correct verify locations

tests/utils.py
```diff
@@ -179,6 +179,7 @@ def http_proxy_server(proxy_host: str, proxy_port: int):
         finally:
             if proc is not None:
                 proc.kill()
+                proc.wait()
```
graingert, line 182:
```suggestion
                proc.communicate()
```

httpcore/_async/http2.py
graingert: https://www.python.org/dev/peps/pep-0663/ that's this right?

graingert: I think we should change the AsyncHTTP2Connection to use the repr of its state

```suggestion
        return f"<AsyncHTTP2Connection [{self._state!r}]>"
```

j178: `Enum.__repr__` has different behavior between python3.10+ and below, we'll get different output among python versions.

graingert: That's working as designed imho

httpcore/_async/connection.py
tomchristie: I think we'd like to have a minimal-as-possible SSL context in `httpcore`, eg...

https://github.com/tomchristie/httpcore-the-directors-cut/blob/main/httpcore/_ssl.py


```python
import certifi
import ssl


def default_ssl_context() -> ssl.SSLContext:
    context = ssl.create_default_context()
    context.load_verify_locations(certifi.where())
    return context
```

graingert: I don't think `certifi.where()` is needed anymore in py3.6+

tests/conftest.py
tomchristie: I don't know if we need `certifi` in *this particular case* or not, but just to help make this clearer for anyone following the conversation...

Python's `ssl.create_default_context()` doesn't hook into the system certificate store, so it's actually *not*  able to correctly validate https hostnames...

```python
>>> import ssl, socket
>>> context = ssl.create_default_context()
>>> sock = socket.create_connection(("www.example.com", 443))
>>> ssl_sock = context.wrap_socket(sock, server_hostname="www.example.com")
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ssl.py", line 423, in wrap_socket
    session=session
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ssl.py", line 870, in _create
    self.do_handshake()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ssl.py", line 1139, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1076)
```

Which is why certifi is needed (A third party package which bundles up Mozilla's certificate store of root certificates)...

```python
>>> import certifi, ssl, socket
>>> context = ssl.create_default_context()
>>> context.load_verify_locations(certifi.where())
>>> sock = socket.create_connection(("www.example.com", 443))
>>> ssl_sock = context.wrap_socket(sock, server_hostname="www.example.com")
```

tomchristie:
Closing as per #364

PR #379, opened 2021-07-03
Add an unclosed warning when connection is being destroyed

j178:
Based on #378 
- Add a warning to `BaseHTTPConnection.__del__`
- Ensure connections are closed in some cases


httpcore/_async/http_proxy.py
```diff
@@ -268,20 +273,28 @@ async def _tunnel_request(
             )
             await self._add_to_pool(connection, timeout)
 
+            proxy_connection.connection.socket = None  # type: ignore
+            await proxy_connection.aclose()
```
j178, line 277: Set .socket to None here, then aclose would not close this socket.

tomchristie:
Closing as per #364

PR #380, opened 2021-07-03
Remove `--use-feature` switch from pip install command

j178:
```
WARNING: --use-feature=2020-resolver no longer has any effect, since it is now the default dependency resolver in pip. This will become an error in pip 21.0.
```


tomchristie:
Ace, thanks!

PR #443, opened 2021-11-30
Prevent orphaned TCP socket

mueslo:
When an exception occurs during the TLS layer creation, the created TCP socket is not connected to any connection and thus never closed leading to a TCP socket leak, see https://github.com/home-assistant/core/issues/60150

### DIFF
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -143,7 +143,11 @@ class AsyncHTTPConnection(AsyncConnectionInterface):
                 "timeout": timeout,
             }
             async with Trace("connection.start_tls", request, kwargs) as trace:
-                stream = await stream.start_tls(**kwargs)
+                try:
+                    stream = await stream.start_tls(**kwargs)
+                except Exception:
+                    await stream.aclose()
+                    raise
                 trace.return_value = stream
         return stream
 
--- a/httpcore/_sync/connection.py
+++ b/httpcore/_sync/connection.py
@@ -143,7 +143,11 @@ class HTTPConnection(ConnectionInterface):
                 "timeout": timeout,
             }
             with Trace("connection.start_tls", request, kwargs) as trace:
-                stream = stream.start_tls(**kwargs)
+                try:
+                    stream = stream.start_tls(**kwargs)
+                except Exception:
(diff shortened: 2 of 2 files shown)


mueslo:
Or should it directly go here? https://github.com/encode/httpcore/blob/0e8bc1f0591b05f40a9c8d0569ba1a914cee4d14/httpcore/backends/asyncio.py#L60-L69

florimondmanca:
Hi! This repository is in the dying end — we've now moved core networking code to `httpx` as of HTTPX 0.21.0.

Are you still encountering this issue while running against HTTPX 0.21.0?

mueslo:
@florimondmanca Yes, the issue occurred with the latest versions of httpx+httpcore (see linked homeassistant thread).

tomchristie:
Thanks, good work in tracking this down!

Wonder if we ought to resolve this at the backend level instead. (So that the `start_tls()` method is responsible for closing the socket if it fails.)

mueslo:
Thanks for the response! 

Yes, I thought about that too, but since 1) it would need to be added to each backend separately and since 2) perhaps in some cases you might want to retry establishing a TLS layer on the same TCP connection I went with the simplest choice of just catching the exception in the uppermost layer possible.

A feasible alternative I think could be to add the TCP stream as a member of the AsyncHTTPConnection class so it can be correctly removed if it exists when the connection is closed/removed, but since the TCP stream is not really used anywhere outside the AsyncHTTPConnection._connect method I thought it'd be unnecessary complexity.

elupus:
We have confirmation that this patch indeed solves our issues. https://github.com/home-assistant/core/issues/57093#issuecomment-997721428

mueslo:
@tomchristie any chance of getting this (or an equivalent) fix merged before early/mid January so the fix can be shipped in HomeAssistant 2022.2?

elupus:
@mueslo can you rebase this?

nikrays:
My desire for the new year was to fix this problem, now another update has arrived, but I see that the problem has not been fixed, what is fashionable to do?🤦🏻

tomchristie:
I've addressed this in #475

PR #442, opened 2021-11-18
configure strict pytest

graingert:

PR #473, opened 2022-01-05
Drop unused testing requirements

tomchristie:
We have several test requirements that are actually no longer being used.

For context, this is a result of the drastic 0.14 re-working, which has made the package much more testable throughout.
We're generally testing against mocked streams, rather than against uvicorn/hypercorn. We have *far* better coverage now than we had before (100%, with only *explicit* exemptions.)

The integration testing we do have is currently using `pytest-httpbin`. Personally I'd like to drop this too in favour of something more explicit, but that's a story for another day.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,4 +28,3 @@ pytest==6.2.5
 pytest-httpbin==1.0.1
 pytest-trio==0.7.0
 pytest-asyncio==0.16.0
-trustme==0.9.0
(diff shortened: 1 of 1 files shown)


tomchristie:
> The integration testing we do have is currently using pytest-httpbin. Personally I'd like to drop this too in favour of something more explicit, but that's a story for another day.

See also this comment: https://github.com/encode/httpcore/pull/442#issuecomment-973003001

PR #475, opened 2022-01-05
If TLS fails, then close the socket

tomchristie:
Alternative to #443


tomchristie:
This fixes the same issue as that alternative approach, yes.

nikrays:
> This fixes the same issue as that alternative approach, yes.

What update will the home assistant include?  The library isn't theirs, is it?

nikrays:
Did you make this change as per the patch?

https://github.com/encode/httpcore/compare/master...mueslo:patch-1

PR #478, opened 2022-01-12
Socks support

tomchristie:
Initial pass at socks 5 support.

Right now this just adds the `_init_socks5_connection` function. This will need wiring into a proxy class.


florimondmanca:
Super glad to see this in with `socksio` :-) :+1:

PR #491, opened 2022-01-27
Resolve race condition around exceptions during streaming a response.

tomchristie:
Closes https://github.com/encode/httpx/issues/2047 by resolving a race condition that can occur when an exception is raised while streaming the response.

Prior to this PR it's possible that the exception is handled first, closing the connection, and *then* the request close callback occurs.

The changes are:

* If an exception occurs while streaming a response, then close the response (and possibly the connection) 
**before allowing the exception to bubble up**.
* If a connection pool is closed while there are still open requests, then close it but then raise a `RuntimeError`.
* Defensive programming around removing the request status from the connection pool, when closing the request.

### DIFF
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -267,7 +267,8 @@ class AsyncConnectionPool(AsyncRequestInterface):
 
         async with self._pool_lock:
             # Update the state of the connection pool.
-            self._requests.remove(status)
+            if status in self._requests:
+                self._requests.remove(status)
 
             if connection.is_closed() and connection in self._pool:
                 self._pool.remove(connection)
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -239,7 +239,7 @@ class AsyncConnectionPool(AsyncRequestInterface):
                     # status so that the request becomes queued again.
                     status.unset_connection()
                     await self._attempt_to_acquire_connection(status)
-            except Exception as exc:
+            except BaseException as exc:
                 await self.response_closed(status)
                 raise exc
             else:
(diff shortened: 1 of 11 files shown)


httpcore/_async/connection_pool.py
```diff
@@ -291,11 +292,19 @@ async def aclose(self) -> None:
...
+            requests_still_in_flight = bool(self._requests)
+
             for connection in self._pool:
                 await connection.aclose()
             self._pool = []
             self._requests = []
 
+            if requests_still_in_flight:
+                raise RuntimeError(
+                    "The connection pool was closed while some HTTP "
```
ahopkins, line 304: Instead of "some", it should be possible to get a number correct (`len(self._requests)`)? I know I have seen this in other exceptions or repr. I certainly appreciated that when I have seen it.

httpcore/_async/http11.py
```diff
@@ -279,13 +279,23 @@ class HTTP11ConnectionByteStream:
...
+            # we want to close the response (and possibly the connection)
+            # before raising that exception.
+            await self.aclose()
+            raise exc
 
     async def aclose(self) -> None:
-        async with Trace("http11.response_closed", self._request):
-            await self._connection._response_closed()
+        if not self._closed:
+            self._closed = True
```
ahopkins, line 299: Any potential problem for setting this value before `self._connection._response_closed()`? It does not need to be set after?

ahopkins: I guess if the point is to not run this multiple times it makes sense to be up front. But, thought I would ask anyway.

ahopkins (approved):

tomchristie:
@ahopkins Great thanks - addressed one of your comments. Note that you'll need to accept the @encode/maintainers invitation before your review gets "merge approval" status.

ahopkins (approved):

PR #502, opened 2022-02-02
Requests which raise a PoolTimeout need to be removed from the pool queue.

tomchristie:
Since https://github.com/encode/httpcore/pull/491 we've started raising a `RuntimeError` if the connection pool is closed while still handling some requests.

This caused an error in the `httpx` test suite, where a pool timeout was being raised causing the client to close. In this case we *hadn't* removed the request from the connection pool.

This pull request resolves that, as well as resolving the case of task-cancellation-of-a-queued-waiting-request.


aminalaee (approved):
Looks good.

PR #503, opened 2022-02-02
Raise H11 version support for v.0.13

fsecada01:
- Allows for concurrent installation with the latest Uvicorn version
- Drop support for Python v.3.6

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,7 @@ setup(
         "Framework :: AsyncIO",
         "Framework :: Trio",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.6",
+#         "Programming Language :: Python :: 3.6",  # remove py36 from active support
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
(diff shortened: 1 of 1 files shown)


tomchristie:
Alrighty - `h11` 0.13 adds type annotations, which as it happens causes our CI to fail.
The code is correct, by it fails `mypy`, which isn't *super* surprising.
Needs a bit of tweaking so that we're still passing `mypy`'s type checking.

setup.py
```diff
@@ -54,7 +54,7 @@ def get_packages(package):
     include_package_data=True,
     zip_safe=False,
     install_requires=[
-        "h11>=0.11,<0.13",
+        "h11>=0.11,<=0.13",
```
mgorny, line 57:
```suggestion
        "h11>=0.11,<0.14",
```

It's rather unlikely that e.g. 0.13.1 will break anything.

GalaxySnail:
I think it may not be a good idea to pin h11 under 0.14, it can cause unnecessary conflicts. See: https://hynek.me/articles/semver-will-not-save-you/

PR #501, opened 2022-02-01
Fix variable declaration in AsyncSocks5Connection __init__ method

devl00p:
Playing with the latest httpcore and httpx versions I noticed an unexpected exception when the connection failed using a SOCKS proxy.

```
httpcore==0.14.6
httpx==0.22.0
```

```
ERROR:asyncio:Task exception was never retrieved
future: <Task finished name='Task-8' coro=<Explorer.async_analyze() done, defined at /home/sirius/wapiti/wapitiCore/net/crawler.py:798> exception=AttributeError("'AsyncSocks5Connection' object has no attribute '_connect_failed'")>
Traceback (most recent call last):
  File "/home/sirius/wapiti/wapitiCore/net/crawler.py", line 823, in async_analyze
    page = await self._crawler.async_send(request)
  File "/home/sirius/wapiti/wapitiCore/net/crawler.py", line 592, in async_send
    page = await self.async_get(resource, headers=headers, follow_redirects=follow_redirects)
  File "/home/sirius/wapiti/wapitiCore/net/crawler.py", line 116, in inner_wrapper
    value = await function(*args, **kwargs)
  File "/home/sirius/wapiti/wapitiCore/net/crawler.py", line 467, in async_get
    response = await self.client.send(
  File "/home/sirius/.local/share/virtualenvs/wapiti-p7I6n6KS/lib/python3.8/site-packages/httpx/_client.py", line 1593, in send
    response = await self._send_handling_auth(
  File "/home/sirius/.local/share/virtualenvs/wapiti-p7I6n6KS/lib/python3.8/site-packages/httpx/_client.py", line 1621, in _send_handling_auth
    response = await self._send_handling_redirects(
  File "/home/sirius/.local/share/virtualenvs/wapiti-p7I6n6KS/lib/python3.8/site-packages/httpx/_client.py", line 1658, in _send_handling_redirects
    response = await self._send_single_request(request)
  File "/home/sirius/.local/share/virtualenvs/wapiti-p7I6n6KS/lib/python3.8/site-packages/httpx/_client.py", line 1695, in _send_single_request
    response = await transport.handle_async_request(request)
  File "/home/sirius/.local/share/virtualenvs/wapiti-p7I6n6KS/lib/python3.8/site-packages/httpx/_transports/default.py", line 353, in handle_async_request
    resp = await self._pool.handle_async_request(req)
[...]

PR #504, opened 2022-02-04
Version 0.14.7

aminalaee:
https://github.com/encode/httpcore/releases/edit/untagged-5bb4ab3a7dc459eb230f

## 0.14.7 (February 4th, 2022)

 - Requests which raise a PoolTimeout need to be removed from the pool queue. (#502)
 - Fix AttributeError that happened when Socks5Connection were terminated. (#501)


tomchristie (approved):
Fantastic yup.

PR #507, opened 2022-02-09
connect_to and sni_hostname extensions for controlling

toppk:
connect_to and sni_hostname extensions for controlling http/tls/ip hostnames separately.  for the messy world we live in.

I choose to not clutter the kwargs dictionary creation, to make the readers eyes less sore.

I think if you wanted to future proof this, you'd make connect_to an array instead of a single string, but I think it would be trivial to handle either  in the future if a decision is made to acknowledge a hostname can have multiple ip addresses :)

I will look at creating a test case can be made for these extensions.  But putting this in for early feedback.

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -68,6 +68,43 @@ print(r.extensions["http_version"])
 
 ## Request Extensions
 
+### `"connect_to"`
+
+A string containing the desired hostname or ip address to open the tcp 
+socket.
+
+This will not affact which hostname is used for initiating the TLS stream, please
+look at `"sni_hostname"` for that.
+
+For example:
+
+```python
+# Use connect_to extention to target a specific server in a pool of servers.
+r = httpcore.request(
+    "GET",
+    "https://www.example.com",
+    extensions={"connect_to": "www1.example.com"}
+)
+```
+
+### `"sni_hostname"`
+
+A string containing the desired hostname or ip address use during TLS handshake.
+
+This will not affact which hostname is used for initiating the connection, please
+look at `"connect_to"` for that.
+
+For example:
+
+```python
+# The certificates in use by the server are for a different hostname.
+r = httpcore.request(
+    "GET",
+    "https://localhost",
+    extensions={"timeout": {"sni_hostname": "www.example.com"}}
+)
+```
+
 ### `"timeout"`
 
 A dictionary of `str: Optional[float]` timeout values.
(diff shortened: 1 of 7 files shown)


toppk:
After a cursory review of the test suite it seems to me that the testsuite is entirely using mock backend, therefore any test with a mock would require  would be  testing an implementation inside mock backend and not the actual code.   Additionally since these socket details aren't currently inside the public api of httpcore or httpx (e.g. exposing the IP address of the server) i don't think it would be worth it at this point.

Therefore I am considering this done pending feedback.

toppk:
just saw your comment on connect_to in the discussion, I'm curious to understand the complexity...

toppk:
to exercise this (using httpx) 


```
import httpx
client = httpx.Client()
resp = client.get("https://www.foobard.org/headers", extensions={"connect_to": "54.221.78.73", "sni_hostname":"httpbin.org"} )
from pprint import pprint
pprint( resp.json() )
```

toppk:
yeah, actually I'm thinking about this and you've got a point.  This current approach will lead to the issue I'm facing in my custom Transport where can_handle_request is returning true, when I've changed the "connect_to" indication.  So for my needs, it would be better just to avoid connect_to entirely, and leave the connection pool as a bunch of IP addresses.  at least until there becomes a reason to add the socket details into origin/connectioninterface.

I think there is still merit to the connect_to extension as a convenience, but I do think the implementation should change to manupulate url.host,headers["host"],sni_hostname as needed.  Even if we remove it as an extension entirely, there is still merit to creating a convenience argument for the command line.  

please let me know if this should still be an extension and I'll fix the implementation.  If there is no extension, I'll change the commandline option from "--x-connect-to" to just "--connect-to"

toppk:
Just so you could see what an extension would look like to see if you have a stomach for it, it would be something like this inside httpcore.Request.   (This is untested for, didn't look hard at the byte/str stuff, this is just to help your decision)

```
--- a/httpcore/_models.py
+++ b/httpcore/_models.py
@@ -342,6 +342,12 @@ class Request:
             content, name="content"
         )
         self.extensions = {} if extensions is None else extensions
+        if "connect_to" in self.extensions:
+            hostname = self.url.host
+            self.url.host = self.extensions["connect_to"]
+            if "sni_hostname" not in self.extensions:
+                self.extensions["sni_hostname"] = hostname
+            self.headers[b"host"] = hostname.encode("ascii")
 

```

clapbr:
this works great for me, don't let this PR die.

Issue #509, opened 2022-02-18
Consider dropping upper pin on h11


madkinsz:
Hi! By pinning the upper version of h11 in `setup.py` each time they make a release projects that depend on `httpcore` can fail with version requirement issues.

For example:
```
ERROR: httpcore 0.14.7 has requirement h11<0.13,>=0.11, but you'll have h11 0.13.0 which is incompatible.
...
pkg_resources.ContextualVersionConflict: (h11 0.13.0 (/.../python3.8/lib/python3.8/site-packages), Requirement.parse('h11<0.13,>=0.11'), {'httpcore'})
```

Could you remove the upper pin and allow consumers of this important library to use compatible versions on release or restrict the version themselves if necessary?

Of course, part of this is `pip`'s fault as it should be resolving these constraints, but it does not appear to do so consistently.

A discussion of upper bounds: https://iscinumpy.dev/post/bound-version-constraints/

Related to: https://github.com/encode/httpcore/pull/503

This could be extended to all of the versioning in the `setup.py` which mostly pin to major versions.

tomchristie:
I think that'd probably be okay.

As you can see from #503 we've a couple of changes we'd need to make to ensure our tests still pass. (Not because the behaviour is wrong, just because it's structured in a way that it fails type checking against the newly added annotations.)

The block to this is probably having someone work on that.

madkinsz:
Hi bot, this issue is still relevant. The upper pin remains and furthermore httpcore is not compatible with h11 0.13+.

madkinsz:
A decision is still needed on https://github.com/encode/httpcore/pull/526 to support h11 0.13.

tomchristie:
Boop. https://github.com/encode/httpcore/pull/549

graingert:
> A decision is still needed on #526 to support h11 0.13.

The decision is about PAUSED handling? If so that's resolved

tomchristie:
Right, let's consider https://github.com/encode/httpcore/pull/579 as resolving this.
We'll keep the pinning (because, we're careful like that) but we've bumped up the version to the latest.

PR #526, opened 2022-03-25
Update types for h11 v0.13

madkinsz:
Replaces #503, performing the additional work necessary for type checks to pass.

A cast is necessary because h11 does not expose the `Sentinel` type and uses it as a return type annotation instead of specifying the subtypes that can be returned.

```python
event = cast(
    Union[h11.Event, h11.NEED_DATA, h11.PAUSED],
    self._h11_state.next_event(),
)
```

I've opened a patch to h11 to update `next_event` to have a return type that is narrower and a part of their public API: https://github.com/python-hyper/h11/pull/144 — Note this is now merged resolving the need for a cast, but is not yet released.

There is currently no handling in `_receive_event` for the `h11.PAUSED` sentinel event. I am not sure what behavior you'd like to add there and have left a stub that should be filled in before this is merged. The documentation for `PAUSED` can be found at https://h11.readthedocs.io/en/latest/api.html#flow-control — Note this is not a concern and handling is performed [here](https://github.com/encode/httpcore/pull/526/files#diff-f1329e0e83048035c0a643ab5f3c666ae6565cba6393111fc0dc1afc4f2dbaa4R168).

Note this partially addresses #509 by allowing a newer version of h11 to be used, but does not close the issue as it requests that the upper pin be removed entirely.

Closes #503
Closes #498

### DIFF
--- a/httpcore/_async/http11.py
+++ b/httpcore/_async/http11.py
@@ -8,8 +8,8 @@ from typing import (
     Optional,
     Tuple,
     Type,
-    cast,
     Union,
+    cast,
 )
 
 import h11
@@ -199,7 +199,7 @@ class AsyncHTTP11Connection(AsyncConnectionInterface):
                 # TODO: Implement handling for paused
                 ...
             else:
-                return event
+                return cast(h11.Event, event)
 
     async def _response_closed(self) -> None:
         async with self._state_lock:
--- a/httpcore/_async/http11.py
+++ b/httpcore/_async/http11.py
@@ -168,7 +168,9 @@ class AsyncHTTP11Connection(AsyncConnectionInterface):
             elif isinstance(event, (h11.EndOfMessage, h11.PAUSED)):
                 break
 
-    async def _receive_event(self, timeout: Optional[float] = None) -> h11.Event:
+    async def _receive_event(
+        self, timeout: Optional[float] = None
+    ) -> Union[h11.Event, h11.PAUSED]:
         while True:
             with map_exceptions({h11.RemoteProtocolError: RemoteProtocolError}):
                 # The h11 type signature uses a private return type
@@ -177,7 +179,7 @@ class AsyncHTTP11Connection(AsyncConnectionInterface):
                     self._h11_state.next_event(),
                 )
 
-            if event is h11.NEED_DATA:
+            if isinstance(event, h11.NEED_DATA):
                 data = await self._network_stream.read(
                     self.READ_NUM_BYTES, timeout=timeout
                 )
@@ -195,11 +197,8 @@ class AsyncHTTP11Connection(AsyncConnectionInterface):
                     raise RemoteProtocolError(msg)
 
                 self._h11_state.receive_data(data)
-            elif event is h11.PAUSED:
-                # TODO: Implement handling for paused
-                ...
             else:
-                return cast(h11.Event, event)
+                return event
 
     async def _response_closed(self) -> None:
         async with self._state_lock:
--- a/httpcore/_async/http11.py
+++ b/httpcore/_async/http11.py
@@ -1,7 +1,16 @@
 import enum
 import time
 from types import TracebackType
-from typing import AsyncIterable, AsyncIterator, List, Optional, Tuple, Type, Union
+from typing import (
+    AsyncIterable,
+    AsyncIterator,
+    List,
+    Optional,
+    Tuple,
+    Type,
+    cast,
+    Union,
+)
 
 import h11
 
@@ -16,15 +25,6 @@ from .._synchronization import AsyncLock
 from .._trace import Trace
 from ..backends.base import AsyncNetworkStream
 from .interfaces import AsyncConnectionInterface
-
-H11Event = Union[
-    h11.Request,
-    h11.Response,
-    h11.InformationalResponse,
-    h11.Data,
-    h11.EndOfMessage,
-    h11.ConnectionClosed,
-]
 
 
 class HTTPConnectionState(enum.IntEnum):
@@ -130,7 +130,7 @@ class AsyncHTTP11Connection(AsyncConnectionInterface):
         await self._send_event(h11.EndOfMessage(), timeout=timeout)
 
     async def _send_event(
-        self, event: H11Event, timeout: Optional[float] = None
+        self, event: h11.Event, timeout: Optional[float] = None
     ) -> None:
         bytes_to_send = self._h11_state.send(event)
         if bytes_to_send is not None:
@@ -168,10 +168,14 @@ class AsyncHTTP11Connection(AsyncConnectionInterface):
             elif isinstance(event, (h11.EndOfMessage, h11.PAUSED)):
                 break
 
-    async def _receive_event(self, timeout: Optional[float] = None) -> H11Event:
+    async def _receive_event(self, timeout: Optional[float] = None) -> h11.Event:
         while True:
             with map_exceptions({h11.RemoteProtocolError: RemoteProtocolError}):
-                event = self._h11_state.next_event()
+                # The h11 type signature uses a private return type
+                event = cast(
+                    Union[h11.Event, h11.NEED_DATA, h11.PAUSED],
+                    self._h11_state.next_event(),
+                )
 
             if event is h11.NEED_DATA:
                 data = await self._network_stream.read(
@@ -191,6 +195,9 @@ class AsyncHTTP11Connection(AsyncConnectionInterface):
                     raise RemoteProtocolError(msg)
 
                 self._h11_state.receive_data(data)
+            elif event is h11.PAUSED:
+                # TODO: Implement handling for paused
+                ...
             else:
                 return event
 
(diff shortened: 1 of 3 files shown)


GalaxySnail:
I think it may not be a good idea to pin h11 under 0.14, it can cause unnecessary conflicts. See: https://hynek.me/articles/semver-will-not-save-you/

madkinsz:
> I think it may not be a good idea to pin h11 under 0.14, it can cause unnecessary conflicts. See: https://hynek.me/articles/semver-will-not-save-you/

That sentiment is the point of issue #509. This pull request is to add compatibility which is necessary whether the pin is removed or not. There are other dependencies with upper versions in this project and that decision should be made separately from this fix.

httpcore/_async/http11.py
```diff
@@ -191,8 +195,11 @@ async def _receive_event(self, timeout: Optional[float] = None) -> H11Event:
                     raise RemoteProtocolError(msg)
 
                 self._h11_state.receive_data(data)
+            elif event is h11.PAUSED:
+                # TODO: Implement handling for paused
```
graingert, line 199: h11.PAUSED is handled in https://github.com/encode/httpcore/pull/526/files#diff-f1329e0e83048035c0a643ab5f3c666ae6565cba6393111fc0dc1afc4f2dbaa4R168

httpcore/_sync/http11.py
```diff
@@ -191,8 +180,11 @@ def _receive_event(self, timeout: Optional[float] = None) -> H11Event:
                     raise RemoteProtocolError(msg)
 
                 self._h11_state.receive_data(data)
+            elif event is h11.PAUSED:
+                # TODO: Implement handling for paused
```
graingert, line 184: paused is handled here https://github.com/encode/httpcore/pull/526/files#diff-7b355550eef6173e16a39fc2fd633b09e9589e707ed9d2792dbb3e205cbd7f0eR155

httpcore/_async/http11.py
```diff
@@ -168,12 +168,18 @@ async def _receive_response_body(self, request: Request) -> AsyncIterator[bytes]
...
             with map_exceptions({h11.RemoteProtocolError: RemoteProtocolError}):
-                event = self._h11_state.next_event()
+                # The h11 type signature uses a private return type
+                event = cast(
+                    Union[h11.Event, h11.NEED_DATA, h11.PAUSED],
+                    self._h11_state.next_event(),
+                )
 
-            if event is h11.NEED_DATA:
+            if isinstance(event, h11.NEED_DATA):
```
graingert, line 182: I worry how much this cast and isinstance will slow stuff down - as this is a hot loop

httpcore/_sync/http11.py
```diff
@@ -127,14 +127,14 @@ def _send_request_body(self, request: Request) -> None:
             event = h11.Data(data=chunk)
             self._send_event(event, timeout=timeout)
 
-        event = h11.EndOfMessage()
-        self._send_event(event, timeout=timeout)
+        self._send_event(h11.EndOfMessage(), timeout=timeout)
 
     def _send_event(
-        self, event: H11Event, timeout: Optional[float] = None
+        self, event: h11.Event, timeout: Optional[float] = None
```
graingert, line 133: I think we still want our send_event to have a specific allowlist of events we send:

```
H11SendEvent = Union[
    h11.Request,
    h11.Data,
    h11.EndOfMessage,
    h11.ConnectionClosed,
]
```

```suggestion
        self, event: H11SendEvent, timeout: Optional[float] = None
```

httpcore/_sync/http11.py
```diff
@@ -127,14 +127,14 @@ def _send_request_body(self, request: Request) -> None:
...
     ) -> None:
         bytes_to_send = self._h11_state.send(event)
-        self._network_stream.write(bytes_to_send, timeout=timeout)
+        if bytes_to_send is not None:
```
graingert, line 136: we know we don't send ConnectionClosed so this check is redundant

[send](https://h11.readthedocs.io/en/latest/api.html#h11.Connection.send) should be defined using [overload](https://docs.python.org/3/library/typing.html#typing.overload)

graingert:
I think this is ready for review now @madkinsz

httpcore/_async/http11.py
graingert: we can hack around this by moving the cast outside the loop: https://github.com/encode/httpcore/blob/627db6cdcbf2d73a9b4c8800cb402b3bd6bf39df/httpcore/_async/http11.py#L30-L49
https://github.com/encode/httpcore/blob/627db6cdcbf2d73a9b4c8800cb402b3bd6bf39df/httpcore/_async/http11.py#L195-L207

simonw:
I'm seeing a weird intermittent bug with https://lite.datasette.io/ which it looks like would be resolved by merging this PR!

```
  File "/lib/python3.10/asyncio/tasks.py", line 232, in __step
    result = coro.send(None)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 310, in add_requirement
    return await self.add_requirement_inner(req)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 391, in add_requirement_inner
    if self.check_version_satisfied(req):
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 337, in check_version_satisfied
    raise ValueError(
ValueError: Requested 'h11<0.13,>=0.11', but h11==0.13.0 is already installed
```
It looks like something deep in the complex world of Pyodide may be installing `h11==0.13` - even though I've tried in my code to ensure that `h11==0.12` is installed first, here: https://github.com/simonw/datasette-lite/blob/d64455c2b7d89a32b8537c63f46db3579f0de91f/webworker.js#L53-L54

I'm going to keep on looking for a workaround, but I thought I'd drop a note here about this issue in case anyone else runs into a similar problem and lands here from a search.

httpcore/_sync/http11.py
madkinsz: 👍 I'll see if I can contribute that upstream

httpcore/_sync/http11.py
madkinsz: 👍 makes sense to me. Seems like we would need this with the [overload](https://github.com/encode/httpcore/pull/526/files#r901136483) for correct typing.

httpcore/_sync/http11.py
madkinsz: Oof, this would actually be a pretty big change upstream as the following is not a valid overload:
```python
    @overload
    def send(self, event: ConnectionClosed) -> None:
        ...

    @overload
    def send(self, event: Event) -> bytes:
        ...

    def send(self, event: Event) -> Optional[bytes]:
```

```
h11/_connection.py:505: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [misc]
```

madkinsz: I'm also exploring a generic

```python
SendType = TypeVar('SendType', None, bytes)

class Event(ABC, Generic[SendType]):
    ...
```

but this would break type-checks for any downstream library that uses `Event` without a value for the generic e.g.
```
h11/_state.py:192: error: Missing type parameters for generic type "Event"  [type-arg]
```

madkinsz:
@graingert I'm not sure the remaining optimizations regarding the overload are feasible in the short term. It may be worth merging the improvements to unblock upgrades and address the overload related checks once we can implement that upstream.

Can we get another maintainer to take a look at the hot-loop cast concerns?

madkinsz:
Note https://github.com/python-hyper/h11/commit/04cc0f781c47ebfb9f9b188a8e8aa423f276c0b1 was merged a week ago so we can remove the aforementioned hot-loop cast once there's another h11 release.

httpcore/_async/http11.py
```diff
@@ -168,12 +168,18 @@ async def _receive_response_body(self, request: Request) -> AsyncIterator[bytes]
...
+        self, timeout: Optional[float] = None
+    ) -> Union[h11.Event, h11.PAUSED]:
         while True:
             with map_exceptions({h11.RemoteProtocolError: RemoteProtocolError}):
-                event = self._h11_state.next_event()
+                # The h11 type signature uses a private return type
+                event = cast(
+                    Union[h11.Event, h11.NEED_DATA, h11.PAUSED],
+                    self._h11_state.next_event(),
+                )
```
Kludex, line 180: So I need to do this on uvicorn as well? :disappointed:

madkinsz: I'm not sure if you do following release of https://github.com/python-hyper/h11/pull/144 which updates that function's signature with narrowed return types.

eseglem:
h11 just released [v0.14.0](https://pypi.org/project/h11/0.14.0/). Does that mean this PR can remove the "hot loop cast" and move forward with another release for httpcore?

I have a server I deal with which behaves poorly and the h11 update will handle its behavior, so I look forward to having it make its way through h11 > httpcore > httpx releases.

madkinsz:
I'll look into updating this for h11 v0.14

tomchristie:
Thanks @madkinsz - should we close this off in favour of #579?

PR #543, opened 2022-04-27
httpcore/backends/trio: map OSError exceptions

vfazio:
The underlying socket implementations can raise OSError for situations
such as `getaddrinfo` failing to resolve an address.

Previously, these errors would not be mapped to httpcore exception types
for the trio backend.

Now, map OSError exceptions in the trio backend to ConnectError.

This will make trio consistent with the asyncio backend.

Signed-off-by: Vincent Fazio <vfazio@xes-inc.com>


vfazio:
@tomchristie @michaeloliverx Just tagging since I saw you two were the last to touch this file. If there are concerns, please let me know.

florimondmanca (approved):
LGTM! Thanks.

Issue #523, opened 2022-03-01
Add types for `extensions` using `typing.TypedDict`


michaeloliverx:
I think it would increase DX if we typed `extensions` using `TypedDict`.

Example:

<img width="280" alt="image" src="https://user-images.githubusercontent.com/55017335/156227311-a2fa842b-1d39-4c67-9e90-13a9d6793051.png">

<img width="280" alt="image" src="https://user-images.githubusercontent.com/55017335/156227461-492aef6a-a223-4e22-a48c-24acf5d1ade7.png">

We would need to add the `typing-extensions` package for versions below 3.8 to use `TypedDict`.

## References
- https://www.python.org/dev/peps/pep-0589/

florimondmanca:
I think this would be a rather nice developer experience addition at a rather low-cost. Would also be an improvement for HTTPCore, by helping prevent using wrong extension name/value types.

vladNed:
working on this

PR #551, opened 2022-06-14
Implementing `TypedDict` class for `request.extensions` 

vladNed:
Implementing a `TypedDict` object for `request.extensions` as in #523. 

For maintaining the optional flags over the extensions the `total` flag was set to false.
I also added the `Dict[str, Any]` for the kwargs as mypy would not recognize the unpacking of kwargs.

### DIFF
--- a/httpcore/_sync/connection.py
+++ b/httpcore/_sync/connection.py
@@ -1,7 +1,7 @@
 import itertools
 import ssl
 from types import TracebackType
-from typing import Iterator, Optional, Type
+from typing import Any, Dict, Iterator, Optional, Type
 
 from .._exceptions import ConnectError, ConnectionNotAvailable, ConnectTimeout
 from .._models import Origin, Request, Response
@@ -99,7 +99,7 @@ class HTTPConnection(ConnectionInterface):
         while True:
             try:
                 if self._uds is None:
-                    kwargs = {
+                    kwargs: Dict[str, Any] = {
                         "host": self._origin.host.decode("ascii"),
                         "port": self._origin.port,
                         "local_address": self._local_address,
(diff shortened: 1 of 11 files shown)


florimondmanca:
Not sure about this…

What’s the behavior for existing transports that provide types in their signatures? Would they now have a type error if they typed extensions as dict, saying « Signature incompatible with parent class »? I strongly worry so.

What would happen if people do turn to this new type, and then want to add extra keys to extensions? Despite total=False, I suspect extra keys would be a typing error.

Is there a way to only tell mypy, « hey, we know these key/values should be here, but don’t be put off by extra keys »?

vladNed:
The first question, 
Can you please provide some examples of that? It will indeed raise typing errors since it's not the same type (I'll research on this one)

The second question:
well yes, if people would add extra keys to extensions then that would raise a typing error. I was assuming that the `extensions` are predefined and you are not able to add extra keys. Checking now the docs for extensions I see that only the `timeout` and `trace` is supported for requests.

I would propose maybe having typed dict for response extensions then. What do you say ?

httpcore/_sync/connection.py
```diff
@@ -99,7 +99,7 @@ def _connect(self, request: Request) -> NetworkStream:
         while True:
             try:
                 if self._uds is None:
-                    kwargs = {
+                    kwargs: Dict[str, Any] = {
```
graingert, line 102: ah this is because kwargs is reassigned in the `else` clause and Trace takes a `dict` when it should take a `collections.abc.Mapping` - for variance support

tomchristie:
> Actually not sure about this anymore…

Yup - personally I'd *probably* prefer to just leave them as simply typed to `dict`.

It's a deliberately loosely defined bit of the API space.

PR #579, opened 2022-09-25
Update types for h11 v0.14

madkinsz:
Extends and supersedes #526. 

Updates the types in `_send_event` and `recieve_event` for compatibility with h11 0.14.0.

Tests and `mypy` were run locally on h11 0.13.0 and 0.14.0.

Closes https://github.com/encode/httpcore/pull/503
Closes https://github.com/encode/httpcore/pull/498
Unblocks https://github.com/encode/httpcore/issues/509

#### Removes cast from the "hot loop"

h11 0.14.0 includes the upstream fix https://github.com/python-hyper/h11/commit/04cc0f781c47ebfb9f9b188a8e8aa423f276c0b1 allowing us to remove the cast and `isinstance` check from the hot loop which @graingert expressed concerns about in https://github.com/encode/httpcore/pull/526#discussion_r901134892.  A cast is still required for the return value as mypy is not type-narrowing on the `is` clause. We'd probably have to do a weird `isinstance/issubclass` combination to avoid the cast, which would bring us back to the original concerns about affected performance.

Interestingly, `mypy` passes with h11 0.13.0 so we may have been able to use this implementation without the upstream change.

#### Addresses narrower types

Updates the `_send_event` types to be narrower as mentioned in https://github.com/encode/httpcore/pull/526#discussion_r901136131. If overloads are added to the h11 library, this will allow us to remove an unnecessary null check as noted in https://github.com/encode/httpcore/pull/526#discussion_r901136483.

#### Bumps minimum h11 version

Bumps the minimum h11 version to 0.13.0 which is required for the `h11.Event` type. I imagine we could create a conditional stub for the type if support for older versions of h11 is important.

### DIFF
--- a/httpcore/_async/http11.py
+++ b/httpcore/_async/http11.py
@@ -8,8 +8,8 @@ from typing import (
     Optional,
     Tuple,
     Type,
-    cast,
     Union,
+    cast,
 )
 
 import h11
@@ -199,7 +199,7 @@ class AsyncHTTP11Connection(AsyncConnectionInterface):
                 # TODO: Implement handling for paused
                 ...
             else:
-                return event
+                return cast(h11.Event, event)
 
     async def _response_closed(self) -> None:
         async with self._state_lock:
--- a/httpcore/_async/http11.py
+++ b/httpcore/_async/http11.py
@@ -26,13 +26,13 @@ from .._trace import Trace
 from ..backends.base import AsyncNetworkStream
 from .interfaces import AsyncConnectionInterface
 
-
 # A subset of `h11.Event` types supported by `_send_event`
 H11SendEvent = Union[
     h11.Request,
     h11.Data,
     h11.EndOfMessage,
 ]
+
 
 class HTTPConnectionState(enum.IntEnum):
     NEW = 0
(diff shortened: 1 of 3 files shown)


madkinsz:
Let me know if you'd like me to rebase or clean this history up. For now, I've just extended the commits from the original branch for clarity.

tomchristie:
> Let me know if you'd like me to rebase to clean this history up.

Not needed, no. We have "squash and merge" enforced on this project, so pull requests always end up as our atomic unit of history.

httpcore/_async/http11.py
```diff
@@ -17,16 +26,14 @@
 from ..backends.base import AsyncNetworkStream
 from .interfaces import AsyncConnectionInterface
 
-H11Event = Union[
+
```
tomchristie, line 29: Looking at the continuous integration I think the listing requires this line to be removed.

The `scripts/lint` tool ought to do this for you automatically, or we can remove it as part of this review, through the GitHub interface.

Here's the commit suggestion...

```suggestion
```

httpcore/_sync/http11.py
```diff
@@ -17,13 +26,12 @@
 from ..backends.base import NetworkStream
 from .interfaces import ConnectionInterface
 
-H11Event = Union[
+
```
tomchristie, line 29: We'd also drop this newline...

```suggestion
```

tomchristie (changes requested):
A linting fix here...

tomchristie (approved):

PR #581, opened 2022-09-29
Support informational responses with HTTP/1.1

tomchristie:
Prompted by https://github.com/encode/httpcore/discussions/572#discussion-4348090

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -179,7 +179,9 @@ The interface provided by the network stream:
 
 This API can be used as the foundation for working with HTTP proxies, WebSocket upgrades, and other advanced use-cases.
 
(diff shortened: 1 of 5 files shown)


tomchristie:
> Just a quick question: should 1xx responses on non-streaming requests return?

Yes, I think that's okay. You'll end up with a 101 response, with body that'll stream the network content.

At the `httpx` layer that'd typically effectively mean you'd get a 101 response, that'd then result in a read timeout waiting for some response content.

But you'd need to have *sent* an upgrade request in the first place, so you'd have to deliberately do odd things, and you'd end up in a reasonable failure state.

Kludex (approved):
:+1:

PR #582, opened 2022-09-29
Add "trailing_headers" extension for HTTP/1.1

tomchristie:
See https://github.com/encode/httpx/issues/1149

Adding support in HTTP/2 should be similar, requiring the `h2.TrailersReceived` event to populate any trailing headers.

https://python-hyper.org/projects/h2/en/stable/api.html#h2.events.TrailersReceived

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -232,9 +232,13 @@ with httpcore.stream("GET", "https://www.example.com") as response:
 
 Trailing headers are a rarely used feature of HTTP, where supplementary headers may be sent at the end of the response data.
 
-The `trailing_headers` response extension is implemented as a list of `(byte, byte)` tuples containing any [trailing headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer#chunked_transfer_encoding_using_a_trailing_header) sent at the end of the response. This list is only populated once the response is complete, and will be empty while streaming the response data.
+The `trailing_headers` response extenstion is implemented as a list of `(byte, byte)` tuples containing any [trailing headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer#chunked_transfer_encoding_using_a_trailing_header) sent at the end of the response. This list is only populated once the response is complete, and will be empty while streaming the response data.
 
 ```python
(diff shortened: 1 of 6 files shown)


docs/extensions.md
```diff
@@ -227,3 +227,19 @@ with httpcore.stream("GET", "https://www.example.com") as response:
     ssl_object = network_stream.get_extra_info("ssl_object")
     print("TLS version", ssl_object.version())
 ```
+
+### `"trailing_headers"`
+
+Trailing headers are a rarely used feature of HTTP, where supplementary headers may be sent at the end of the response data.
+
+The `trailing_headers` response extenstion is implemented as a list of `(byte, byte)` tuples containing any [trailing headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer#chunked_transfer_encoding_using_a_trailing_header) sent at the end of the response. This list is only populated once the response is complete, and will be empty while streaming the response data.
```
Kludex, line 235:
```suggestion
The `trailing_headers` response extension is implemented as a list of `(byte, byte)` tuples containing any [trailing headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer#chunked_transfer_encoding_using_a_trailing_header) sent at the end of the response. This list is only populated once the response is complete, and will be empty while streaming the response data.
```

Kludex (commented):
There's no mention about the [`TE` headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/TE). Is it intended? 🤔

tomchristie:
Ah gotcha... I see that now yes. The presence of a `TE` header on the request [doesn't matter to `h11`](https://h11.readthedocs.io/en/latest/supported-http.html#details-of-our-http-support-for-http-nerds)...

> The other headers in RFC 7230 are TE:, Trailer:, and Via:; h11 also supports these in the sense that it ignores them and that’s really all it should be doing.

But yes, it makes sense to include it in the example.

It'd be really nice to have a real URL we can point at for this one. 🤣

tomchristie:
Right. Needs a `CONNECT` and/or `Upgrade` test case to bump the coverage back to 100%

PR #596, opened 2022-10-16
Bump anyio range to <5.0

Kludex:
AnyIO 4.0 is still not released. Just preparing it.

EDIT: Actually, this can already be merged...


tomchristie (approved):
Sure thing.
Do we have any indication of what API changes we should be expecting here.
Are we confident there won't be any breaking changes for us?

Kludex:
@alex-oleshkevich told me to test against anyio master to be sure it's fine. I'll do it later.

PR #605, opened 2022-11-17
Handle 1xx interim response codes

tomchristie:
This [discussion on httpx](https://github.com/encode/httpx/discussions/2438) indicates that the latest `httpcore` version has introduced a change in behaviour for "100 Continue" status codes.

This highlights that https://github.com/encode/httpcore/pull/581 has *fixed* our behaviour for "101 Switching Protocols" which is a *final* response code. But *breaks* our behaviour for "100 Continue", "102 Processing", and "103 Early Hints" statues codes, which are all *interim* response codes.

Our behaviour for now ought to treat any `1xx` response codes as interim responses, with the exception of "101 Switching Protocols".

- [x] Tests for 100 and 103 status codes.
- [x] Fix behaviour for 100 and 103 status codes.

Note that in this PR I've only added tests for 100 and 103, since that's sufficient.

We could feasibly have additional smarts around how we deal with the interim cases. For example we *could* defer sending request bodies on "Expect: continue" requests, until we see a "100 continue". But that's really a different story.

### DIFF
--- a/httpcore/_async/http11.py
+++ b/httpcore/_async/http11.py
@@ -153,7 +153,12 @@ class AsyncHTTP11Connection(AsyncConnectionInterface):
 
         while True:
             event = await self._receive_event(timeout=timeout)
-            if isinstance(event, (h11.Response, h11.InformationalResponse)):
+            if isinstance(event, h11.Response):
+                break
+            if (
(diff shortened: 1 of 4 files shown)


Kludex (approved):
Got it! :+1:

PR #612, opened 2022-11-17
Enable mypy option `--warn-return-any`

michaeloliverx:


httpcore/backends/trio.py
```diff
@@ -26,7 +26,7 @@ async def read(
         exc_map = {trio.TooSlowError: ReadTimeout, trio.BrokenResourceError: ReadError}
         with map_exceptions(exc_map):
             with trio.fail_after(timeout_or_inf):
-                return await self._stream.receive_some(max_bytes=max_bytes)
+                return await self._stream.receive_some(max_bytes=max_bytes)  # type: ignore[no-any-return]
```
tomchristie, line 29: Can you explain the `ignore` to me here?
Are we able to drop that if we split this across two lines? (A typed assignment, and then the return)

michaeloliverx: Yes we can drop it if we split into two lines. 

We can also drop it if we used `typing.cast` but I would prefer not to use it as it has a runtime cost even though it is insignificant.

Happy for me to split it then?

tomchristie (approved):

PR #614, opened 2022-11-18
Enable mypy option `--disallow-any-generics`

michaeloliverx:


tomchristie (approved):
Seems reasonable, yup.

Issue #514, opened 2022-03-01
Enable Mypy `--strict` internally


michaeloliverx:
I think being compatible with mypy `--strict` benefits end users who are also using it, if https://github.com/encode/httpcore/pull/513 is merged then there isn't much work left to be compatible with strict mode.

See also https://github.com/encode/httpcore/issues/512

michaeloliverx:
What do @encode/maintainers think of this?

adriangb:
Would we do it only for the codebase or also for tests? The benefit of doing it for tests is that you get to asses what the experience would be like for a user in strict mode.

adriangb:
I see you did both in your PR. I'm in favor of it 😁

tomchristie:
Based on #524 I'd probably be somewhat against it - looks like extra noise & fluff from my perspective.

michaeloliverx:
> Based on #524 I'd probably be somewhat against it - looks like extra noise & fluff from my perspective.

I agree its quite a bit of noise.. another option would be to do it incrementally as files are changed?

RA80533:
> Based on #524 I'd probably be somewhat against it - looks like extra noise & fluff from my perspective.

Are there particular changes in the PR that you would rather see addressed separately, @tomchristie?

1. Narrowing `# type: ignore` to a particular class of error
2. Correcting incomplete `dict` and `list` types
3. Using type aliases on values with domain significance (`HeadersAsMapping` versus `dict`)
4. Casting types brought in from the h2 package

2\. is the most important for the purpose of the PR. Each correction means one less error showing up to the user in VS Code.

graingert:
adding typing in anyio found a large number of type level bugs, I'm in favor of this change - so unmarking stale

adriangb:
I think we probably want to turn on `--strict` one file at a time right?

tomchristie:
Yeah that sort of thing.
It'd probably be okay to make the fixes incrementally, without necessarily ratcheting them in using the CI.
We could improve them bit by bit and only hard-enforce it once we get there.

tomchristie:
See https://github.com/encode/httpx/issues/2436 for how I think we should approach this.

PR #618, opened 2022-11-21
Remove global `ignore_missing_imports` mypy option

michaeloliverx:
The mypy [docs](https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker) says the following about using a global `ignore_missing_imports` option:

> We recommend using this approach only as a last resort: it’s equivalent to adding a # type: ignore to all unresolved imports in your codebase.

This change explicitly states which modules to ignore instead.

Refs #514

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,6 +24,7 @@ flake8==3.9.2  # See: https://github.com/PyCQA/flake8/pull/1438
 isort==5.10.1
 importlib-metadata==4.13.0
 mypy==0.981
+types-certifi==2021.10.8.3
 pytest==7.1.3
 pytest-httpbin==2.0.0rc1
 pytest-trio==0.7.0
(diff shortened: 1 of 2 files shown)


tomchristie:
* Do we agree with the `mypy` team's assessment, in this particular context?
* Are there typing packages for `trio`, `certifi`, `h2`, etc... that would resolve any of these?

tomchristie (approved):
Okay, yup.

I think a follow up would be to add `types-certify` to the `requirements.txt`, and drop the `ignore` from here.

See `httpx` https://github.com/encode/httpx/blob/57aa5b08a48c82a6e1a1500e131d94a31128b6f9/requirements.txt#L32

tomchristie:
👍

PR #625, opened 2022-11-25
Allow `ws` and `wss` schemes.

tomchristie:
Allow URLs that use `ws` and `wss` schemes.
 
Refs https://github.com/encode/httpx/issues/304


Kludex (approved):
:+1:

PR #629, opened 2022-11-30
Use type package for `certifi`

michaeloliverx:
Use third party type package for `certify`, `types-certify` from typeshed https://github.com/python/typeshed/tree/main/stubs/certifi.

Refs #618 #514


tomchristie (approved):
Great, yup.

Can we do the same for trio? - https://github.com/python-trio/trio-typing

michaeloliverx:
> Can we do the same for trio? - https://github.com/python-trio/trio-typing

Sure thing, I didn't realise they shipped types.

PR #630, opened 2022-11-30
Use official typing package for `trio`

michaeloliverx:


michaeloliverx:
Refs https://github.com/encode/httpcore/pull/618

httpcore/backends/trio.py
```diff
@@ -118,14 +120,15 @@ async def connect_tcp(
...
         with map_exceptions(exc_map):
             with trio.fail_after(timeout_or_inf):
                 stream: trio.abc.Stream = await trio.open_tcp_stream(
-                    host=host, port=port, **kwargs
+                    host=host,
+                    port=port,
+                    # Trio supports 'local_address' from 0.16.1 onwards.
+                    # We only include the keyword argument if a local_address
+                    # argument has been passed.
+                    local_address=local_address,
```
michaeloliverx, line 131: Had to inline this as mypy could only infer kwargs as `dict[str, str]`, this is the same behaviour as before though pretty sure.

tomchristie: Gotcha.

It is a minor behavioural change in that it means our trio dependancy is now *strictly* >= 0.16.1, whereas before our dependancy only required >= 0.16.1 *if* `local_address` was being used.

But I think we're okay with that... https://pypi.org/project/trio/#history

tomchristie (approved):

httpcore/backends/trio.py
michaeloliverx: Ahh I get it now, we could split the calls if required but probably not needed as 0.16.1 is quite old now.

If we are going forward with the new behaviour shall I fix the comment?

tomchristie: I think we can just drop the comment yup.

PR #632, opened 2022-11-30
Remove unnecessary comprehensions

michaeloliverx:
Use generators instead of comprehensions.


michaeloliverx:
Seen some code clean-up opportunities while working on the typing stuff so will open a few random PRs to group work.

graingert (approved):

PR #637, opened 2022-12-02
Forwarding HTTP proxies should use a fresh connection for every remote host

tomchristie:
Closes https://github.com/encode/httpx/issues/2488


madkinsz (approved):
This seems straight-forward; I presume the previous behavior is always bad per the issue so this is fine as a change without a major version bump?

tomchristie:
> so this is fine as a change without a major version bump?

Okay without the version bump because we're not changing any (external) API.

tomchristie:
Thank you @madkinsz

PR #639, opened 2022-12-07
Lazy import either 'anyio' or 'trio'

### LINKED ISSUE #626: Avoid importing/using `anyio` when `trio` backend is being used.
Somewhat prompted by https://github.com/encode/httpcore/pull/588#issuecomment-1305647618

When using `trio`, we currently use native trio for [the backend implementation](https://github.com/encode/httpcore/blob/master/httpcore/backends/trio.py), but use `anyio` [for some synchronisation primitives](https://github.com/encode/httpcore/blob/master/httpcore/_synchronization.py).

With a bit of careful thinking, and use of `sniffio`, we could re-jig those synchronisation primitives to avoid that, and ensure that...

* When running under asyncio, we use `anyio` for the network backend and synchronisation primitives.
* When running under trio, we use `trio` for the network backend and synchronisation primitives.

This would seem to me to be an improvement:

* We would no longer run into any clashes between `anyio` and `trio` versions.
* We'd be in a better position to move to installations with fewer dependancies. *Potentially: `'pip install httpcore[asyncio]'` and `'pip install httpcore[trio]'`*.

tomchristie:
Closes #626

**example.py**

```python
import trio
import httpcore

async def main():
    async with httpcore.AsyncConnectionPool() as http:
        response = await http.request("GET", "https://www.example.com/")
        print(response)


trio.run(main)
```

```shell
$ pip install httpcore, trio
$ pip uninstall anyio
$ python example.py
<Response [200]>
```

Turns out this *doesn't* fix the #588 conflict, which is still triggered by the usage of `@pytest.mark.anyio` in the test suite. 🤷‍♂️

### DIFF
--- a/httpcore/_synchronization.py
+++ b/httpcore/_synchronization.py
@@ -7,6 +7,11 @@ import sniffio
 from ._exceptions import ExceptionMapping, PoolTimeout, map_exceptions
 
 
+# Our async synchronization primatives use either 'anyio' or 'trio' depending
+# on if they're running under asyncio or trio.
+#
+# We take care to only lazily import whichever of these two we need.
+
 class AsyncLock:
     def __init__(self) -> None:
         self._backend = ""
--- a/httpcore/_synchronization.py
+++ b/httpcore/_synchronization.py
@@ -6,11 +6,11 @@ import sniffio
 
 from ._exceptions import ExceptionMapping, PoolTimeout, map_exceptions
 
-
 # Our async synchronization primatives use either 'anyio' or 'trio' depending
 # on if they're running under asyncio or trio.
 #
 # We take care to only lazily import whichever of these two we need.
+
 
 class AsyncLock:
     def __init__(self) -> None:
@@ -141,6 +141,9 @@ class AsyncSemaphore:
             self._trio_semaphore.release()
         else:
             self._anyio_semaphore.release()
+
+
+# Our thread-based synchronization primitives...
 
 
 class Lock:
--- a/httpcore/_synchronization.py
+++ b/httpcore/_synchronization.py
@@ -2,17 +2,39 @@ import threading
 from types import TracebackType
 from typing import Optional, Type
 
-import anyio
+import sniffio
 
 from ._exceptions import ExceptionMapping, PoolTimeout, map_exceptions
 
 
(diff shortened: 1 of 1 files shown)


httpcore/_synchronization.py
```diff
@@ -2,17 +2,44 @@
...
+        """
+        self._backend = sniffio.current_async_library()
+        if self._backend == "trio":
+            import trio
+
+            self._trio_lock = trio.Lock()
+        else:
+            import anyio
+
+            self._anyio_lock = anyio.Lock()
```
Kludex, line 32: Do we need to differentiate the name, tho?
```suggestion
            import trio

            self._lock = trio.Lock()
        else:
            import anyio

            self._lock = anyio.Lock()
```

httpcore/_synchronization.py
```diff
@@ -2,17 +2,44 @@
...
 
     async def __aenter__(self) -> "AsyncLock":
-        await self._lock.acquire()
+        if not self._backend:
+            self.setup()
+
+        if self._backend == "trio":
+            await self._trio_lock.acquire()
+        else:
+            await self._anyio_lock.acquire()
```
Kludex, line 41: Can't this remain the same?
```suggestion
        await self._lock.acquire()
```

httpcore/_synchronization.py
```diff
@@ -21,32 +48,102 @@ async def __aexit__(
         exc_value: Optional[BaseException] = None,
         traceback: Optional[TracebackType] = None,
     ) -> None:
-        self._lock.release()
+        if self._backend == "trio":
+            self._trio_lock.release()
+        else:
+            self._anyio_lock.release()
```
Kludex, line 54:
```suggestion
        self._lock.release()
```

httpcore/_synchronization.py
tomchristie: I think we'd get awkward type errors if we used the same attribute for both cases. (?)

tomchristie:
@Kludex I'm going with the "yes these need to be separate, because we're being careful and strict about typing". Functionally it *would* work with a bit of overlap here, but I'd suggest let's keep things neatly separated.

Kludex (approved):

PR #641, opened 2022-12-13
Handle HTTP/1.1 half-closed connections gracefully.

### LINKED ISSUE #723: Handle HTTP/1.1 half-closed connections gracefully.
There's an HTTP/1.1 case that can occur where...

* The client starts sending a request.
* The server half-closes the connection.
* The server sends a response, such as an HTTP 413 Content Too Large.

Currently our behaviour here is that we'll see a `WriteError` occur here and never get a response.
A more graceful behaviour is handle this case, and return the 413 response.

Prompted via https://github.com/encode/httpx/discussions/2503.

*A follow up question to this will be... is there an equivalent to this for HTTP/2 streams? But let's only consider that once we've dealt with this as a precursor.*

tomchristie:
Handle cases like HTTP 413, where the request write fails part way through sending, but a properly formed HTTP error response is then sent.

Closes #723

Originally prompted via https://github.com/encode/httpx/discussions/2503

### DIFF
--- a/httpcore/_sync/http11.py
+++ b/httpcore/_sync/http11.py
@@ -18,6 +18,7 @@ from .._exceptions import (
     ConnectionNotAvailable,
     LocalProtocolError,
     RemoteProtocolError,
+    WriteError,
     map_exceptions,
 )
 from .._models import Origin, Request, Response
@@ -76,13 +77,24 @@ class HTTP11Connection(ConnectionInterface):
 
         try:
             kwargs = {"request": request}
-            with Trace("http11.send_request_headers", request, kwargs) as trace:
-                self._send_request_headers(**kwargs)
-            with Trace("http11.send_request_body", request, kwargs) as trace:
-                self._send_request_body(**kwargs)
-            with Trace(
-                "http11.receive_response_headers", request, kwargs
-            ) as trace:
+            try:
+                trace_message = "http11.send_request_headers"
+                with Trace(trace_message, request, kwargs) as trace:
+                    self._send_request_headers(**kwargs)
+
+                trace_message = "http11.send_request_body"
+                with Trace(trace_message, request, kwargs) as trace:
+                    self._send_request_body(**kwargs)
+            except WriteError:
+                # If we get a write error while we're writing the request,
+                # then we supress this error and move on to attempting to
+                # read the response. Servers can sometimes close the request
+                # pre-emptively and then respond with a well formed HTTP
+                # error response.
+                pass
+
+            trace_message = "http11.receive_response_headers"
+            with Trace(trace_message, request, kwargs) as trace:
                 (
                     http_version,
                     status,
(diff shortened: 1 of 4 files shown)


tomchristie:
Test case failure is...

> "ResourceWarning: Async generator 'httpcore._models.ByteStream.__aiter__' was garbage collected before it had been exhausted. Surround its use in 'async with aclosing(...):' to ensure that it gets cleaned up as soon as you're done using it."

Looks like this block...

https://github.com/encode/httpcore/blob/6a97dade2a57283491c7830385d52daccbeaf93b/httpcore/_async/http11.py#L133-L135

Needs an `aclosing` wrapper.

(No, I don't fully understand the generator clean-up stuff here.)

madkinsz:
Very weird. I presume this is the `WriteError` interrupting iteration during `_send_request_body`?

https://docs.python.org/3/library/contextlib.html#contextlib.aclosing

> Significantly, aclosing() supports deterministic cleanup of async generators when they happen to exit early by [break](https://docs.python.org/3/reference/simple_stmts.html#break) or an exception
>
> This pattern ensures that the generator’s async exit code is executed in the same context as its iterations (so that exceptions and context variables work as expected, and the exit code isn’t run after the lifetime of some task it depends on).

This is only available in Python 3.10+ though? You can `try/finally/aclose` I guess? I'm having a hard time creating a little reproduction.

We're not the only ones :) https://github.com/python/cpython/issues/88684#issuecomment-1189563263

tomchristie:
Related: #657 (?)

cdeler:
> "ResourceWarning: Async generator 'httpcore._models.ByteStream.aiter' was garbage collected before it had been exhausted. Surround its use in 'async with aclosing(...):' to ensure that it gets cleaned up as soon as you're done using it."

I copied the test from this pr and see this error as well. Perhaps it's might be something else. Will try to dig into it

httpcore/_sync/http11.py
```diff
@@ -84,10 +85,21 @@ def handle_request(self, request: Request) -> Response:
...
-                self._send_request_body(**kwargs)
+            try:
+                with Trace(
+                    "send_request_headers", logger, request, kwargs
+                ) as trace:
+                    self._send_request_headers(**kwargs)
+                with Trace("send_request_body", logger, request, kwargs) as trace:
+                    self._send_request_body(**kwargs)
+            except WriteError:
+                # If we get a write error while we're writing the request,
```
cdeler, line 96: What happens if server just closes the request, but don't answer (I saw it in some proxies)?

tomchristie: The connection is closed and the client raises an exception...

```python
httpcore.RemoteProtocolError("Server disconnected without sending a response.")
```

I've added a test case to demonstrate this.

cdeler (approved):
Looks good to me

tomchristie:
Ah... Should have added to `CHANGELOG`.
Nevermind, reminder to self and all for future reviews.

PR #643, opened 2022-12-29
Add `retries` option passthrough to SOCKS proxy classes

madkinsz:
Required for https://github.com/encode/httpx/pull/2517

The `SOCKSProxy` and `AsyncSOCKSProxy` classes document a `retries` option but do not accept it.

This adds the parameter and passes it through to the connection pool.


tomchristie (approved):
Yup.

PR #652, opened 2023-01-26
Adjust max streams when remote settings change

dunkmann00:
If the remote server sends a settings update for `max_concurrent_streams`, check and potentially update the max number of streams we will use for a connection.

This is related to https://github.com/encode/httpx/discussions/2416.

### DIFF
--- a/httpcore/_async/http2.py
+++ b/httpcore/_async/http2.py
@@ -88,8 +88,18 @@ class AsyncHTTP2Connection(AsyncConnectionInterface):
                 async with Trace("http2.send_connection_init", request, kwargs):
                     await self._send_connection_init(**kwargs)
                 self._sent_connection_init = True
-                max_streams = self._h2_state.local_settings.max_concurrent_streams
-                self._max_streams_semaphore = AsyncSemaphore(max_streams)
+
+                # Initially start with just 1 until the remote server provides
+                # its max_concurrent_streams value
+                self._max_streams = 1
+
+                local_settings_max_streams = (
+                    self._h2_state.local_settings.max_concurrent_streams
+                )
+                self._max_streams_semaphore = AsyncSemaphore(local_settings_max_streams)
+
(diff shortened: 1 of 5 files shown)


dunkmann00:
Not sure where to put this so I'll just add it as another message. When I tested this against APNS, I was able to see the settings change come in. This is what the change in max concurrent streams specifically looked like:

```
First event:
ChangedSetting(setting=SettingCodes.MAX_CONCURRENT_STREAMS, original_value=None, new_value=1)

Second Event:
ChangedSetting(setting=SettingCodes.MAX_CONCURRENT_STREAMS, original_value=1, new_value=1000)
```

dunkmann00:
Hello @tomchristie, just wanted to reach out to see if you or somebody from the encode team could review this? Thanks!

tomchristie:
Great stuff, thanks! 🎉

The CI reports [a linting failure](https://github.com/encode/httpcore/actions/runs/4019237343/jobs/6905823498) indicating that the async and sync versions don't match. You can resolve this locally by running `scripts/unasync`.

I'd suggest resolving that first.

Then you might want to take a look at https://github.com/encode/httpcore/blob/master/tests/_async/test_http2.py and see if you're able to add some test coverage for this.

dunkmann00:
Okay I think this is good to go! Let me know if there are any other changes you would like me to make.

dunkmann00:
Hey @tomchristie! Just checking in to see what you think of the recent commits with the added tests?

tomchristie:
Would be interesting to see this working together with our latest logging update...

https://www.encode.io/httpcore/logging/

dunkmann00:
Thanks for the merge!

> Would be interesting to see this working together with our latest logging update...

Are you planning on bumping the `httpcore` dep in `httpx`. I'm not in a hurry or anything, but it would be easier to try this out if it was in `httpx`.

Issue #550, opened 2022-06-03
Request errors on overloaded systems may poison the connection pool


vlaci:
It is somewhat related to #502 as we started first observing the below issue in httpcore 0.14.5. Upgrading to 0.14.7 seemingly fixed the issue but we found out that there is still some underlying race condition that can poison the pool, making it unable to serve requests ever again. The issue came up using `httpx==0.22.0 httpcore==0.14.7`, it is not present using `httpx==0.16.1 httpcore==0.12.3`

Here is a script brute forcing the client and reliably triggering the issue on my machine. (The script is written for the sync version, but we are having this issue in production with the async version as well).

<details>
<summary>Repro script</summary>

`PARALLELISM` and `REQUESTS` may need adjustments to reproduce the issue.

```python
import multiprocessing.pool
import socket
import httpx

PARALLELISM = 20
REQUESTS = 200

unavailable_listener = socket.socket()
unavailable_listener.bind(("localhost", 0))
unavailable_listener.listen()

unavailable_url = "http://{}:{}".format(*unavailable_listener.getsockname())

client = httpx.Client(limits=httpx.Limits(max_connections=1, max_keepalive_connections=0))

pool = multiprocessing.pool.ThreadPool(PARALLELISM)

# client.get(...) works well at this point

# Raises httpcore.PoolTimeout as expected because the pool has been exhausted
pool.map(lambda _: client.get(unavailable_url), range(REQUESTS))

# client.get(...) will still raise PoolTimeout even though there supposed to be no outstanding connections
```

</details>

After running the above code, the pool is no longer operational, containing a stuck connection:

```python
>>> client._transport._pool._pool
[<HTTPConnection [CONNECTING]>]
>>> client._transport._pool._pool[0]._connect_failed
False
>>> client._transport._pool._pool[0]._connection
>>> 
>>> client._transport._pool._pool[0].is_avaliable()
>>> False
```

My hunch is that `RequestStatus.set_connection()` will succeed from `_attempt_to_acquire_connection` but `status.wait_for_connection()` can still time-out on an overloaded system leaving the half-created connection there. Subsequent request attempts won't be able to use this connection instance as it isn't `available`, and it won't be cleaned up as it is not `closed` nor `failed`.

The interesting about this situation is that a just crafted connection can be put in the pool without guaranteeing that it will be used during the current requests.

K900:
Not stale.

nihilSup:
[...]

PR #653, opened 2023-01-29
Fixes pool poisoning

nihilSup:
Fixes encode/httpcore#550

Implemented approach with cleaning up zombie connections before creating new connection for both sync and async implementations. Added new status `is_connecting` to `AsyncConnectionInterface` and `ConnectionInterface`. Added tests for pool poisoning and new code.

Other possible solutions:
- https://github.com/encode/httpcore/issues/550#issuecomment-1382257435 prevents stale connections from happening, not sure that it is safe to reuse connection after PoolTimeout in common
- https://github.com/encode/httpcore/issues/550#issuecomment-1384977626 in place clean up, need review

### DIFF
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -185,6 +185,9 @@ class AsyncHTTPConnection(AsyncConnectionInterface):
             return self._connect_failed
         return self._connection.is_closed()
 
+    def is_connecting(self) -> bool:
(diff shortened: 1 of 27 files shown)


tomchristie:
Thanks so much for taking a look at this. 🙏🏼
Apologies for the delay, I am getting around to reviewing this now.

See also https://github.com/encode/httpcore/pull/681

PR #657, opened 2023-02-13
Fix unclosed generator on trio

florimondmanca:
Companion PR for https://github.com/encode/httpx/pull/2587


tomchristie:
Hrm... what I'd like to see in order to progress a pull request here would be a demonstration of what behaviour is breaking.

I'm going to close this off for now, but if anyone's able to narrow this down to a clear problem case then let's review.

PR #668, opened 2023-04-12
Add support for socket options

### LINKED ISSUE #662: Allow socket options to be configured.
We should add a `socket_options` configuration to the `ConnectionPool` and `HTTPProxy` classes.

This would need to be passed through to the network backends `connect_tcp(...)` and `connect_unix_socket()` methods.

For example the sync case, here...

https://github.com/encode/httpcore/blob/21450a75323c715555bf147a941968fa6d662ef5/httpcore/backends/sync.py#L80-L97

...which we would adapt to something like this...

```python
def connect_tcp( 
    self, 
    host: str, 
    port: int, 
    timeout: typing.Optional[float] = None, 
    local_address: typing.Optional[str] = None,
    socket_options: typing.Sequence[typing.Tuple[int, int, typing.Union[int, bytes]]] = ()
 ) -> NetworkStream: 
    address = (host, port) 
    source_address = None if local_address is None else (local_address, 0) 
    exc_map: ExceptionMapping = { 
        socket.timeout: ConnectTimeout, 
        OSError: ConnectError, 
    }
    with map_exceptions(exc_map): 
        sock = socket.create_connection( 
            address, timeout, source_address=source_address 
        )

    # See https://docs.python.org/3/library/socket.html#socket.socket.setsockopt
    for option in socket_options:
        sock.setsockopt(*option)

    return SyncStream(sock) 
```

This would resolve https://github.com/encode/httpcore/pull/651 and https://github.com/encode/httpx/discussions/2635

We *could* also consider changing our defaults for `TCP_NODELAY`, but we should consider that option as a follow-up.

karosis88:
Closes #662

### DIFF
--- a/docs/connection-pools.md
+++ b/docs/connection-pools.md
@@ -79,9 +79,9 @@ The connection pool instance is also the main point of configuration. Let's take
                    a particular address family. Using `local_address="0.0.0.0"` will
                    connect using an `AF_INET` address (IPv4), while using `local_address="::"`
                    will connect using an `AF_INET6` address (IPv6).
-* `socket_options`: Socket options that have to be included in the TCP socket when the connection was established.
 * `uds`: Path to a Unix Domain Socket to use instead of TCP sockets.
 * `network_backend`: A backend instance to use for handling network I/O.
+* `socket_options`: Socket options that have to be included in the TCP socket when the connection was established.
 
 ## Pool lifespans
 
--- a/docs/connection-pools.md
+++ b/docs/connection-pools.md
@@ -79,7 +79,7 @@ The connection pool instance is also the main point of configuration. Let's take
                    a particular address family. Using `local_address="0.0.0.0"` will
                    connect using an `AF_INET` address (IPv4), while using `local_address="::"`
                    will connect using an `AF_INET6` address (IPv6).
-* `socket_options`: Socket options that  have to be included in the TCP socket when the connection was established.
+* `socket_options`: Socket options that have to be included in the TCP socket when the connection was established.
 * `uds`: Path to a Unix Domain Socket to use instead of TCP sockets.
 * `network_backend`: A backend instance to use for handling network I/O.
 
--- a/docs/connection-pools.md
+++ b/docs/connection-pools.md
@@ -79,6 +79,7 @@ The connection pool instance is also the main point of configuration. Let's take
                    a particular address family. Using `local_address="0.0.0.0"` will
                    connect using an `AF_INET` address (IPv4), while using `local_address="::"`
                    will connect using an `AF_INET6` address (IPv6).
+* `socket_options`: Socket options that  have to be included in the TCP socket when the connection was established.
 * `uds`: Path to a Unix Domain Socket to use instead of TCP sockets.
 * `network_backend`: A backend instance to use for handling network I/O.
 
--- a/httpcore/_sync/connection.py
+++ b/httpcore/_sync/connection.py
@@ -1,7 +1,7 @@
 import itertools
 import ssl
 from types import TracebackType
-from typing import Iterator, Optional, Type
+from typing import Iterable, Iterator, Optional, Type
 
 from .._exceptions import ConnectError, ConnectionNotAvailable, ConnectTimeout
 from .._models import Origin, Request, Response
@@ -9,7 +9,7 @@ from .._ssl import default_ssl_context
 from .._synchronization import Lock
 from .._trace import Trace
 from ..backends.sync import SyncBackend
-from ..backends.base import NetworkBackend, NetworkStream
+from ..backends.base import SOCKET_OPTION, NetworkBackend, NetworkStream
 from .http11 import HTTP11Connection
 from .interfaces import ConnectionInterface
 
@@ -34,6 +34,7 @@ class HTTPConnection(ConnectionInterface):
         local_address: Optional[str] = None,
         uds: Optional[str] = None,
         network_backend: Optional[NetworkBackend] = None,
+        socket_options: Optional[Iterable[SOCKET_OPTION]] = None,
     ) -> None:
         self._origin = origin
         self._ssl_context = ssl_context
@@ -50,6 +51,7 @@ class HTTPConnection(ConnectionInterface):
         self._connection: Optional[ConnectionInterface] = None
         self._connect_failed: bool = False
         self._request_lock = Lock()
+        self._socket_options = () if socket_options is None else socket_options
 
     def handle_request(self, request: Request) -> Response:
         if not self.can_handle_request(request.url.origin):
@@ -104,6 +106,7 @@ class HTTPConnection(ConnectionInterface):
                         "port": self._origin.port,
                         "local_address": self._local_address,
                         "timeout": timeout,
+                        "socket_options": self._socket_options,
                     }
                     with Trace(
                         "connection.connect_tcp", request, kwargs
@@ -114,6 +117,7 @@ class HTTPConnection(ConnectionInterface):
                     kwargs = {
                         "path": self._uds,
                         "timeout": timeout,
+                        "socket_options": self._socket_options,
                     }
                     with Trace(
                         "connection.connect_unix_socket", request, kwargs
--- a/httpcore/_sync/connection.py
+++ b/httpcore/_sync/connection.py
@@ -55,7 +55,7 @@ class HTTPConnection(ConnectionInterface):
         self._connection: Optional[ConnectionInterface] = None
         self._connect_failed: bool = False
         self._request_lock = Lock()
-        self._socket_options = () if socket_options is None else socket_options
+        self._socket_options = socket_options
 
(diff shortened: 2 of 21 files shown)


httpcore/backends/sync.py
```diff
@@ -16,6 +16,47 @@
...
 from .base import NetworkBackend, NetworkStream
 
+SOCKET_OPTIONS = typing.Union[
+    typing.Tuple[int, int, int],
+    typing.Tuple[int, int, typing.Union[bytes, bytearray]],
+    typing.Tuple[int, int, None, int],
+]
+
+
+def create_connection(
```
karosis88, line 26: This function is very similar to the built-in `create_connection` function, but it can also accept socket options.
Because there are options that affect the connection itself, we must set socket options before connecting; we cannot set options after connecting.

karosis88:
If we want to support socket options for async interfaces, we need to change a lot of things. Currently, the 'get asynclib' function is used to get a specific backend interface provided by anyio, then the "connect tcp" and "connect unix" asynchronous functions of that module are used to create a connection; how can we override these methods while keeping it as simple as possible?

karosis88:
As a simple solution, we can associate each library with its function and then call that function.

Old
``` python
async def connect_unix(path: Union[str, "PathLike[str]"]) -> UNIXSocketStream:
    path = str(Path(path))
    return await get_asynclib().connect_unix(path)
```
New
``` python
async def connect_unix(path: Union[str, "PathLike[str]"]) -> UNIXSocketStream:
    path = str(Path(path))
    library = sniffio.current_async_library()
    map = {
        "trio": trio_connect_tcp,
        "asyncio": asyncio_connect_tcp,
    }
    return map[library](path)
```

httpcore/backends/sync.py
tomchristie: Okay, I'd suggest we add a comment noting that.

Perhaps rephrased to "This function is equivalent to Python's `socket.create_connection()` function, but also accepts socket options." and linking to the source on GitHub?

tomchristie:
> If we want to support socket options for async interfaces, we need to change a lot of things. Currently, the 'get_asynclib' function is used to get a specific backend interface provided by anyio, then the "connect_tcp" and "connect_unix" asynchronous functions of that module are used to create a connection

It took me a little while to track down what you're referring to here.

> how can we override these methods while keeping it as simple as possible?

We'd need to work this through on a case-by-case basis.

The three cases to cover are `sync`, `trio`, and `asyncio` via `anyio`.

We'd want the PR to start by adapting each of...

* `httpcore/backends/sync.py`
* `httpcore/backends/trio.py`
* `httpcore/backends/asyncio.py`

...and then work through from there.

Let's link to each of these here so it's easier to review how we'd handle this...

https://github.com/encode/httpcore/blob/82466e18bdc4eb7f442cbc52d51deb55b2063273/httpcore/backends/sync.py#L94-L96

https://github.com/encode/httpcore/blob/82466e18bdc4eb7f442cbc52d51deb55b2063273/httpcore/backends/trio.py#L127-L129

https://github.com/encode/httpcore/blob/82466e18bdc4eb7f442cbc52d51deb55b2063273/httpcore/backends/asyncio.py#L111-L115

You've already demonstrated how we'd adapt the sync case.

The `trio` case would require less duplication to resolve if https://github.com/python-trio/trio/issues/281 was addressed, but otherwise [needs their `connect_tcp()` function to be duplicated](https://github.com/python-trio/trio/blob/4f17d2b2693d961f2c12df76a5ce4f80ec3f3178/trio/_highlevel_open_tcp_stream.py#L175-L178).

I think you're pointing out that the asyncio-via-anyio case is actually really awkward to resolve. And yes, that looks correct to me. It does look like duplicating [the anyio `connect_tcp` implementation](https://github.com/agronholm/anyio/blob/master/src/anyio/_core/_sockets.py#L122-L250) would be a bit awkward.

🤔 Parking this for a bit.

tomchristie:
Related, and possibly helpful here - it might be worthwhile taking a look at what `aiohttp` use for their `connect_tcp` implementation.

karosis88:
There is an elegant solution, which is to ignore the socket option, which affects the connect method itself.

Our issue is that the socket bind and connect methods behave differently depending on socket options, such as `socket.SO_REUSEADDR` can help us in connecting to a socket that was previously force closed and is now waiting 1-2 minutes for the socket to be released.

That is very fun, but trio does not support local address binding...

There is also REUSEPORT, which allows you to listen to an IP port pair with multiple sockets, but it is far from httpcore because httpcore is a client library and does not support REUSEPORT at all.

I looked for options that affect bind and connect methods and that httpcore or httpx users can use in their applications, but I couldn't find anything.

I believe it is better to ignore such options and set options after the backend library has been created a connection.
This solution avoids code duplication while keeping code as simple as possible.

This is how the updated function could look.
``` python
class Backend(AsyncNetworkBackend):
    async def connect_tcp(
         ...,
         socket_options
    ):
        ...
        stream = await backend.connect_tcp(
                    remote_host=host,
                    remote_port=port,
                    local_host=local_address,
                )
        for option in socket_options:
             stream._socket.setsockopt(*option)
        return AsyncIOStream(stream)
```

karosis88:
Also, because we don't need options affecting bind or connect, we can use socket.create_connection instead of our newly created one to avoid code duplication

tomchristie:
Okay, setting socket options after the connect is *probably(?)* okay for our use-case.

For the async case I'd suggest we push the *implementation* into `httpcore/backends/trio.py` and 
`httpcore/backends/asyncio.py`, with `httpcore/backends/auto.py` calling through to them.

karosis88:
> We should also take a call on _if we really do want this_, or if what we actually need is just a better default for the sync backed so that it includes `TCP_NODELAY`.
> 
> See [#651 (comment)](https://github.com/encode/httpcore/pull/651#issuecomment-1554579913)

I think we should make TCP_NODELAY the default, but also make socket options configurable; there are many options in unix sockets that developers can use, and this can be a powerful feature for developers **who know** what they're doing.

Additionally, fixing https://github.com/encode/httpx/discussions/2635 requires these configurable sockets. which initially inspired me to open this PR  :D

httpcore/_sync/connection.py
```diff
@@ -54,6 +55,7 @@ def __init__(
         self._connection: Optional[ConnectionInterface] = None
         self._connect_failed: bool = False
         self._request_lock = Lock()
+        self._socket_options = () if socket_options is None else socket_options
```
tomchristie, line 58:
```suggestion
        self._socket_options = socket_options
```

let's keep this optional just now.

httpcore/_sync/connection_pool.py
```diff
@@ -109,6 +110,7 @@ def __init__(
         self._network_backend = (
             SyncBackend() if network_backend is None else network_backend
         )
+        self._socket_options = () if socket_options is None else socket_options
```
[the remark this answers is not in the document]
karosis88: Should we always make sure to check if `socket_options` is not None before using it?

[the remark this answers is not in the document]
tomchristie: It would make sense for us to do the `if socket_options is None` dance at the last possible point, in the network backends.

tests/_sync/test_connection_pool.py
```diff
@@ -187,7 +187,7 @@ def test_debug_request(caplog):
         (
             "httpcore.connection",
             logging.DEBUG,
-            "connect_tcp.started host='example.com' port=80 local_address=None timeout=None socket_options=()",
+            "connect_tcp.started host='example.com' port=80 local_address=None timeout=None socket_options=None",
```
tomchristie, line 190: 👍 yes, this is better.

docs/connection-pools.md
```diff
@@ -79,6 +79,7 @@ The connection pool instance is also the main point of configuration. Let's take
                    a particular address family. Using `local_address="0.0.0.0"` will
                    connect using an `AF_INET` address (IPv4), while using `local_address="::"`
                    will connect using an `AF_INET6` address (IPv6).
+* `socket_options`: Socket options that have to be included in the TCP socket when the connection was established.
```
tomchristie, line 82: let's be consistent with our placement of `socket_options` in the list, so that the docs match the code.

httpcore/backends/sync.py
```diff
@@ -83,34 +83,49 @@ def connect_tcp(
         port: int,
         timeout: typing.Optional[float] = None,
         local_address: typing.Optional[str] = None,
+        socket_options: typing.Optional[typing.Iterable[SOCKET_OPTION]] = None,
     ) -> NetworkStream:
+        if socket_options is None:
+            socket_options = []  # pragma: no cover
```
tomchristie, line 89:
```suggestion
                # Note that we automatically include `TCP_NODELAY`
                # in addition to any other custom socket options.
                if socket_options is None:
            socket_options = []  # pragma: no cover
```

tomchristie (approved):
Looks good, yeah.

I've included some possible comments here. (?)

Issue #670, opened 2023-04-18
ValueError: list.remove(x): x not in list


dimaqq:
Apologies for a not-very-well-formed bug report.

I was only able to hit this under rather weird conditions (streaming responses, backpressure, connection pool full).

```
ERROR:hypercorn.error:Error in ASGI Framework
Traceback (most recent call last):
  File "/home/proxy/.cache/pypoetry/virtualenvs/proxy-9E6wT_fw-py3.10/lib/python3.10/site-packages/httpcore/_async/connection_pool.py", line 227, in handle_async_request
    connection = await status.wait_for_connection(timeout=timeout)
  File "/home/proxy/.cache/pypoetry/virtualenvs/proxy-9E6wT_fw-py3.10/lib/python3.10/site-packages/httpcore/_async/connection_pool.py", line 34, in wait_for_connection
    await self._connection_acquired.wait(timeout=timeout)
  File "/home/proxy/.cache/pypoetry/virtualenvs/proxy-9E6wT_fw-py3.10/lib/python3.10/site-packages/httpcore/_synchronization.py", line 38, in wait
    await self._event.wait()
  File "/home/proxy/.cache/pypoetry/virtualenvs/proxy-9E6wT_fw-py3.10/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 1842, in wait
    if await self._event.wait():
  File "/usr/local/lib/python3.10/asyncio/locks.py", line 214, in wait
    await fut
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/proxy/.cache/pypoetry/virtualenvs/proxy-9E6wT_fw-py3.10/lib/python3.10/site-packages/hypercorn/asyncio/task_group.py", line 23, in _handle
    await app(scope, receive, send, sync_spawn, call_soon)
  File "/home/proxy/.cache/pypoetry/virtualenvs/proxy-9E6wT_fw-py3.10/lib/python3.10/site-packages/hypercorn/app_wrappers.py", line 33, in __call__
    await self.app(scope, receive, send)
  File "/home/proxy/proxy/app.py", line 119, in app
    async with config.upstream_client.stream(
  File "/usr/local/lib/python3.10/contextlib.py", line 199, in __aenter__
    return await anext(self.gen)
[...]

PR #680, opened 2023-05-08
Fix ValueError: list.remove(x): x not in list (#670)

Tunglies:
Fixs ValueError: list.remove(x): x not in list (#670)

It seemed delete item from self._requests but item not exists.


Tunglies (commented):
Review

Tunglies:
> We'd need this change in both the `_sync` and `_async` cases.
> 
> (Our development workflow is to make the changes in the `_async` case and then run `./scripts/unasync`, to ensure the two cases match)

Thanks, I just want know why Test Suite not passed.

tomchristie (approved):

dimaqq:
Thank you!

PR #681, opened 2023-05-10
Fix erronous pool timeout case

tomchristie:
Refs #550.

Fixes an issue where a pool timeout can occur, even though a request has an assigned connection.

It's not clear to me if this is a complete resolution, or if something more substantial (see #653) is actually required.
But *either* way around I think this resolves an edge case for `pool timeout: 0` cases.

Needs a test case.


nihilSup:
@tomchristie can you use two tests from #653?  `test_pool_under_load` and `test_pool_timeout_connection_cleanup` to be more specific. Maybe I could add them myself but I don't know how

tomchristie:
I've revised this in #688.
Added failing test cases first in the first commit, then resolved them in the second.

PR #683, opened 2023-05-11
Raise `ConnectionNotAvailable` instead of `RemoteProtocolError` on request with terminated HTTP/2 connections

zanieb:
Follow-up to https://github.com/encode/httpcore/pull/679#issuecomment-1543672906

Attempts to improve handling of HTTP/2 behavior when connections are terminated.

Refs #730

### DIFF
--- a/httpcore/_async/http2.py
+++ b/httpcore/_async/http2.py
@@ -143,10 +143,14 @@ class AsyncHTTP2Connection(AsyncConnectionInterface):
                 # a GOAWAY event. Other flows of control may then raise
                 # a protocol error at any point they interact with the 'h2_state'.
                 #
-                # In this case we'll have stored the event, and should raise
-                # it as a RemoteProtocolError.
+                # In this case we'll have stored the event, and should raise it.
                 if self._connection_error_event:
-                    raise RemoteProtocolError(self._connection_error_event)
+                    if isinstance(
+                        self._connection_error_event, h2.events.ConnectionTerminated
+                    ):
+                        raise ConnectionNotAvailable(self._connection_error_event)
+                    else:
+                        raise RemoteProtocolError(self._connection_error_event)
                 # If h2 raises a protocol error in some other state then we
                 # must somehow have made a protocol violation.
                 raise LocalProtocolError(exc)  # pragma: nocover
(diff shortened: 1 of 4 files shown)


tests/_async/test_http2.py
```diff
@@ -83,6 +83,9 @@ async def test_http2_connection_closed():
         with pytest.raises(RemoteProtocolError):
             await conn.request("GET", "https://example.com/")
 
+        with pytest.raises(ConnectionNotAvailable):
+            await conn.request("GET", "https://example.com/")
```
madkinsz, line 87: This doesn't seem right as we still raise a `RemoteProtocolError` on the first failure

tomchristie:
If I was going to tackle this I would probably start with just the failing test case(s).

* Changing the behaviour of the existing test, ensuring that `ConnectionNotAvailable` is raised instead of `RemoteProtocolError`.
* Adding a test at the connection pool level - sending 2 requests with a mock connection that always sends GoAway after the first response, and ensuring that both requests succeed. (Because the second gracefully retries)

httpcore/_async/http2.py
```diff
@@ -143,10 +143,14 @@ async def handle_async_request(self, request: Request) -> Response:
...
-                # it as a RemoteProtocolError.
+                # In this case we'll have stored the event, and should raise it.
                 if self._connection_error_event:
-                    raise RemoteProtocolError(self._connection_error_event)
+                    if isinstance(
+                        self._connection_error_event, h2.events.ConnectionTerminated
+                    ):
+                        raise ConnectionNotAvailable(self._connection_error_event)
+                    else:
+                        raise RemoteProtocolError(self._connection_error_event)
```
madkinsz, line 153: I don't love this repeated `isinstance` check. We could move it into a utility or just always treat `_connection_error_event` as `ConnectionNotAvailable` instead of a protocol error.

madkinsz:
Failing coverage checks now. I didn't add a test at the connection pool level yet. There are not HTTP/2 examples there, but it doesn't seem hard to add a mock stream as you suggested. Feel free to push changes.

tomchristie:
I've added issue #730 so that we're tracking this neatly.

tomchristie:
Thanks! Now superseeded by #733

PR #684, opened 2023-05-15
Add HTTP/2 "data_stream" extension

tomchristie:
Refs https://github.com/encode/httpx/issues/1150

- [ ] Does the stream interface here expose the correct API. (Eg. do we want `read(max_bytes=...)`?)
- [ ] Add test case.
- [ ] Add documentation.

### DIFF
--- a/httpcore/_async/http2.py
+++ b/httpcore/_async/http2.py
@@ -187,6 +187,9 @@ class AsyncHTTP2Connection(AsyncConnectionInterface):
     # Sending the request...
 
     async def _send_request_headers(self, request: Request, stream_id: int) -> None:
+        """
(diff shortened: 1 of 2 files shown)


tomchristie:
I'm going to close this off for now...

If there's motivation from anyone to proceed with `HTTP/2` bi-directional streams then we'd want:

* To start with a design discussion.
* A pull request that begins with a test case & docs.
* This implementation is a good reference for a resolution, but we also need to not terminate the outgoing data stream when we send the request. We also probably want to avoid altering the `_recieve_response_body` method in order to support this niche use-case. We can instead have a bit more implementation detail inside the `HTTP2DataStream` class.

PR #687, opened 2023-05-17
Internal refactor of HTTP/2 data stream handling.

tomchristie:
Adds the following internal methods...

* `._send_stream_data(...)`
* `._send_end_stream(...)`
* `._receive_stream_data(...)`

Which are called into by the existing `._send_request_body()` and `_receive_response_body()` methods.

This is an internal refactoring, so doesn't need calling out in our release notes.

Motivation... precursor to adding a `"data_stream"` extension allowing for `read/write` operations onto a bidirectional HTTP/2 stream.


tomchristie:
Inconsistently missing coverage onto https://github.com/encode/httpcore/blob/dc0bca9af9ed3a37926fa592f309bfb55f38ff26/httpcore/_async/http2.py#L286

PR #688, opened 2023-05-17
Fix pool timeout edge-case.

tomchristie:
Second attempt at #681, with tests.

- [x] Add tests.
- [x] Fix implementation.
- [x] Update CHANGELOG.

Refs #550 (Possible that it resolves it, unclear?)


Pliner (approved):

jhaenchen:
Should this be:

```
if self.connection is not None:
```

instead of

```
if self.connection is None:
```

PR #694, opened 2023-05-18
Adding received CONNECT headers to response headers

MadRay:
Adding possibility to analyze received Headers from proxy CONNECT requests on application level.

**Problem**: In some cases transports (proxies f.e.) provides additional data (Request-Id, Node Information, etc).

**Example**:
```
<!-- Proxy Headers -->
HTTP/1.1 200 OK   
X-Request-id: e99629fd-ea06-deb1-d37d-27396af70983  // Our target Data
X-Request-node: maelstorm   // Our target data

<-- Real request to Url -->
HTTP/2 200 
server: nginx
date: Thu, 18 May 2023 18:42:15 GMT
content-type: application/json
content-length: 185
```

**Approach**:
Add intermediate requests headers to final Response.

Very likely this solution isn't the best one. Thanks in advance for any feedback!

### DIFF
--- a/httpcore/_async/http_proxy.py
+++ b/httpcore/_async/http_proxy.py
@@ -321,7 +321,7 @@ class AsyncTunnelHTTPConnection(AsyncConnectionInterface):
         response = await self._connection.handle_async_request(request)
         try:
             # Adding Received Headers from CONNECT
-            if getattr(connect_response, 'headers', None):
+            if getattr(connect_response, "headers", None):
                 response.headers += connect_response.headers
         except Exception:
             pass
--- a/httpcore/_async/http_proxy.py
+++ b/httpcore/_async/http_proxy.py
@@ -317,7 +317,15 @@ class AsyncTunnelHTTPConnection(AsyncConnectionInterface):
                     )
 
                 self._connected = True
-        return await self._connection.handle_async_request(request)
+
+        response = await self._connection.handle_async_request(request)
+        try:
+            # Adding Received Headers from CONNECT
+            if getattr(connect_response, 'headers', None):
+                response.headers += connect_response.headers
+        except Exception:
+            pass
+        return response
 
     def can_handle_request(self, origin: Origin) -> bool:
         return origin == self._remote_origin
--- a/httpcore/_sync/http_proxy.py
+++ b/httpcore/_sync/http_proxy.py
@@ -317,7 +317,15 @@ class TunnelHTTPConnection(ConnectionInterface):
                     )
 
                 self._connected = True
-        return self._connection.handle_request(request)
+
+        response = self._connection.handle_request(request)
+        try:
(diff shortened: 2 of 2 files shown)


MadRay:
Okay, seems like i should come back when I'm done with the tests )

tomchristie:
Ah grand thanks.

In terms of process we should aim to start proposals like this as a discussion rather than a pull request, but all good.

First let's figure out what behaviour we want here...

* Should proxy headers be included directly in the response headers?
* Should proxy headers be included indirectly in the response via a response extension, `proxy_headers`?
* Should proxy headers only be included in the first response on the connection, or on all responses on the connection?
* What behaviour do other clients display here?... what does `curl` output for requests like this? what behaviour does `requests`/`urllib3` have here?

MadRay:
Hi, Tom!

Sorry if I violated any established traditions :) And thanks for your response!

**- Should proxy headers be included directly in the response headers?**
Yes, my vision (based on idea how to deliver this information to application-level in the simplest way) is to concatenate connect headers with response headers. Proxy servers, which inject debug info, should behave like this. But in the wild there are some solutions preferring to send its' headers independently (MiTM problems i suppose).

**- Should proxy headers be included indirectly in the response via a response extension, proxy_headers?**
In my opinion, there is no need to create new response extension. Without going into details on RFC-6648, let's suppose that additional proxy headers are very rare case and it is very unlike that it will intersect with regular response headers.

**- Should proxy headers only be included in the first response on the connection, or on all responses on the connection?**
This is a great question, my current use-case is about opening new tunnel for each request. So, i think it is a great aspect to discuss if this case is acceptable to you.

**- What behavior do other clients display here?... what does curl output for requests like this? what behaviour does requests/urllib3 have here?**
According to my data, almost all python http libraries discard such headers.
Sync libraries based on core http package, which drops proxy headers. For example, requests has[ closed issue on very same problem](https://github.com/psf/requests/issues/3061).
Root of problem is http's _tunnel method - [https://github.com/python/cpython/blob/3.11/Lib/http/client.py#L938](https://github.com/python/cpython/blob/3.11/Lib/http/client.py#L938)
But requests has ability to use custom adapters, so it is possible to write own ProxyConnector (as it was done [here](https://gist.github.com/bpartridge/9c758c5e70222bac6ce6e1db7bb4d8ea))

 
Btw, we use httpx package, which is based on your module.

Anyway, my PR was about signalizing on such rare cases. If it is not a acceptable case for you - i will understand this.

Thanks!

tomchristie:
I'd suggest that if we do want to expose these headers we should probably do so via `response.extensions['proxy_headers']` rather than merging them directly into the response headers.

tomchristie:
Closing this one in it's current form, but I'd be happy to help you work though how to approach it as a response extension.
Let me know if that's something you'd like to follow up on.

PR #699, opened 2023-05-24
Documenting the network backend API

### LINKED ISSUE #695: Expose network backends as part of the public API.
We should add a `Network Backends` section to our documentation, which documents our supported network backends.

The sort of functionality that this would allow for includes...

* Network recording / replay.
* In-depth debug tooling.
* Network mocking.
* Custom network behaviour such as handling non-standard SSL or DNS requirements.

For example...

```python
import httpcore
from httpcore.backends.base import NetworkBackend, NetworkStream
from httpcore.backends.sync import SyncBackend


class RecordingNetworkStream(NetworkStream):
    def __init__(self, record_file, stream):
        self.record_file = record_file
        self.stream = stream

    def read(self, max_bytes, timeout=None):
        data = self.stream.read(max_bytes, timeout=timeout)
        self.record_file.write(data)
        return data

    def write(self, buffer, timeout=None):
        self.stream.write(buffer, timeout=timeout)

    def close(self) -> None:
        self.stream.close()

    def start_tls(
        self,
        ssl_context,
        server_hostname=None,
        timeout=None,
    ):
        self.stream = self.stream.start_tls(
            ssl_context, server_hostname=server_hostname, timeout=timeout
        )
        return self

    def get_extra_info(self, info):
        return self.stream.get_extra_info(info)


class RecordingNetworkBackend(NetworkBackend):
    def __init__(self, record_file):
        self.record_file = record_file
        self.backend = SyncBackend()

    def connect_tcp(
        self,
        host,
        port,
        timeout=None,
        local_address=None,
    ) -> NetworkStream:
        stream = self.backend.connect_tcp(
            host, port, timeout=timeout, local_address=local_address
        )
        return RecordingNetworkStream(self.record_file, stream)

    def connect_unix_socket(self, path, timeout=None) -> NetworkStream:
        stream = self.backend.connect_unix_socket(path, timeout=timeout)
        return RecordingNetworkStream(self.record_file, stream)

    def sleep(self, seconds: float) -> None:
        self.backend.sleep(seconds)


with open("network-recording", "wb") as record_file:
    network_backend = RecordingNetworkBackend(record_file)
    with httpcore.ConnectionPool(network_backend=network_backend) as http:
        http.request("GET", "https://www.example.com/")
```

My expectation is that documenting the network backends as public API would also push us towards a few minor naming/API changes.

Other potential benefits of exposing this API publicly...

* Helping developers understand how `httpcore` is designed.
* Exposes a set of `sync`/`asyncio`/`trio` network primitives, available to other packages that want to support sync+async network operations.

tomchristie:
This pull request adds the network backends as public API.

This will help our users understand the architecture of `httpcore`, allow for more in depth debugging, and allow for complex use cases such as network record/replay or custom DNS handling.

The documentation I'm adding here doesn't follow the existing (internal) class names, but instead takes a docs-first approach to 

**TODO**:

- [x] Add documentation
- [x] Change the naming as described in the API here.
- [ ] ~~Ensure UDS case raises a descriptive warning if not implemented.~~ - Already has a `NotImplementedError`.
- [x] Link from the "network_stream" extension to the relevant part of the backends docs.
- [x] `httpcore.backends` can become private.
- [x] Make top-level `httpcore.<...>` classes available, with conditionals for `async` cases if needed.
- [x] Tests should import from public API wherever possible.
- [x] Add to CHANGELOG

---

**Notes to self**:

* Add docstring'ed API reference.
* Change `start_tls` to in-place? (Would also need a docs update in extensions.md)
* Drop `sleep` from the existing interface.
* Add `AutoBackend`?
* I've had to add `SOCKET_OPTIONS` to the public API, because of the interface here.
* We might want to reconsider a `AsyncIOBackend` now.

Closes #695


tomchristie:
This is ready for review from a docs-first perspective.

docs/network-backends.md
```diff
@@ -1,3 +1,268 @@
...
+
+async def main():
+    network_backend = httpcore.AnyIONetworkBackend()
+    async with httpcore.AsyncConnectionPool(network_backend=network_backend) as http:
+        response = await http.request('GET', 'https://www.example.com')
+
+asyncio.run(main())
+```
+
+The `AnyIONetworkBackend` will work when running under either `asyncio` or `trio`. However, if you're working with async using the [`trio` framework](https://trio.readthedocs.io/en/stable/), then we recommend using the `httpcore.TrioNetworkBackend`.
```
Kludex, line 86: Is there a reason to answer why we recommend that? Because that question pop up in my head.

docs/network-backends.md
```diff
@@ -1,3 +1,268 @@
...
+ssl_context.load_verify_locations(certifi.where())
+
+# A basic HTTP/1.1 request as a plain bytestring.
+request = b''.join([
+    b'GET / HTTP/1.1\r\n',
+    b'Host: www.example.com\r\n',
+    b'Accept: */*\r\n',
+    b'Connection: close\r\n',
+    b'\r\n'
+])
```
karosis88, line 50: Should we use a tuple here?

```suggestion
request = (
    b'GET / HTTP/1.1\r\n'
    b'Host: www.example.com\r\n'
    b'Accept: */*\r\n'
    b'Connection: close\r\n'
    b'\r\n'
)
```

docs/network-backends.md
```diff
@@ -1,3 +1,268 @@
...
+                (b"content-type", b"plain/text"),
+            ]
+        ),
+        flags=["END_HEADERS"],
+    ).serialize(),
+    hyperframe.frame.DataFrame(
+        stream_id=1, data=b"Hello, world!", flags=["END_STREAM"]
+    ).serialize(),
+]
+# Note that we instantie the mock backend with an `http2=True` argument.
```
florimondmanca, line 147: Woops

```suggestion
# Note that we instantiate the mock backend with an `http2=True` argument.
```

docs/network-backends.md
Kludex: No, the \r\n is supposed to be in the beginning of the join

docs/network-backends.md
tomchristie: Thanks. Added some text to describe this more clearly.

docs/network-backends.md
tomchristie: > No, the \r\n is supposed to be in the beginning of the join

True, that'd be neater. I've done that now.

tomchristie:
I've now updated the API in this pull request, in line with the docs so it's possible to properly work through the examples...

Documentation: https://github.com/encode/httpcore/blob/network-backends-api/docs/network-backends.md

Install from this branch in order to work through...

```python
$ python -m venv venv
$ venv/bin/pip install git+https://github.com/encode/httpcore.git@network-backends-api
$ venv/bin/python
```

tomchristie:
I've bumped a bunch of items into possible follow-ons.

karosis88 (approved):

tomchristie:
I have bumped out the following conversations on the back of this...

* https://github.com/encode/httpcore/discussions/712
* https://github.com/encode/httpcore/discussions/711
* https://github.com/encode/httpcore/discussions/710

tomchristie:
Okay, let's have it.

Thanks all for the review time on this. 😊

Issue #658, opened 2023-02-21
asyncio.exceptions.CancelledError not releasing connection from the connection pool


niraj8989:
We have an application that is downloading file from third party server.
While downloading if anything happens with the process we are getting CancelledError but it is not releasing the connection from the connection pool.

 File "test/filesystem/cxfiles.py", line 304, in _download_file_once
    async for chunk in self.cxfiles_client.download_service_request_attachment(
  File "test/filesystem/cxfiles_client.py", line 391, in download_service_request_attachment
    yield chunk
  File "/usr/lib/python3.8/contextlib.py", line 189, in __aexit__
    await self.gen.athrow(typ, value, traceback)
  File "/home/niraj/candela/candela_nx/src/test/filesystem/cxfiles_client.py", line 285, in stream_post
    yield response
  File "/usr/lib/python3.8/contextlib.py", line 189, in __aexit__
    await self.gen.athrow(typ, value, traceback)
  File "/home/niraj/venv/lib/python3.8/site-packages/httpx/_client.py", line 1579, in stream
    await response.aclose()
  File "/home/niraj/venv/lib/python3.8/site-packages/httpx/_models.py", line 993, in aclose
    await self.stream.aclose()
  File "/home/niraj/venv/lib/python3.8/site-packages/httpx/_client.py", line 152, in aclose
    await self._stream.aclose()
  File "/home/niraj/venv/lib/python3.8/site-packages/httpx/_transports/default.py", line 244, in aclose
    await self._httpcore_stream.aclose()  # type: ignore
  File "/home/niraj/venv/lib/python3.8/site-packages/httpcore/_async/connection_pool.py", line 356, in aclose
    await self._pool.response_closed(self._status)
  File "/home/niraj/venv/lib/python3.8/site-packages/httpcore/_async/connection_pool.py", line 277, in response_closed
    async with self._pool_lock:
  File "/home/niraj/venv/lib/python3.8/site-packages/httpcore/_synchronization.py", line 15, in __aenter__
    await self._lock.acquire()
  File "/home/niraj/venv/lib/python3.8/site-packages/anyio/_core/_synchronization.py", line 121, in acquire
    await checkpoint_if_cancelled()
  File "/home/niraj/venv/lib/python3.8/site-packages/anyio/lowlevel.py", line 42, in checkpoint_if_cancelled
    await get_asynclib().checkpoint_if_cancelled()
[...]

PR #704, opened 2023-06-01
Fix response clean up

### LINKED ISSUE #642: Support async cancellations.
Over at @anthropics we're keen users of httpx on Trio, and I've noticed that we sometimes have issues where cancelling a task doesn't get all the teardown logic right.  For example, if you try to `await` (or `async for/with`) in an `except BaseException:` or `finally:` block, Trio will immediately re-raise a `Cancelled` exception instead of running your teardown logic.

This can be pretty subtle, since it's happening as you abandon a task anyway, but can cause various subtle or unsubtle problems that motivated us to build [flake8-trio](https://pypi.org/project/flake8-trio/), including the `TRIO102` error for this case.  I'm pretty sure that httpcore has multiple such issues, e.g. 

https://github.com/encode/httpcore/blob/f0657cb43cb707d1672b93b61bb53b4cfb166820/httpcore/_async/connection_pool.py#L251-L253

doesn't look cancel-safe to me.  Do you consider "this doesn't close the connection properly when cancelled" a bug?  If so, please consider this a fairly general bug-report!

(I've also found it infeasible to *consistently* get this right without tool support.  If you want to try flake8-trio many of the checks are framework-agnostic; and if it'd be useful for httpcore and httpx we'd be happy to add `anyio` support to the linter 🙂)

karosis88:
closes #642 and #658


karosis88:
## Our problem

httpcore/_async/http11.py
```python
async def _response_closed(self) -> None:
    async with self._state_lock:
        if (
            self._h11_state.our_state is h11.DONE
            and self._h11_state.their_state is h11.DONE
        ):
            self._state = HTTPConnectionState.IDLE
            self._h11_state.start_next_cycle()
            if self._keepalive_expiry is not None:
                now = time.monotonic()
                self._expire_at = now + self._keepalive_expiry
        else:
            await self.aclose()
```
Let's cancel this coroutine in the middle of its execution.

Oops, now we have a problem: our connection is in the ACTIVE state when it should be in the IDLE or CLOSED state, and our file descriptor was not closed because `Connection.aclose` was not called.

But wait, we were smart enough and put finally block for those cases, now the finally block should be executed in `ConnectionPool.aclose`

```python
    try:
        if hasattr(self._stream, "aclose"):
            await self._stream.aclose()
    finally:
        await self._pool.response_closed(self._status)
```

Unfortunately, our `response_closed` function will not work as expected because our `nursery` object has now closed all tasks.

Simple example
``` python
import trio

async def checkpoint():
    await trio.sleep(0)

async def clean_up():
    await checkpoint()  # in our case: async with self._pool_lock:
    print('done')

async def main():
    with trio.move_on_after(1):
        try:
            await trio.sleep(1.1)
        finally:
            await clean_up()
trio.run(main)
```

In this case, the word 'done' will never be printed.
We can fix it by wrapping "await clean_up" in "CancelScope(shield=True)" to prevent cancellation from top-level scopes.

This code will function properly.
```python
async def main():
    with trio.move_on_after(1):
        try:
            await trio.sleep(1.1)
        finally:
            with trio.CancelScope(shield=True):
                await clean_up()
```

The same solution was used to solve our issue; I simply wrapped the clean up function in "CancelScope(shielf=True)"

karosis88:
## What should we pay attention to?

If the clean up function, which we call in the cancel isolated environment, hangs, that's bad; nursery also hangs forever, so we need to avoid that.

### How?

- timeout: We can use timeout to ensure that our clean up function does not run forever.
- do nothing: We are confident that our cleaning function is functioning properly, so please do nothing.

I chose the second option in this solution because we know our clean up function will work as expected. 

### What are the guarantees?
Let's take a look at our cleanup function, which consists of three lines. 

```python
if cancelled and hasattr(self._stream, "_connection"):  # pragma: no cover
    await self._stream._connection.aclose()

await self._pool.response_closed(self._status)
```

Because we know that synchronous "if" statements do not block, let's look at `await self._stream._connection.aclose()` and `await self._pool.response_closed(self._status)`

What `await self._stream._connection.aclose()` does?

```python
        # Note that this method unilaterally closes the connection, and does
        # not have any kind of locking in place around it.
        self._state = HTTPConnectionState.CLOSED
        await self._network_stream.aclose()
```

So it simply sets the connection state to CLOSED and attempts to close the network stream.

According to https://trio.readthedocs.io/en/stable/reference-io.html?highlight=AsyncResource#trio.abc.AsyncResource, we can be certain that `aclose()` will not block execution indefinitely.

What `await self._pool.response_closed(self._status)` does?
This function is also safe because it is awaiting a lock as well as "stream.aclose()," which cannot hang, so it is safe to wait for it.

tomchristie:
I'm starting to understand this more clearly, thanks. Let's get this worked through. 💪
Okay, we want to shield certain parts of our cleanup from cancellation.

Rather than adding API to the network backends interface, I think we want to treat this in the same kind of way [that we handle async locking and semaphores](https://github.com/encode/httpcore/blob/master/httpcore/_synchronization.py).

Currently we have the following primitives available to us there...

* `AsyncLock`
* `AsyncEvent`
* `AsyncSemaphore`

As well as the corresponding...

* `Lock`
* `Event`
* `Semaphore`

It seems to me that we ought to mirror the `with trio.CancelScope(shield=True):` style, by adding (something like) `AsyncShield` and `Shield` primitives there. (Rather than API on the network backend)

Does that seem like a reasonable suggestion to you?

karosis88:
Is it necessary to create a `Shield` class, or can we simply add another subsitution to `unasync.py` that simply removes AsyncShield?
For example, instead of replacing "@pytest.mark.anyio" with "@pytest.mark.sync", we simply remove it.

karosis88:
I agree with you; I believe it is more intuitive to combine this with synchronizations.

karosis88:
We can simply encapsulate the cleanup logic in the special function "X" that should be marked with the decorator, and then write a regex to remove that decorator. 

possible regex
```python
(r"\s*AsyncShield", '')
```

tomchristie:
Are we able to resolve this with a decorator syntax, or do we need a context-manager syntax?

karosis88:
I am offer to use decorator, because it's easy to remove with the regex, which we can not say about context managers.
As a result, the distinction between async and sync interfaces is reduced to a single line.

tomchristie:
Okay, so supposing we have a `@shield_cancellation` decorator available to us - what changes to `_async/connection_pool.py` would be required?

tomchristie:
(I'm not sure this is viable, since the function call would need to be async, which makes it a potential switchpoint?)

karosis88:
> (I'm not sure this is viable, since the function call would need to be async, which makes it a potential switchpoint?)

Is it enough to be a switchpoint? Can't we rely on that function?

karosis88:
>How?
>- timeout: We can use timeout to ensure that our clean up function does not run forever.
>- do nothing: We are confident that our cleaning function is functioning properly, so please do nothing.

I misunderstood how 'aclose' for 'trio.abc.AsyncResource' works; we should most likely use timeouts.

tomchristie:
> Is it enough to be a switchpoint? Can't we rely on that function?

I don't *think?* we can take that approach, because...

https://trio.readthedocs.io/en/stable/reference-core.html#checkpoints

"Third-party async functions / iterators / context managers can act as checkpoints"

So if we were calling into a decorated `@shield_cancellation` method, using `await my_method()` then we might get cancelled before the shielding decorator takes effect.

My go-to with something like this would typically be "trio gets this just right. how would this look if we're using trio" and to work from there.

karosis88:
I think I'm missing something; isn't this decorator safe?

```python
    @staticmethod
    def shield_cancellation(
        fnc: typing.Callable[..., typing.Awaitable[_R]]
    ) -> typing.Callable[..., typing.Awaitable[_R]]:
        # Makes an async function that runs in a cancellation-isolated environment.

        @wraps(fnc)
        async def inner(*args: typing.Any, **kwargs: typing.Any) -> _R:
            with anyio.CancelScope(shield=True):
                return await fnc(*args, **kwargs)

        return inner
```

httpcore/_async/connection_pool.py
```diff
@@ -11,6 +12,8 @@
 from .connection import AsyncHTTPConnection
 from .interfaces import AsyncConnectionInterface, AsyncRequestInterface
 
+GRACEFUL_CLOSE_TIMEOUT = 2
```
karosis88, line 15: Instead of two, we should probably think about which timeout to use here.

karosis88:
> I am offer to use decorator, because it's easy to remove with the regex, which we can not say about context managers. As a result, the distinction between async and sync interfaces is reduced to a single line.

Let us just avoid removing the decorator from the synchronous interface because it does not interfrare us.

tomchristie:
*manager hat on*

Okay, lovely stuff tho let's slow this down...

You've given a great example [over here](https://github.com/encode/httpx/discussions/2742#discussioncomment-6138483). I think that's the first time I've seen a nice simple "look, we're not supporting cancellation at the moment".

My suggestion would be...

* The title of the PR as "Support async cancellations". That's more clear about exactly what issue we're resolving.
* The description of the PR to include a nice little example like the one you've given for trio.
* We start with a PR that *just* includes a failing test case, before we look at the resolution.

karosis88:
> We start with a PR that just includes a failing test case, before we look at the resolution.

Do we need to open another PR and close this one?

tomchristie:
Up to you.

karosis88:
Was re-opened in #719

PR #714, opened 2023-06-09
Add support for HTTPS Proxy

karosis88:
Add support for **HTTPS** proxies

Refs
  - HTTPS Proxy support: https://github.com/encode/httpx/issues/1434

PR #716, opened 2023-06-10
Add proxy mode

karosis88:
Refs: #714


karosis88:
It's up to you, but I'd suggest adding this feature now to support HTTPS forwarding, as most proxy servers do.

#714 appears to be a lot of work; is it worthwhile to wait for it?

PR #717, opened 2023-06-10
Add exceptions for invalid trace callback

karosis88:
# Examples

```python
import httpcore

async def log(a, b):
    print(a, b)

httpcore.request("GET", "https://google.com", extensions={"trace": log})
```

## Old output
```
RuntimeWarning: coroutine 'log' was never awaited
  self.trace_extension(prefix_and_name, info)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

## New output
```
TypeError: If you are using a synchronous interface, the callback of the `trace` extension should be a normal function instead of an asynchronous function.
```

```python
import trio

def log(a, b):
    print(a,b)

async def req():
    cl = httpcore.AsyncConnectionPool()
    await cl.request("GET", "https://google.com", extensions={"trace": log})

trio.run(req)
```

## Old output
```
TypeError: object NoneType can't be used in 'await' expression
```

## New output
```
TypeError: If you're using an asynchronous interface, the callback of the `trace` extension should be an asynchronous function rather than a normal function.
```

### DIFF
--- a/httpcore/_trace.py
+++ b/httpcore/_trace.py
@@ -1,9 +1,10 @@
+import inspect
 import logging
 from types import TracebackType
 from typing import Any, Dict, Optional, Type
 
 from ._models import Request
-import inspect
+
 
 class Trace:
     def __init__(
@@ -27,9 +28,11 @@ class Trace:
     def trace(self, name: str, info: Dict[str, Any]) -> None:
         if self.trace_extension is not None:
             if inspect.iscoroutinefunction(self.trace_extension):
-                raise TypeError("If you are using a synchronous interface, "
-                                "the callback of the `trace` extension should "
-                                "be a normal function instead of an asynchronous function.")
+                raise TypeError(
+                    "If you are using a synchronous interface, "
+                    "the callback of the `trace` extension should "
+                    "be a normal function instead of an asynchronous function."
+                )
             prefix_and_name = f"{self.prefix}.{name}"
             self.trace_extension(prefix_and_name, info)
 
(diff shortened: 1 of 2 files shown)


karosis88:
When migrating from synchronous to asynchronous interface, it's common to forget to change the synchronous trace callback to an asynchronous function, so let's add a user-friendly behavior for those cases.

httpcore/_trace.py
```diff
@@ -26,6 +27,12 @@ def __init__(
 
     def trace(self, name: str, info: Dict[str, Any]) -> None:
         if self.trace_extension is not None:
+            if inspect.iscoroutinefunction(self.trace_extension):  # pragma: no cover
```
madkinsz, line 30: It would probably be better to check the return value of the function and see if it's a coroutine  — it's generally more robust to use of decorators and avoids false negatives.

karosis88: Thank you for reviewing, but what if it's not a function at all? Then we can't check the return value because an exception would be thrown; do we need to wrap it in try-except? Isn't it too complicated?

madkinsz: If it's not a function then it'll fail with a type error when it's called which seems pretty reasonable. It seems like the goal here is to throw a helpful message and avoid hidden unawaited coroutines. In other cases, like not passing a callable or a callable that takes the wrong arguments, it seems fine to just let the standard errors raise.

karosis88: Yes, it appears I went too far; thanks for the review.🚀

madkinsz: 👍 thanks for contributing! I think this seems helpful fwiw and checking the return value rather than using inspect is kind of a nit. I've just had difficulties with decorated asynchronous functions in the past and we're less likely to introduce regressions with the alternative approach.

tomchristie (approved):
Sure.

PR #719, opened 2023-06-12
Support async cancellations

karosis88:
# Problem

Cancellations in the middle of a request/response cycle are not supported by the httpcore.

Something like this should brake the connection pool

```python
import anyio
import httpcore

async def main():
    pool = httpcore.AsyncConnectionPool()
    with anyio.move_on_after(1):
        await pool.request(
            "GET", "http://example.com"
        )  # long running request
    print(pool.connections)  # One active connection that will remain active indefinitely
    # [<AsyncHTTPConnection ['http://example.com:80', HTTP/1.1, ACTIVE, Request Count: 1]>]

anyio.run(main)
```

# TODO

- [x] Write a failing test
- [ ] Make a test pass


tests/test_cancecllations.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@pytest.mark.anyio
+async def test_async_cancellation():
+    pool = httpcore.AsyncConnectionPool(network_backend=SlowBackend())
+    with anyio.move_on_after(1):
+        await pool.request(
+            "GET", "http://example.com"
+        )
+    assert not pool.connections
```
tomchristie, line 42: I don't think this is quite correct.
The behaviour we'd want to see here is that there *is* a connection in the pool, but it is IDLE.

tests/test_cancecllations.py
```diff
@@ -0,0 +1,42 @@
...
+        timeout: Optional[float] = None,
+        local_address: Optional[str] = None,
+        socket_options: typing.Optional[typing.Iterable[SOCKET_OPTION]] = None,
+    ) -> AsyncNetworkStream:
+        return SlowStream()
+
+
+@pytest.mark.anyio
+async def test_async_cancellation():
+    pool = httpcore.AsyncConnectionPool(network_backend=SlowBackend())
```
tomchristie, line 37: Let's go with the neater usage of `async with ... as pool:`.
That's what we use elsewhere in the tests.

tests/test_cancecllations.py
```diff
@@ -0,0 +1,42 @@
...
+    with anyio.move_on_after(1):
```
tomchristie, line 38: A nice quick timeout in our test cases would be good.
Is 0.001 okay here?

tests/test_cancecllations.py
```diff
@@ -0,0 +1,42 @@
+import typing
+from typing import Optional
+
+import anyio
+import pytest
+
+import httpcore
+from httpcore.backends.auto import AutoBackend
+from httpcore.backends.base import SOCKET_OPTION, AsyncNetworkBackend
+from httpcore.backends.mock import AsyncNetworkStream
```
tomchristie, line 10: Could we just use a top-level `import httpcore` instead?

tests/test_cancecllations.py
```diff
@@ -0,0 +1,42 @@
+import typing
```
Pliner, line 1: There is a typo in the filename (test_cancecllations -> test_cancellations).

tests/test_cancecllations.py
karosis88: I thought about it and can't remember when, but I remember thinking that this is much safer.

karosis88:
First, we must decide how we will support "CancelScope" while keeping sync and async interfaces similar. I'm not sure whether it's better to do it in another PR or in this one.

tomchristie:
> First, we must decide how we will support "CancelScope" while keeping sync and async interfaces similar.

Okay. Before we get to implementation, could you start off by showing me *where* we need shielding rather than *how* we're going to implement it?

In order to have confidence that we're *really getting it right* I'd be interested in looking at that from the ground up...

Suppose we're just working directly with an `HTTPConnection` instance directly, rather than with a connection pool. Where would we need cancellation shielding?

karosis88:
https://github.com/encode/httpcore/blob/eb5957d60b240a0deec503a4acca0e69da2833db/httpcore/_async/http11.py#L220-L232

This function appears to be the main source of the problem; the issue here is that if we call this method while cancelled, the resources will not be closed.

tomchristie:
Perhaps we should add a test case at that level of the API?

karosis88:
Sounds good

tests/test_cancellations.py
```diff
@@ -35,3 +38,39 @@ async def test_async_cancellation():
...
+    async with httpcore.AsyncHTTP11Connection(origin, stream) as conn:
+        # mock
+        conn._state = HTTP11ConnectionState.ACTIVE
+
+        with anyio.CancelScope() as cancel_scope:
+            async with conn._state_lock:
+                cancel_scope.cancel()
+                await conn._response_closed()
+        assert conn._state != HTTP11ConnectionState.ACTIVE
+        assert conn._state in (HTTP11ConnectionState.IDLE, HTTP11ConnectionState.CLOSED)
```
tomchristie, line 56: Okay so, challenge...

Tests ought to be against our public API. In this case, using `conn.request(...)` with a cancellation.

tests/test_cancellations.py
```diff
@@ -0,0 +1,70 @@
...
+@pytest.mark.anyio
+async def test_h11_response_closed():
+    origin = httpcore.Origin(b"https", b"example.com", 443)
+    stream = httpcore.AsyncMockStream([])
+    async with httpcore.AsyncHTTP11Connection(origin, stream) as conn:
+        # mock
+        with anyio.CancelScope() as cancel_scope:
+            cancel_scope.cancel()
+            await conn.request("GET", "https://example.com")
+        assert conn._state == HTTP11ConnectionState.CLOSED
```
tomchristie, line 52: I assume we want the test to mirror the same style as the `test_async_cancellation` case above.

Something like this?...

```suggestion
    stream = SlowStream()
    async with httpcore.AsyncHTTP11Connection(origin, stream) as conn:
        with anyio.move_on_after(0.001):
            await pool.request("GET", "http://example.com")
        assert conn.is_closed()
```

karosis88: This looks much better, thanks

tests/test_cancellations.py
```diff
@@ -0,0 +1,57 @@
...
+import pytest
+
+import httpcore
+
+
+class SlowStream(httpcore.AsyncNetworkStream):
+    async def write(
+        self, buffer: bytes, timeout: typing.Optional[float] = None
+    ) -> None:
+        await httpcore._backends.auto.AutoBackend().sleep(2)
```
tomchristie, line 14:
```suggestion
        await anyio.sleep(2)
```

tests/test_cancellations.py
```diff
@@ -0,0 +1,57 @@
+import typing
+from typing import Optional
```
tomchristie, line 2: Let's just use the top level `import typing` style throughout.

tests/test_cancellations.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@pytest.mark.anyio
+async def test_h2_response_closed():
+    origin = httpcore.Origin(b"http", b"example.com", 80)
+    stream = SlowStream()
+    async with httpcore.AsyncHTTP2Connection(origin, stream) as conn:
+        with anyio.move_on_after(0.001):
+            await conn.request("GET", "http://example.com")
+        assert conn.is_closed()
```
tomchristie, line 57: Okay these are looking great - good starting point for us working through this really comprehensively.

karosis88: If we're done with failing tests, I believe we should decide how and where we want to see shield logic.

tomchristie:
Starting with the HTTP/1.1 case it looks to me like we want...

* Two test cases - the existing one that we currently have for a timeout during writing, and an additional one for a timeout during reading.
* Shielding semantics around our exception handling in the `handle_async_request` method...

```python
        except BaseException as exc:
            with ShieldFromCancellation():
                async with Trace("response_closed", logger, request) as trace:
                    await self._response_closed()
            raise exc
```

* Shielding semantics around our exception handling in the byte stream `__aiter__` method...

```python
        except BaseException as exc:
            # If we get an exception while streaming the response,
            with ShieldFromCancellation():
                await self.aclose()
            raise exc
```

I'm not presuming how we implement the `ShieldFromCancellation` here, just presenting what I think the semantics ought to look like. (Based on `trio`'s design in this problem space. Because *structured concurrency FTW*.)

karosis88:
To me, it appears to be completely correct.

karosis88:
> Two test cases - the existing one that we currently have for a timeout during writing, and an additional one for a timeout during reading.

Are you referring to slow-write test for `ConnectionPool` or `HTTP11Connection`

tomchristie:
> Are you referring to slow-write test for ConnectionPool or HTTP11Connection?

I'm suggesting that we'll 


```python
class SlowReadStream(httpcore.AsyncNetworkStream):
    def __init__(self, buffer):
        self.buffer = buffer

    async def write(self, buffer, timeout = None):
        pass

    async def read(self, max_bytes: int, timeout: typing.Optional[float] = None) -> bytes:
        if not self._buffer:
            await trio.sleep(1)
        else:
            return self._buffer.pop(0)

    async def aclose(self):
        pass

stream = SlowReadStream([
    b"HTTP/1.1 200 OK\r\n",
    b"Content-Type: plain/text\r\n",
    b"Content-Length: 1000\r\n",
    b"\r\n",
    b"Hello, world!...", 
    # The stream will hang after this portion of the incomplete response,
    # so that we're able to test our async cancellation semantics.
])
```

karosis88:
Is a buffer really necessary?

tomchristie:
> Is a buffer really necessary?

You'd need this or something *like* this in order to test the cancellation handling for `HTTP11ConnectionByteStream.__aiter__`.

tomchristie:
Okay, want to try...

* Adding the `ShieldCancellation` from https://github.com/encode/httpcore/pull/726
* Making the first HTTP/1.1 test pass.
* Making the second HTTP/1.1 test pass.

tests/test_cancellations.py
```diff
@@ -0,0 +1,95 @@
...
+            b"Content-Length: 1000\r\n",
+            b"\r\n",
+            b"Hello, world!...",
+        ]
+    )
+    async with httpcore.AsyncHTTP11Connection(origin, stream) as conn:
+        with anyio.move_on_after(0.001):
+            async with conn.stream("GET", "http://example.com") as resp:
+                async for chunk in resp.aiter_stream():
+                    pass
```
tomchristie, line 94: Just `conn.request(...)` would be sufficient here.

(We'll still end up calling into `HTTP11ConnectionByteStream.__aiter__`)

tomchristie:
Thanks so much for your work on this.
I've started pulling it into #726, so I think we can close this one as being superseeded at this point.

PR #726, opened 2023-06-15
Support async cancellations.

tomchristie:
- [x] Add an `AsyncShieldCancellation` context manager.
- [x] Unit test HTTP/1.1 timeouts on connections.
- [x] Unit test HTTP/2 timeouts on connections.
- [x] Test timeouts on connection pools
- [x] Add to CHANGELOG.

Will close #719.

### DIFF
--- a/httpcore/_async/http11.py
+++ b/httpcore/_async/http11.py
@@ -23,7 +23,7 @@ from .._exceptions import (
     map_exceptions,
 )
 from .._models import Origin, Request, Response
-from .._synchronization import AsyncLock
+from .._synchronization import AsyncLock, AsyncShieldCancellation
 from .._trace import Trace
 from .interfaces import AsyncConnectionInterface
 
@@ -115,8 +115,9 @@ class AsyncHTTP11Connection(AsyncConnectionInterface):
                 },
             )
         except BaseException as exc:
-            async with Trace("response_closed", logger, request) as trace:
-                await self._response_closed()
+            with AsyncShieldCancellation():
+                async with Trace("response_closed", logger, request) as trace:
+                    await self._response_closed()
             raise exc
 
     # Sending the request...
--- a/httpcore/_async/http11.py
+++ b/httpcore/_async/http11.py
@@ -320,7 +320,8 @@ class HTTP11ConnectionByteStream:
             # If we get an exception while streaming the response,
             # we want to close the response (and possibly the connection)
             # before raising that exception.
-            await self.aclose()
+            with AsyncShieldCancellation():
+                await self.aclose()
             raise exc
 
     async def aclose(self) -> None:
--- a/httpcore/_synchronization.py
+++ b/httpcore/_synchronization.py
@@ -182,7 +182,7 @@ class AsyncShieldCancellation:
     def __init__(self) -> None:
         """
         Detect if we're running under 'asyncio' or 'trio' and create
-        a semaphore with the correct implementation.
+        a shielded scope with the correct implementation.
         """
         self._backend = sniffio.current_async_library()
 
--- a/httpcore/_synchronization.py
+++ b/httpcore/_synchronization.py
@@ -171,7 +171,7 @@ class AsyncSemaphore:
             self._anyio_semaphore.release()
 
 
-class AsyncShieldCancellation:  # pragma: nocover
+class AsyncShieldCancellation:
     # For certain portions of our codebase where we're dealing with
     # closing connections during exception handling we want to shield
     # the operation from being cancelled.
@@ -263,7 +263,7 @@ class Semaphore:
         self._semaphore.release()
 
 
-class ShieldCancellation:  # pragma: nocover
+class ShieldCancellation:
     # Thread-synchronous codebases don't support cancellation semantics.
     # We have this class because we need to mirror the async and sync
     # cases within our package, but it's just a no-op.
--- a/httpcore/_synchronization.py
+++ b/httpcore/_synchronization.py
@@ -202,9 +202,6 @@ class AsyncShieldCancellation:
             self._anyio_shield = anyio.CancelScope(shield=True)
 
     def __enter__(self) -> "AsyncShieldCancellation":
-        if not self._backend:
-            self.setup()
-
         if self._backend == "trio":
             self._trio_shield.__enter__()
         else:
--- a/httpcore/_synchronization.py
+++ b/httpcore/_synchronization.py
@@ -171,7 +171,7 @@ class AsyncSemaphore:
             self._anyio_semaphore.release()
 
 
-class AsyncShieldCancellation:
+class AsyncShieldCancellation:  # pragma: nocover
     # For certain portions of our codebase where we're dealing with
     # closing connections during exception handling we want to shield
     # the operation from being cancelled.
@@ -263,7 +263,7 @@ class Semaphore:
         self._semaphore.release()
 
 
-class ShieldCancellation:
+class ShieldCancellation:  # pragma: nocover
     # Thread-synchronous codebases don't support cancellation semantics.
     # We have this class because we need to mirror the async and sync
     # cases within our package, but it's just a no-op.
--- a/httpcore/_synchronization.py
+++ b/httpcore/_synchronization.py
@@ -171,6 +171,58 @@ class AsyncSemaphore:
             self._anyio_semaphore.release()
 
 
+class AsyncShieldCancellation:
+    # For certain portions of our codebase where we're dealing with
+    # closing connections during exception handling we want to shield
+    # the operation from being cancelled.
+    #
+    # with AsyncShieldCancellation():
+    #     ... # clean-up operations, shielded from cancellation.
+
+    def __init__(self) -> None:
+        """
+        Detect if we're running under 'asyncio' or 'trio' and create
+        a semaphore with the correct implementation.
+        """
+        self._backend = sniffio.current_async_library()
+
+        if self._backend == "trio":
+            if trio is None:  # pragma: nocover
+                raise RuntimeError(
+                    "Running under trio requires the 'trio' package to be installed."
+                )
+
+            self._trio_shield = trio.CancelScope(shield=True)
+        else:
+            if anyio is None:  # pragma: nocover
+                raise RuntimeError(
+                    "Running under asyncio requires the 'anyio' package to be installed."
+                )
+
+            self._anyio_shield = anyio.CancelScope(shield=True)
+
+    def __enter__(self) -> "AsyncShieldCancellation":
+        if not self._backend:
+            self.setup()
+
+        if self._backend == "trio":
+            self._trio_shield.__enter__()
+        else:
+            self._anyio_shield.__enter__()
+        return self
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]] = None,
+        exc_value: Optional[BaseException] = None,
+        traceback: Optional[TracebackType] = None,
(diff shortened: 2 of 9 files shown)


httpcore/_synchronization.py
```diff
@@ -171,6 +171,58 @@ async def release(self) -> None:
...
+        else:
+            if anyio is None:  # pragma: nocover
+                raise RuntimeError(
+                    "Running under asyncio requires the 'anyio' package to be installed."
+                )
+
+            self._anyio_shield = anyio.CancelScope(shield=True)
+
+    def __enter__(self) -> "AsyncShieldCancellation":
+        if not self._backend:
```
Pliner, line 205: It might be redundant: `_backend` is initialised in the constructor and there is no `setup` method :)

tomchristie: Oop well spotted, thanks!

httpcore/_synchronization.py
```diff
@@ -171,6 +171,58 @@ async def release(self) -> None:
...
+            self.setup()
+
+        if self._backend == "trio":
+            self._trio_shield.__enter__()
+        else:
+            self._anyio_shield.__enter__()
```
zanieb, line 211: Why not just set `self._shield` in `__init__` since the interfaces are identical?

tomchristie: 1. Maybe?
2. Not sure that'd pass type-checking?
3. Consistency with other cases?

zanieb: I think it'd pass type checking especially if you annotated it with a `Union`. No strong preference here just an observation of multiple conditional checks on the backend when really the backend stops mattering once you are done with `__init__`.

I'd do whatever you've been doing elsewhere in httpcore!

httpcore/_synchronization.py
```diff
@@ -171,6 +171,55 @@ async def release(self) -> None:
...
+    # closing connections during exception handling we want to shield
+    # the operation from being cancelled.
+    #
+    # with AsyncShieldCancellation():
+    #     ... # clean-up operations, shielded from cancellation.
+
+    def __init__(self) -> None:
+        """
+        Detect if we're running under 'asyncio' or 'trio' and create
+        a semaphore with the correct implementation.
```
karosis88, line 185: Is the term "semaphore" appropriate in this context?

tomchristie:
Commenting to show work-in-progress.

At this point I've...

* Added the context manager.
* Put `pragma: nocover` on it so that it doesn't cause coverage to fail, and we can observe that the existing tests are all passing.

Next up...

* Start adding failing test cases from @karosis88's work on #719
* Resolve failing test cases, using new context manager.
* Remove the `nocover` from the context manager.

tests/test_cancellations.py
```diff
@@ -0,0 +1,242 @@
...
+    the request writing.
+    """
+
+    async def write(
+        self, buffer: bytes, timeout: typing.Optional[float] = None
+    ) -> None:
+        await anyio.sleep(999)
+
+    async def aclose(self) -> None:
+        pass
```
zanieb, line 23: It seems like good practice to include an await in these noop methods i.e. `await anyio.sleep(0)` to simulate typical yield behavior for the async function.

zanieb (approved):
lgtm!

Pliner (approved):
Thanks 👍🏻

httpcore/_async/http11.py
```diff
@@ -319,7 +320,8 @@ async def __aiter__(self) -> AsyncIterator[bytes]:
             # If we get an exception while streaming the response,
             # we want to close the response (and possibly the connection)
             # before raising that exception.
-            await self.aclose()
+            with AsyncShieldCancellation():
+                await self.aclose()
```
karosis88, line 324: It is risky to shield this operation without a timeout because it can hang indefinitely, causing our program to hang even if it is cancelled.

tomchristie: > because it can hang indefinitely

Is that true for the network closes?

graingert: If it's a TLS connection then it could wait for shutdown

karosis88: In trio - yes.
Before opening #719, I asked the author of trio.

graingert: I think what you want to do is handle this aclose() in a background TaskGroup

karosis88: From docs

> For example, suppose we call [aclose()](https://trio.readthedocs.io/en/stable/reference-io.html?highlight=Async%20resource#trio.abc.AsyncResource.aclose) on a TLS-encrypted connection. This requires sending a “goodbye” message; but if the peer has become non-responsive, then our attempt to send this message might block forever, and eventually time out and be cancelled. In this case the [aclose()](https://trio.readthedocs.io/en/stable/reference-io.html?highlight=Async%20resource#trio.abc.AsyncResource.aclose) method on [SSLStream](https://trio.readthedocs.io/en/stable/reference-io.html?highlight=Async%20resource#trio.SSLStream) will immediately close the underlying transport stream using [trio.aclose_forcefully()](https://trio.readthedocs.io/en/stable/reference-io.html?highlight=Async%20resource#trio.aclose_forcefully) before raising [Cancelled](https://trio.readthedocs.io/en/stable/reference-core.html#trio.Cancelled).

karosis88: Even if we call `aclose` from a cancelled environment, the resource will be closed anyway. The main question is which timeout is appropriate to wait for a graceful tear down.

tomchristie: > This requires sending a “goodbye” message

Ah, we're on safe ground here...

https://github.com/python-trio/trio/issues/412#issuecomment-359888397

> for an http client, calling aclose always closes the connection immediately (because you'll be using SSLStream in https_compatible=True mode), which is both fastest and nicest to the server.

karosis88: I experimented with the trio `aclose` and discovered something strange. It always sends `RST` packets when closing `TLS` connections, regardless of whether `https_compatible` is true or false, meaning that `aclose` would never block.

karosis88: @tomchristie Take a look at [this code](https://gist.github.com/karosis88/fe4c889df9c6e9a8bfcde7dd615a080c)

tomchristie: Okay, that's really helpful...

We can start to turn that into test cases.

Here's a rough sketch of something that we want to pass:

```python
def main():
    context = ssl.create_default_context()
    context.load_verify_locations('client.pem')
    timeout = {'timeout': {'connect': 5, 'read': 5, 'write': 5, 'pool': 5}}
    print('connection pool opened')
    with httpcore.ConnectionPool(ssl_context=context) as http:
        print('sending request')
        try:
            r = http.request('GET', 'https://127.0.0.1:5051/', extensions=timeout)
        except httpcore.TimeoutException:
            print('timed out')
        else:
            print('completed')
        assert http.connections == []
    print('connection pool closed')
```

Thoughts here...

* We want the default `timeout` extensions to function correctly. For `sync`, `trio`, and `asyncio` cases. With the above case (and async equivalents) they work as expected.
* We also want the above to work when using `async` cancel timeout primitives.

karosis88: > We also want the above to work when using async cancel timeout primitives.

Let's add support for `timeout` argument to our AsyncShieldCancellation, which will act as a `deadline` in trio.CancelScope and anyio.CancelScope.
To convert integer timeout to clock time, we can use the `anyio.current_time` and `trio.current_time functions`.

Exmaple:

``` python
with AsyncShieldCancellation(timeout=5):
    ...
```

karosis88: > @tomchristie Take a look at [this code](https://gist.github.com/karosis88/fe4c889df9c6e9a8bfcde7dd615a080c)

I thought this gist was convincing enough to believe it could hang.

karosis88: Are you requesting a failing test case? Should we demonstrate the test case first, or write some pseudo code that could be our solution?

tomchristie: > I thought this gist was convincing enough to believe it could hang.

The following shows `httpcore` working as desired...

```python
import anyio
import ssl
import socket
import threading
import time
import httpcore

PORT = 5051

def serve():
    sock = socket.socket()
    sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
    sock.bind(('127.0.0.1', PORT))
    sock.listen()
    context = ssl.SSLContext(protocol=ssl.PROTOCOL_TLS_SERVER)
    context.keylog_filename = 'keylogs'
    context.load_cert_chain(certfile='server.pem', keyfile='server.key')
    tls = context.wrap_socket(sock, server_side=True) 
    c, _ = tls.accept()
    time.sleep(10000)

threading.Thread(target=serve).start()
time.sleep(0.1)


async def main():
    context = ssl.create_default_context()
    context.load_verify_locations('client.pem')
    print('connection pool opened')
    async with httpcore.AsyncConnectionPool(ssl_context=context) as http:
        print('sending request')
        async with anyio.move_on_after(5):
            r = await http.request('GET', 'https://127.0.0.1:5051/')
        print(http.connections)
    print('connection pool closed')

anyio.run(main)
```

On this branch the timeout works and the connection has been removed from the pool.
On master the timeout works, but the connection remains stuck in the pool.

So, it's not obvious that we've got a remaining issue to address here?

tomchristie: With DEBUG logs...

```
DEBUG [2023-07-03 11:54:22] asyncio - Using selector: KqueueSelector
connection pool opened
sending request
DEBUG [2023-07-03 11:54:22] httpcore.connection - connect_tcp.started host='127.0.0.1' port=5051 local_address=None timeout=None socket_options=None
DEBUG [2023-07-03 11:54:22] httpcore.connection - connect_tcp.complete return_value=<httpcore._backends.anyio.AnyIOStream object at 0x1092405b0>
DEBUG [2023-07-03 11:54:22] httpcore.connection - start_tls.started ssl_context=<ssl.SSLContext object at 0x1092006c0> server_hostname='127.0.0.1' timeout=None
DEBUG [2023-07-03 11:54:22] httpcore.connection - start_tls.complete return_value=<httpcore._backends.anyio.AnyIOStream object at 0x1091fda80>
DEBUG [2023-07-03 11:54:22] httpcore.http11 - send_request_headers.started request=<Request [b'GET']>
DEBUG [2023-07-03 11:54:22] httpcore.http11 - send_request_headers.complete
DEBUG [2023-07-03 11:54:22] httpcore.http11 - send_request_body.started request=<Request [b'GET']>
DEBUG [2023-07-03 11:54:22] httpcore.http11 - send_request_body.complete
DEBUG [2023-07-03 11:54:22] httpcore.http11 - receive_response_headers.started request=<Request [b'GET']>
DEBUG [2023-07-03 11:54:27] httpcore.http11 - receive_response_headers.failed exception=CancelledError()
DEBUG [2023-07-03 11:54:27] httpcore.http11 - response_closed.started
DEBUG [2023-07-03 11:54:27] httpcore.http11 - response_closed.complete
[]
connection pool closed
```

karosis88: It's my fault. It appears that my gist simply ignores the fact that `standard_compatible` should be set to False in order to avoid the closing handshake.

tomchristie: Ah that makes sense now then. Okay, I think we're good to go on this one then.

karosis88 (approved):

Issue #722, opened 2023-06-14
Support HTTPS proxies.


tomchristie:
Of the following proxy cases...

* HTTP proxy + HTTP destination
* HTTP proxy + HTTPS destination
* HTTPS proxy + HTTP destination
* HTTPS proxy + HTTPS destination

...we currently only support the first two.

In order to support the other two cases we should add a `proxy_ssl_context` argument, which should be used for the SSL connection to the proxy itself.

Implementing the last case will be a little complicated for the sync case, since the stdlib `wrap_socket` behaviour doesn't support TLS-in-TLS.

Note that [the `urllib3` project documentation](https://urllib3.readthedocs.io/en/stable/advanced-usage.html#http-and-https-proxies), do a good job of documenting the possible proxy types.

Also see the `httpx` issue for this... https://github.com/encode/httpx/issues/1434

---

Progress...

- [ ] Supported in standard sync.
- [x] Supported in asyncio.
- [x] Supported in trio.

tomchristie:
I think that closing #721 is a necessary precursor to resolving this enhancement.

PR #732, opened 2023-06-16
Add support for synchronous `TLS-in-TLS` connections.

### LINKED ISSUE #721: Tests for network backends.
Although we have test coverage for our network backends, we're not currently testing against them directly.
Since #699 added the network backends to the public API we should also implement some tests against them.

There's three classes of test case that we'd want covered here...

* Tests for the backends, without TLS.
* Tests for the backends, with TLS.
* Tests for the backends, with TLS-in-TLS.

I'd suggest that we *start* just by implementing some test cases for the first of those, with tests for each of...

* `httpcore.SyncBackend`
* `httpcore.AnyIOBackend`
* `httpcore.TrioBackend`

So, just testing connecting, reading, writing.

karosis88:
`Asyncio` and `Trio` support **TLS-in-TLS** by default, but the standard `ssl` module does not, so this PR adds **TLS-in-TLS** implementation on top of the standard `ssl` module to fully support **TLS-in-TLS** backends. 

**TLS-in-TLS** implementations are required for all backends in order to support `HTTPS` proxies.

Closes #721
Refs: https://github.com/encode/httpcore/issues/722 https://github.com/encode/httpcore/pull/714 https://github.com/encode/httpcore/issues/722

# Todo

- [x] Add tests for the backends, without TLS.
- [x] Add tests for the backends, with TLS.
- [x] Add failing tests for the backends, with TLS-in-TLS.
- [x] Add TLS-in-TLS implementation.
- [x] Add changelog.


karosis88:
We will most likely need to write these tests.

without tls
- `test_connect_without_tls`
- `test_write_without_tls`
- `test_read_without_tls`

with tls
- `test_connect_with_tls`
- `test_write_with_tls`
- `test_read_with_tls`

with tls in tls
- `test_connect_with_tls_in_tls`
- `test_write_with_tls_in_tls`
- `test_need_with_tls_in_tls`

karosis88:
This tests are simply [smoke tests](https://en.wikipedia.org/wiki/Smoke_testing_(software)) that do not require any assertion; we simply want to know if it runs or not.

httpcore/_backends/base.py
```diff
@@ -1,13 +1,19 @@
...
 SOCKET_OPTION = typing.Union[
     typing.Tuple[int, int, int],
     typing.Tuple[int, int, typing.Union[bytes, bytearray]],
     typing.Tuple[int, int, None, int],
 ]
 
+_SELF_NETWORK_STREAM = typing.TypeVar("_SELF_NETWORK_STREAM", bound="NetworkStream")
+_SELF_ASYNC_NETWORK_STREAM = typing.TypeVar(
+    "_SELF_ASYNC_NETWORK_STREAM", bound="AsyncNetworkStream"
+)
```
zanieb, line 15: Perhaps we should add `typing_extensions` as a CI requirement so we can just use `Self`? We'd have to import under a `if TYPE_CHECKING` block but that seems okay.

michaeloliverx: I think `typing_extensions` is fine to have as a dependency. I would like to say say it's widely used.

tests/_backends/test_anyio.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+    async with stream:
+        tls_stream = await stream.start_tls(
+            ssl_context=ssl_context, timeout=CONNECT_TIMEOUT
+        )
+        async with tls_stream:
+            http_request = [b"GET / HTTP/1.1\r\n", b"\r\n"]
+            for chunk in http_request:
+                await tls_stream.write(chunk, timeout=WRITE_TIMEOUT)
+            await tls_stream.read(1024, timeout=READ_TIMEOUT)
```
tomchristie, line 96: This set of tests are looking good. I wonder if we could drop `httpbin` and instead test against plain old socket code? We'll probably need to do that in order to be able to add testing for the TLS-in-TLS case.

karosis88: For simplicity, we can use httpbin for these tests and write a fixture that implements simple tls-in-tls tunneling.

karosis88: We can use something like this to remove httpbin and test tls-in-tls.
```python
def serve(sock):
    with client_socket as sock.accept()[0]:
        while True:
            data = client_socket.recv(1024)
            if not data:
                client_socket.close()
                break
            client_socket.send_all(data)

@pytest.fixture
def tls_server():
    with socket.socket() as socket:
        sock.bind((IP, PORT))
        sock.listen()
        with context.wrap_socket(sock) as tls_sock:
            Thread(target=serve, args=[tls_sock]).start()
            yield tls_sock

tls_server_copy = tls_server

@pytest.fixture
def tls_in_tls_tunnel(tls_server, tls_server_copy):
    Thread(target=tunnel, args=[tls_server, tls_server_copy]).start()
    yield tls_server
```

karosis88: We can use something like this to remove httpbin and test tls-in-tls.
```python
def serve(sock):
    with client_socket as sock.accept()[0]:
        while True:
            data = client_socket.recv(1024)
            if not data:
                client_socket.close()
                break
            client_socket.send_all(data)

@pytest.fixture
def tls_server_socket():
    with socket.socket() as socket:
        sock.bind((IP, PORT))
        sock.listen()
        with context.wrap_socket(sock) as tls_sock:
            yield tls_sock

@pytest.fixture
def tls_server(tls_server_socket):
    Thread(server, tls_server_socket).start()
    yield tls_server_socket.getsocketname()  # ip, port tuple

tls_server_socket_copy = tls_server_socket

@pytest.fixture
def tls_in_tls_tunnel(tls_server_socket, tls_server_socket_copy):
    Thread(target=tunnel, args=[tls_server_socket, tls_server_socket_copy]).start()
    yield tls_server
```

karosis88: But, should we use threads to solve this? It appears to be somewhat complicated.

karosis88: I implemented it using socket logic. if it appears too complicated, we can revert it, but I believe it looks good.

karosis88:
Now we have three failed tests, all of which are synchornous `tls_in_tls` tests that we do not yet support, but adding `tls_in_tls` support from #714 causes all tests to pass, so I believe it is ready for review.

tomchristie:
> Now we have three failed tests, all of which are synchornous tls_in_tls tests that we do not yet support, but adding tls_in_tls support from https://github.com/encode/httpcore/pull/714 causes all tests to pass

Okay, so I'd suggest that the next step would be to add the TLS-in-TLS support from #714. (Not the `proxy_ssl_context`)

At that point we ought to have the tests passing again, and we can review.

karosis88:
Is the `OverallTimeout` class overly complicated? We can remove it, but we will be left with an unreadable '_perform_io' method.

T-256 (approved):
LGTM, opened few changes in https://github.com/karosis88/httpcore/pull/1

httpcore/_backends/sync.py
```diff
@@ -17,6 +19,120 @@
...
+    An overall timeout for TLS operation.
+
+    Because TLS socket operations necessitate
+    more than one TCP socket operation, we need
+    an overall timeout that decreases with each TCP operation.
+    """
+
+    def __init__(self, timeout: typing.Optional[float] = None) -> None:
+        self.timeout = timeout or None  # Replaces `0` with `None`
+        self._start: typing.Optional[float] = None
```
T-256, line 33:
```suggestion
        self._start: float = 0.0
```

httpcore/_backends/sync.py
```diff
@@ -53,10 +170,15 @@ def start_tls(
         }
         with map_exceptions(exc_map):
             try:
-                self._sock.settimeout(timeout)
-                sock = ssl_context.wrap_socket(
-                    self._sock, server_hostname=server_hostname
-                )
+                if isinstance(self._sock, ssl.SSLSocket):
```
tomchristie, line 173: We could do with a comment here explaining why/when we're using TLS and when we're using TLS-in-TLS

httpcore/_backends/sync.py
```diff
@@ -17,6 +19,121 @@
...
+
+    def __exit__(
+        self, exc_type: typing.Any, exc_val: typing.Any, exc_tb: typing.Any
+    ) -> None:
+        if self.timeout is not None:
+            assert self._start
+            self.timeout -= perf_counter() - self._start
+
+
+class SyncTLSStream(NetworkStream):
```
tomchristie, line 49: Let's have a docstring explaining why we need this class.

httpcore/_backends/sync.py
```diff
@@ -17,6 +19,110 @@
...
+        self._incoming = ssl.MemoryBIO()
+        self._outgoing = ssl.MemoryBIO()
+
+        self.ssl_obj = ssl_context.wrap_bio(
+            incoming=self._incoming,
+            outgoing=self._outgoing,
+            server_hostname=server_hostname,
+        )
+
+        self._perform_io(partial(self.ssl_obj.do_handshake), timeout)
```
T-256, line 47:
```suggestion
        self._perform_io(self.ssl_obj.do_handshake, timeout)
```
no need it here

httpcore/_backends/sync.py
```diff
@@ -17,6 +19,110 @@
...
+
+    def _perform_io(
+        self,
+        func: typing.Callable[..., typing.Any],
+        timeout: typing.Optional[float],
+    ) -> typing.Any:
+        ret = None
```
T-256, line 54:
```suggestion
        ret = None
        timeout = timeout or None  # Replaces `0` with `None`

```

httpcore/_backends/sync.py
```diff
@@ -17,6 +19,110 @@
...
+            except (ssl.SSLWantReadError, ssl.SSLWantWriteError) as e:
+                errno = e.errno
+
+            self._sock.settimeout(timeout)
+            operation_start = perf_counter()
+            self._sock.sendall(self._outgoing.read())
+            # If the timeout is `None`, don't touch it.
+            timeout = timeout and timeout - (perf_counter() - operation_start)
+            if timeout is not None and timeout <= 0:  # pragma: no cover
+                raise socket.timeout()
```
T-256, line 68: here shouldn't come before operation?

T-256:
created function to do operation and calculate new timeout via https://github.com/karosis88/httpcore/pull/2 (could we move it as `SyncTLSStream` method?)

another thought here is, `self._sock.settimeout` will be work as expected when using multi threads on one `SyncTLSStream` instance?

httpcore/_backends/sync.py
karosis88: Sure.

httpcore/_backends/sync.py
karosis88: No, we don't do it for other streams.

karosis88: Thanks

T-256 (approved):

httpcore/_backends/sync.py
```diff
@@ -17,6 +19,117 @@
...
+
+        self._perform_io(self.ssl_obj.do_handshake, timeout)
+
+    def _perform_io(
+        self,
+        func: typing.Callable[..., typing.Any],
+        timeout: typing.Optional[float],
+    ) -> typing.Any:
+        ret = None
+        timeout = timeout or None  # Replaces `0` with `None`
```
tomchristie, line 58: ?

karosis88: We need to distinguish between cases where we got 0 after decreasing our timeout and cases where we don't want to handle timeout at all.

T-256: if given paramter's value was `0.0`, it replaces it with `None` to avoid `socket.timeout` at [here](https://github.com/encode/httpcore/pull/732/files#diff-d3a40cf09f980d710bfc8d32385d11086ebd179cec344a9ac5c095251f7b6224R67)

karosis88:
Could you please review this? @tomchristie 

It's a complicated PR that may be difficult to review, but I believe we need it in 0.18.0.

httpcore/_backends/base.py
```diff
@@ -16,6 +20,17 @@ def read(self, max_bytes: int, timeout: typing.Optional[float] = None) -> bytes:
     def write(self, buffer: bytes, timeout: typing.Optional[float] = None) -> None:
         raise NotImplementedError()  # pragma: nocover
 
+    def __enter__(self) -> "Self":
```
tomchristie, line 23: I don't think we need the `typing_extensions.Self` here. (?)
Can we just have this return `NetworkStream`.

The override point is `close()`, not the `__enter__`/`__exit__` which will stay the same even for subclasses.

tomchristie:
Okay, so...

* I'd quite like us to drop the `typing_extensions` here - I don't think we actually need that.
* I'd prefer we drop the `operation_start` timeout smarts, and just have the timeout apply at the SSL read layer. Yes there's a little bit of scope for a connection timeout to be slightly more relaxed than exactly correct, but I think that's an okay trade-off in order to keep things clean and simple.

Do those two changes seem acceptable to you?

Otherwise looking great. Thanks! 😊

httpcore/_backends/base.py
karosis88: If we can inherit, I believe we should always use Self to avoid strange type issues, such as when the instance of SlowNetworkStream is NetworkStream.

karosis88: Example:

```
import typing
from httpcore._backends.base import NetworkStream
from time import sleep

class SlowNetworkStream(NetworkStream):
    
    def read(self, max_bytes: int, timeout: float | None = None) -> bytes:
        sleep(100)

with SlowNetworkStream() as stream:
    reveal_type(stream)
```

OUTPUT test.py:11: note: Revealed type is "httpcore._backends.base.NetworkStream"

karosis88:
> I'd prefer we drop the operation_start timeout smarts, and just have the timeout apply at the SSL read layer. Yes there's a little bit of scope for a connection timeout to be slightly more relaxed than exactly correct, but I think that's an okay trade-off in order to keep things clean and simple.

If we do that, we will be able to hang out indefinitely even though a timeout has been set; do we want that?

httpcore/_backends/sync.py
```diff
@@ -17,6 +19,117 @@
...
+            except (ssl.SSLWantReadError, ssl.SSLWantWriteError) as e:
+                errno = e.errno
+
+            if timeout is not None and timeout <= 0:  # pragma: no cover
+                raise socket.timeout()
+
+            self._sock.settimeout(timeout)
+            operation_start = perf_counter()
+            self._sock.sendall(self._outgoing.read())
+            # If the timeout is `None`, don't touch it.
```
T-256, line 73:
```suggestion
            # If the timeout is `None` or `Zero`, don't touch it.
```

httpcore/_backends/sync.py
```diff
@@ -17,6 +19,117 @@
...
+
+            if errno == ssl.SSL_ERROR_WANT_READ:
+                if timeout is not None and timeout <= 0:  # pragma: no cover
+                    raise socket.timeout()
+
+                self._sock.settimeout(timeout)
+                operation_start = perf_counter()
+                buf = self._sock.recv(self.TLS_RECORD_SIZE)
+
+                # If the timeout is `None`, don't touch it.
```
T-256, line 84:
```suggestion
                # If the timeout is `None` or `Zero`, don't touch it.
```

karosis88: It can not be Zero

T-256: It can be! check [this](https://github.com/encode/httpcore/pull/732/commits/8e38d22e3f06e5324076ef49f80c9fd86c5d4e11#diff-d3a40cf09f980d710bfc8d32385d11086ebd179cec344a9ac5c095251f7b6224R48) and [here](https://github.com/encode/httpcore/pull/732/files#diff-d3a40cf09f980d710bfc8d32385d11086ebd179cec344a9ac5c095251f7b6224R67)

karosis88: If the timeout is zero, then this part should raise an exception, am i wrong?

https://github.com/encode/httpcore/blob/92041d488ed10a48462711885255aa36b0bb429f/httpcore/_backends/sync.py#L67

httpcore/_backends/sync.py
```diff
@@ -17,6 +19,117 @@
...
+                errno = e.errno
+
+            if timeout is not None and timeout <= 0:  # pragma: no cover
+                raise socket.timeout()
+
+            self._sock.settimeout(timeout)
+            operation_start = perf_counter()
+            self._sock.sendall(self._outgoing.read())
+            # If the timeout is `None`, don't touch it.
+            timeout = timeout and timeout - (perf_counter() - operation_start)
```
T-256, line 74:
```suggestion
            timeout = None if timeout is None else timeout - (perf_counter() - operation_start)
```

here during next rounds, it can be exactly equals to zero, so is better to use condition here.

httpcore/_backends/sync.py
```diff
@@ -17,6 +19,117 @@
...
+            if errno == ssl.SSL_ERROR_WANT_READ:
+                if timeout is not None and timeout <= 0:  # pragma: no cover
+                    raise socket.timeout()
+
+                self._sock.settimeout(timeout)
+                operation_start = perf_counter()
+                buf = self._sock.recv(self.TLS_RECORD_SIZE)
+
+                # If the timeout is `None`, don't touch it.
+                timeout = timeout and timeout - (perf_counter() - operation_start)
```
T-256, line 85:
```suggestion
                timeout = None if timeout is None else timeout - (perf_counter() - operation_start)
```

karosis88:
> I'm not suggesting that we remove the timeout, I'm suggesting that we keep it nice and simple and just call `settimeout()` once within that method. (In practice I think that'll actually be perfectly sufficient behaviour)

I'm not sure how we can do that :cry:  ; could you please provide an example?

httpcore/_backends/base.py
tomchristie: Ah okay right.
Could we use the `TypeVar` style, then?

Eg... in `httpx`

https://github.com/encode/httpx/blob/76c9cb65f2a159adb764c2236d139f85b46e1506/httpx/_client.py#L60
https://github.com/encode/httpx/blob/76c9cb65f2a159adb764c2236d139f85b46e1506/httpx/_client.py#L1263

Really prefer us avoiding introducing new third party packages wherever possible.

T-256:
> I'm not suggesting that we remove the timeout, I'm suggesting that we keep it nice and simple and just call `settimeout()` once within that method. (In practice I think that'll actually be perfectly sufficient behaviour)

IMO the current behavior is correct.
if we use `settimeout` once, that it would applies for each socket calls, that is not what we want. we need to set timeout for total calls
.

httpcore/_backends/base.py
karosis88: See https://github.com/encode/httpcore/blob/e31572e0371557d6163d2b7c28676e2b1727673b/httpcore/_backends/base.py

Initially, we used TypeVar, but it was too complicated.

tomchristie:
This...

```python
    def _perform_io(
        self,
        func: typing.Callable[..., typing.Any],
        timeout: typing.Optional[float] = None,
    ) -> typing.Any:
        ret = None

        self._sock.settimeout(timeout)

        while True:
            errno = None
            try:
                ret = func()
            except (ssl.SSLWantReadError, ssl.SSLWantWriteError) as e:
                errno = e.errno

            self._sock.sendall(self._outgoing.read())

            if errno == ssl.SSL_ERROR_WANT_READ:
                buf = self._sock.recv(self.TLS_RECORD_SIZE)

                if buf:
                    self._incoming.write(buf)
                else:
                    self._incoming.write_eof()  # pragma: no cover

            if errno is None:
                return ret
```

The timeouts aren't perfect if we're in TLS-in-TLS, but we probably don't care. (vs the cost of this method being complex to understand.)

httpcore/_backends/sync.py
T-256: Oh correct, actually that that was also my suggestion to raise for `timeout <= 0` before calls 😅

T-256:
And other question here is if `self._sock` already set the timeout, shouldn't we revert the actual value?

karosis88:
> So, according to current timeout calculation, what kind of [timeout](https://www.encode.io/httpcore/extensions/#timeout) it would be? Do we need new timeout type?

It has already been implemented, and it functions exactly like a standard TLS connection, with write timeouts for write operations, read timeouts for read operations, and so on.

@tomchristie  suggests setting the timeout only once, but I'm afraid we won't be able to do so.

One TLS operation, such as `read`, **may necessitate N socket read operations** because we must read not only raw bytes, but **also TLS layer information and encoded data**.

If the socket timeout was only set once, it would be the **same for each socket operation**, so we need to decrease the timeout based on how long our socket operation takes.

So, if we do this, we may have to wait 50 seconds when the timeout is set to 5 or even hang indefinitely.

karosis88:
As a result, the problem I described is easily reproducible.

change the `handle_tunnel_connection` to this

``` python
def handle_tunnel_connection(client_sock: socket.socket) -> None:
    from time import sleep
    with client_sock, socket.create_connection(TLS_ADDRESS) as remote_socket:  # type: ignore
        while True:
            try:
                try:
                    client_sock.settimeout(TUNNEL_READ_WRITE_TIMEOUT)
                    buffer = client_sock.recv(1024)
                    remote_socket.sendall(buffer)
                except socket.timeout:  # pragma: no cover
                    pass

                try:
                    remote_socket.settimeout(TUNNEL_READ_WRITE_TIMEOUT)
                    buffer = remote_socket.recv(1024)

                    while buffer:
                        client_sock.sendall(buffer[:1])
                        buffer = buffer[1:]
                        sleep(1)
                except socket.timeout:  # pragma: no cover
                    pass
            except OSError:
                break
```

And run `pytest -s tests/_backends/test_sync.py::test_read_with_tls_in_tls`

The `read timeout error` should be raised with the **current implementation**.

But, with @tomchristie's [implementation](https://github.com/encode/httpcore/pull/732#issuecomment-1669510161), the test will hang indefinitely..

karosis88:
As a result, the problem I described is easily reproducible.

change the `handle_tunnel_connection` to this

``` python
def handle_tunnel_connection(client_sock: socket.socket) -> None:
    from time import sleep
    with client_sock, socket.create_connection(TLS_ADDRESS) as remote_socket:  # type: ignore
        while True:
            try:
                try:
                    client_sock.settimeout(TUNNEL_READ_WRITE_TIMEOUT)
                    buffer = client_sock.recv(1024)
                    remote_socket.sendall(buffer)
                except socket.timeout:  # pragma: no cover
                    pass

                try:
                    remote_socket.settimeout(TUNNEL_READ_WRITE_TIMEOUT)
                    buffer = remote_socket.recv(1024)

                    while buffer:
                        client_sock.sendall(buffer[:1])
                        buffer = buffer[1:]
                        sleep(1)
                except socket.timeout:  # pragma: no cover
                    pass
            except OSError:
                break
```

And run `pytest -s tests/_backends/test_sync.py::test_read_with_tls_in_tls`

The `read timeout error` should be raised with the **current implementation**.

But, with @tomchristie's [implementation](https://github.com/encode/httpcore/pull/732#issuecomment-1669510161), the test will hang indefinitely..

T-256:
> And other question here is if `self._sock` already set the timeout, shouldn't we revert the actual value?

@karosis88 
I mean, assume the socket before start_tls called, it had already timeout set. but when it goes to nested tls streams, the timeout is changing again and again. does each stream respect other stream timeout?

karosis88:
> > And other question here is if `self._sock` already set the timeout, shouldn't we revert the actual value?
> 
> @karosis88 I mean, assume the socket before start_tls called, it had already timeout set. but when it goes to nested tls streams, the timeout is changing again and again. does each stream respect other stream timeout?

Sorry for the delayed response. I don't think we'll have any problems with this because we're updating the timeout before each network operation to ensure that our network operations use the correct timeout. Please correct me if I am wrong.

T-256:
IMO we should follow implementation in other libraries, we should not apply any changes on sockets as it could have errors in other instances using the same socket.

karosis88:
> IMO we should follow implementation in other libraries, we should not apply any changes on sockets as it could have errors in other instances using the same socket.

I'm not sure what you mean, but if there is an issue, you can describe it and write some simple reproducible code so we can work through it.

T-256:
> I'm not sure what you mean, but if there is an issue, you can describe it and write some simple reproducible code so we can work through it.

I think in this PR the socket's timeout changes over time can cause problems in outer scopes where we set constant value for socket timeout.
But I agree to merge it and then solve those issues by re-set timeouts at those places.

karosis88:
@encode/maintainers Anyone? :(

tomchristie:
I've issued #786 which is a minimal-viable-product alternative to this pull request.

It's a bit atypical in that I'm advocating for a bit of a relaxed approach to testing there, see the description for my motivation there.

tomchristie:
Resolved in #786. (Credit to @karosis88)

PR #733, opened 2023-06-16
Graceful handling for HTTP/2 GoAway frames.

### LINKED ISSUE #730: Graceful handling for HTTP/2 GoAway frames.
When working with an HTTP/2 connection, a GoAway frame can be used to gracefully close the connection.
Our handling for this hasn't been working correctly, see https://github.com/encode/httpx/discussions/2112 and https://github.com/PrefectHQ/prefect/issues/7442.

This was largely resolved by @zanieb's work on #679, which ensured that we at least properly push the connection into a closable state.

However, there's still some improvement to make here. When an HTTP/2 connection is closed with GoAway we'll currently raise exceptions for any requests that are made against the connection before we've seen the frame. Really we ought to gracefully retry any requests that are beyond the `last_stream_id`.

lovelydinosaur:
Closes #730.

* Raise `ConnectionNotAvailable` when appropriate with GoAway frames for requests that can gracefully be retried.
* Some code refactoring
  * Stricter typing for `._events`.
  * Stricter typing for `._connection_terminated`.
  * Make logic in `._receive_events()` clearer by having a switch on the event class.
* New test cases for HTTP/2 GoAway frame behaviour with connection pools.
* Behavior for some existing test cases switches from `RemoteProtocolError` to `ConnectionNotAvailable`.

### DIFF
--- a/httpcore/_sync/http2.py
+++ b/httpcore/_sync/http2.py
@@ -339,7 +339,7 @@ class HTTP2Connection(ConnectionInterface):
         with self._read_lock:
             if self._connection_terminated is not None:
                 last_stream_id = self._connection_terminated.last_stream_id
-                if last_stream_id and stream_id > last_stream_id:
+                if stream_id and last_stream_id and stream_id > last_stream_id:
                     self._request_count -= 1
                     raise ConnectionNotAvailable()
                 raise RemoteProtocolError(self._connection_terminated)


tomchristie:
Okay, this is a bit of a chunky one for review.
I'm happy to take it step-by-step instead if that's preferred.

httpcore/_sync/http2.py
```diff
@@ -338,16 +360,21 @@ def _receive_events(
...
+                    elif isinstance(
+                        event,
+                        (
+                            h2.events.ResponseReceived,
+                            h2.events.DataReceived,
+                            h2.events.StreamEnded,
+                            h2.events.StreamReset,
+                        ),
+                    ):
+                        event_stream_id = getattr(event, "stream_id", 0)
```
Pliner, line 372: Is `getattr` here still needed? I've checked all four event types and it seems that stream_id exists in all of them (at least it is initialised to None in the constructors).

tomchristie: Good spot thanks. That's a nice bit of cleaning up there.

httpcore/_async/http2.py
```diff
@@ -319,8 +337,12 @@ async def _receive_events(
         for a given stream ID.
         """
         async with self._read_lock:
-            if self._connection_error_event is not None:  # pragma: nocover
-                raise RemoteProtocolError(self._connection_error_event)
+            if self._connection_terminated is not None:
+                last_stream_id = self._connection_terminated.last_stream_id
+                if last_stream_id and stream_id > last_stream_id:
```
Pliner, line 342: Are there any chances for stream_id to be None here?

Pliner (approved):

zanieb:
Wonderful. Thanks for taking this on @tomchristie !

PR #734, opened 2023-06-17
proxy mode for http proxies

### LINKED ISSUE #735: Support force-forwarding and force-tunneling
Discussed in #703 

We need an API to specify whether a tunnel proxy or a forward proxy connection should be established.
We are currently using forward proxies for **http** destinations and tunnel proxies for **https** destinations; but, in some cases, we might have to use forward for **https** or tunnel for **http** destinations, which is why this issue was created.

T-256:
Closes #735 
Refs: https://github.com/encode/httpcore/discussions/703 https://github.com/encode/httpcore/pull/716

- [x] tests (forward and tunnel tests already implemented)
- [x] docs + warn about insecure use of `ProxyMode.FORWARD`
- [x] changelog


T-256:
current implementation is modified and upgraded of first option discussed in https://github.com/encode/httpcore/discussions/703#discussioncomment-6201442.
for second option we can use only one flagged param:
```py
class ProxyMode:
    HTTPS_TUNNEL = 1
    HTTP_FORWARD = 2


def HTTPProxy(
    ...
    mode: Optional[ProxyMode] = None,
): ...
```

httpcore/_async/http_proxy.py
```diff
@@ -228,6 +238,7 @@ def __init__(
         http2: bool = False,
         network_backend: Optional[AsyncNetworkBackend] = None,
         socket_options: Optional[Iterable[SOCKET_OPTION]] = None,
+        is_tls: bool = True,
```
karosis88, line 241: Do we really need this? We can simply check if `remote_origin` is https or not

httpcore/_async/http_proxy.py
```diff
@@ -134,7 +137,13 @@ def __init__(
             ] + self._proxy_headers
 
     def create_connection(self, origin: Origin) -> AsyncConnectionInterface:
-        if origin.scheme == b"http":
+        is_tls = origin.scheme == b"https"
+
+        forwarable = not (self._mode & ProxyMode.HTTP_TUNNEL)
+        if is_tls:
+            forwarable = bool(self._mode & ProxyMode.HTTPS_FORWARD)
```
karosis88, line 144: Let's just use `is` operator in such cases

For example
```python
forwardable = not (self._mode is ProxyMode.HTTP_TUNNEL)
```

httpcore/_models.py
```diff
@@ -481,3 +482,9 @@ async def aclose(self) -> None:
             )
         if hasattr(self.stream, "aclose"):
             await self.stream.aclose()
+
+
+class ProxyMode(enum.IntFlag):
```
karosis88, line 487:
```suggestion
class ProxyMode(enum.Enum):
```

karosis88:
We should probably wait for #714, then open the issue, and only then go through this for further clarity.

httpcore/_async/http_proxy.py
T-256: yes, makes sense, thanks for review

httpcore/_async/http_proxy.py
T-256: Consider when `mode=ProxyMode.HTTP_TUNNEL | ProxyMode.HTTPS_FORWARD`

httpcore/_models.py
T-256: As I said above, we need to usage as flag for possibility of both selection.

httpcore/_async/http_proxy.py
karosis88: Is there any differences between `mode=ProxyMode.HTTP_TUNNEL | ProxyMode.HTTPS_FORWARD`
and `mode=ProxyMode.DEFAULT`?

T-256: Yes obviously, `ProxyMode.HTTP_TUNNEL | ProxyMode.HTTPS_FORWARD` would reverse the defaults.
Here is all options:


`ProxyMode.DEFAULT`:
HTTP dest: forwarding
HTTPS dest: tunneling

`ProxyMode.HTTP_TUNNEL | ProxyMode.HTTPS_FORWARD`:
HTTP dest: tunneling
HTTPS dest: forwarding

`ProxyMode.HTTP_TUNNEL`:
HTTP dest: tunneling
HTTPS dest: tunneling

`ProxyMode.HTTPS_FORWARD`:
HTTP dest: forwarding
HTTPS dest: forwarding

T-256:
> We should probably wait for #714, then open the issue, and only then go through this for further clarity.

I didn't understand. Will #714 create any blocker for this PR or vice versa?

karosis88:
I've started an [issue](https://github.com/encode/httpcore/issues/735) for this PR.

karosis88:
Let's change the title of this PR to something like "Add force-tunneling and force-forwarding support for proxies."

tomchristie:
Really I'd prefer that we work through design discussions first before starting on the pull request, but...

* I'm okay with reviewing this once there's docs.
* We also need to consider what the implications/expectations are for pushing this downstream to `httpx`.

httpcore/_async/http_proxy.py
```diff
@@ -134,7 +137,11 @@ def __init__(
             ] + self._proxy_headers
 
     def create_connection(self, origin: Origin) -> AsyncConnectionInterface:
-        if origin.scheme == b"http":
+        fwd = False if self._mode & ProxyMode.HTTP_TUNNEL else True
+        if origin.scheme == b"https":
+            fwd = True if self._mode & ProxyMode.HTTPS_FORWARD else False
+
+        if fwd:
```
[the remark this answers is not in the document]
tomchristie: Oops, I've updated the suggestion in line with point 1.

docs/proxies.md
```diff
@@ -51,6 +51,35 @@ proxy = httpcore.HTTPProxy(
...
+```py
+import httpcore
+import base64
+
+proxy = httpcore.HTTPProxy(
+    proxy_url="http://127.0.0.1:8080/",
+    mode=httpcore.ProxyMode.HTTP_TUNNEL | httpcore.ProxyMode.HTTPS_FORWARD,
+)
+```
+(WARN: `ProxyMode.HTTPS_FORWARD` is unsafe, use it when you trust proxy server)
```
tomchristie, line 82: how do other texts phrase this? can we explain the implications to readers more clearly?

docs/proxies.md
```diff
@@ -51,6 +51,35 @@ proxy = httpcore.HTTPProxy(
...
+)
+```
+`ProxyMode` is a flag parameter, also you can reverse current default behavior:
+```py
+import httpcore
+import base64
+
+proxy = httpcore.HTTPProxy(
+    proxy_url="http://127.0.0.1:8080/",
+    mode=httpcore.ProxyMode.HTTP_TUNNEL | httpcore.ProxyMode.HTTPS_FORWARD,
```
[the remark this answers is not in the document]
T-256: by default, plain http requests use forwarding method to connect proxy server, and https requests use tunneling.
this will reverse the default.

docs/proxies.md
T-256: perhaps here needs fully explain that what is forwarding and tunneling proxy.

T-256:
Any new thoughts here?
it stuck on implementation design. few I can suggest are:
- separated parameters for http and https (implemented in [here](https://github.com/encode/httpcore/compare/master...T-256:httpcore:28c0e901a12e62baa7e122a918f1fabb5d5919f1))
- one flagged parameter (current implementation)
- one routing parameter: `dict[str, Literal["TUNNEL"] | Literal["FORWARD"]]` (like in [httpx](https://www.python-httpx.org/advanced/#routing))

And if you have new idea these are things should be considered to work:
- don't change current default: forward http and tunnel https
- able to set force forward or force tunnel
- opposite of current default: tunnel http and forward https (maybe not mandatory? any use cases?)

docs/proxies.md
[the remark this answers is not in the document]
tomchristie: > this will reverse the default

I can't see any reason we'd want to support this.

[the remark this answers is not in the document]
T-256: I tried to cover all possibilities,
So, as it is useless, should I remove it?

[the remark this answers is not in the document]
tomchristie: Yes let's not support that as a use-case.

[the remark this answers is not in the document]
T-256: done in https://github.com/encode/httpcore/pull/734/commits/067d76f3c92611629d768d01cb5f03eddb53b894

T-256:
async cancellation test failing randomly. not related to this PR.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 
 - Handle HTTP/1.1 half-closed connections gracefully. (#641)
+- Added proxy_mode for HTTPProxy and AsyncHTTPProxy. (#734)
```
karosis88, line 10:
```suggestion
- Added proxy_mode for `HTTPProxy` and `AsyncHTTPProxy`. (#734)
```

docs/proxies.md
```diff
@@ -51,6 +51,29 @@ proxy = httpcore.HTTPProxy(
...
+proxy = httpcore.HTTPProxy(
+    proxy_url="http://127.0.0.1:8080/",
+    proxy_mode=httpcore.ProxyMode.TUNNEL,  # Forces HTTP requests also use Tunneling
+)
+```
+
+Note that `ProxyMode.FORWARD` will enable forwarding https requests, so they will be visible for proxy server.
+It means handling TLS stuffs like certificate validation would on proxy side.
+
+
```
karosis88, line 76:
```suggestion

```

httpcore/_models.py
```diff
@@ -481,3 +482,9 @@ async def aclose(self) -> None:
             )
         if hasattr(self.stream, "aclose"):
             await self.stream.aclose()
+
+
+class ProxyMode(enum.IntEnum):
+    DEFAULT = 0
```
karosis88, line 488: Let's get rid of this `DEFAULT` field; I don't think we need it.

httpcore/_sync/http_proxy.py
```diff
@@ -134,7 +137,9 @@ def __init__(
             ] + self._proxy_headers
 
     def create_connection(self, origin: Origin) -> ConnectionInterface:
-        if origin.scheme == b"http":
+        if (self._proxy_mode == ProxyMode.FORWARD) or (
+            self._proxy_mode == ProxyMode.DEFAULT and origin.scheme == b"http"
```
karosis88, line 141:
```suggestion
        if (self._proxy_mode is ProxyMode.FORWARD) or (
            self._proxy_mode is None and origin.scheme == b"http"
```

httpcore/_sync/http_proxy.py
```diff
@@ -74,6 +74,7 @@ def __init__(
         uds: Optional[str] = None,
         network_backend: Optional[NetworkBackend] = None,
         socket_options: Optional[Iterable[SOCKET_OPTION]] = None,
+        proxy_mode: ProxyMode = ProxyMode.DEFAULT,
```
karosis88, line 77:
```suggestion
        proxy_mode: Optional[ProxyMode] = None,
```

karosis88 (changes requested):

T-256:
According to [Mozzilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Proxy_servers_and_tunneling):

> There are two types of proxies: forward proxies (or tunnel, or gateway) and reverse proxies (used to control and protect access to a server for load-balancing, authentication, decryption or caching).

So, should we change the `ProxyMode.FORWARD` name?
Which one is proper name?
- ABSOLUTE_URL
- ABSOULTE
- PLAIN
- GATEWAY

T-256:
Is it possible to include in https://github.com/encode/httpcore/pull/768?

tomchristie:
So...

1. FORWARD/TUNNEL makes more sense to me. That's a reading that'd make sense to me from [these docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Proxy_servers_and_tunneling). It's also [consistent with `urllib3`](https://urllib3.readthedocs.io/en/stable/advanced-usage.html#http-and-https-proxies).
2. I'm not fully convinced yet that we couldn't instead have a `use_forwarding_for_https` flag. That's the API that `urllib3` use for this manual control, and they don't seem to have open issues suggesting that "allow CONNECT tunnelling to `http`" is something their users are missing. I can see that we discussed this briefly at https://github.com/encode/httpcore/discussions/703#discussioncomment-6174581. Is someone able to provide an example of a proxy that would require this configuration, so I'm able to test it out and verify that we really do want to support that as a configuration option?

karosis88:
1. Completely agree
2. If the proxy server supports tunneling and it's not a public proxy, then using tunneling over HTTP makes sense.
Assume that we are using a local HTTP proxy server that supports tunneling. We do not need to configure certificates and make the proxy server secure because we actually trust it.

Example

```
  trusted
host -> proxy

   not trusted
proxy -> remote server

| Solution |

HTTP tunneling gives
      http       https
host   ->   proxy -> remote-server
```

docs/proxies.md
```diff
@@ -51,6 +51,28 @@ proxy = httpcore.HTTPProxy(
 )
 ```
 
+## Proxy Connection Mode
+
+There are two types of HTTP proxy:
+
+1. Forwading (HTTP [absolute-url](https://tools.ietf.org/html/rfc7230#section-5.3.2))
```
T-256, line 58:
```suggestion
1. Forwarding (HTTP [absolute-url](https://tools.ietf.org/html/rfc7230#section-5.3.2))
```

docs/proxies.md
```diff
@@ -51,6 +51,28 @@ proxy = httpcore.HTTPProxy(
...
 ```
 
+## Proxy Connection Mode
+
+There are two types of HTTP proxy:
+
+1. Forwarding (HTTP [absolute-url](https://tools.ietf.org/html/rfc7230#section-5.3.2))
+2. Tunneling (HTTP [`CONNECT` method](https://tools.ietf.org/html/rfc7231#section-4.3.6))
+
+By default we forwarding plain http requests and tunneling https requests.
```
tomchristie, line 61:
```suggestion
By default `httpcore` will use forwarding for http requests and tunneling for https requests.
```

docs/proxies.md
```diff
@@ -51,6 +51,28 @@ proxy = httpcore.HTTPProxy(
...
+You can change this behavior with `proxy_mode: httpcore.ProxyMode` parameter:
```
tomchristie, line 62:
```suggestion

You can change this behavior with `proxy_mode: httpcore.ProxyMode` parameter:

```

docs/proxies.md
```diff
@@ -51,6 +51,28 @@ proxy = httpcore.HTTPProxy(
...
+```py
+import httpcore
+import base64
+
+proxy = httpcore.HTTPProxy(
+    proxy_url="http://127.0.0.1:8080/",
+    proxy_mode=httpcore.ProxyMode.TUNNEL,  # Forces HTTP requests also use Tunneling
```
tomchristie, line 69: I guess let's just drop the comment here, since the docs are already giving us context...

```suggestion
    proxy_mode=httpcore.ProxyMode.TUNNEL,
```

karosis88 (approved):
LGTM!

PR #745, opened 2023-07-04
Add support for HTTPS proxies (available to trio/asyncio)

### LINKED ISSUE #722: Support HTTPS proxies.

karosis88:
Refs #722 

Related #732 (Required for sync support)

## TODO

- [x] Add `proxy_ssl_context` argument to `AsyncForwardHTTPConnection`, `AsyncTunnelHTTPConnection` and `AsyncHTTPProxy` classes
- [x] Add changelog


T-256 (approved):

tomchristie:
Okay, so attempting to test this...

I'm using `trio` in this example, so as not to be blocked by https://github.com/encode/httpcore/pull/732.

Installation...

```shell
$ pip install httpcore trio trustme proxy.py
```

Use `trustme` to generate certs...

```python
$ python -m trustme
Generated a certificate for 'localhost', '127.0.0.1', '::1'
Configure your server to use the following files:
  cert=/Users/tomchristie/GitHub/encode/httpcore/server.pem
  key=/Users/tomchristie/GitHub/encode/httpcore/server.key
Configure your client to use the following files:
  cert=/Users/tomchristie/GitHub/encode/httpcore/client.pem
```

Running the proxy...

```shell
$ proxy --cert-file server.pem --key-file server.key 
```

Make the requests...

```python
import ssl
import trio
import httpcore


ctx = ssl.create_default_context()
ctx.load_verify_locations('client.pem')

async def main():
    async with httpcore.AsyncHTTPProxy(proxy_url="http://127.0.0.1:8899/", proxy_ssl_context=ctx) as http:
        response = await http.request("GET", "https://www.example.com/")
        print(response)
        response = await http.request("GET", "http://www.example.com/")
        print(response)


trio.run(main)
```

Fails with...

```python
[...]
ssl.SSLError: [SSL: HTTPS_PROXY_REQUEST] https proxy request (_ssl.c:992)
```

What am I getting wrong here?

tomchristie:
Also, occurs to me that something we could do here would be decouple this from https://github.com/encode/httpcore/pull/732.

Indicate in the CHANGELOG/docs that HTTPS proxies are only supported with `trio` and `asyncio`, and have this PR raise a runtime error if TLS-in-TLS is attempted with the sync backend.

karosis88:
What am I getting wrong here?

Just change the proxy url scheme to https.

karosis88:
> Also, occurs to me that something we could do here would be decouple this from #732.
> 
> Indicate in the CHANGELOG/docs that HTTPS proxies are only supported with `trio` and `asyncio`, and ensure that we raise a runtime error if TLS-in-TLS is attempted with the sync backend.

We should probably merge #732 first, then go over this one.

tomchristie:
Wonderful thanks, tested now and all working as expected.

Okay so, should we add the following error cases here?...

* If `proxy_url` is set to `https` and `proxy_ssl_context` is unset, then raise an exception.
* If `proxy_url` is set to `http` and `proxy_ssl_context` is set, then raise an exception.

I'd suggest that this PR is neat and uncontentious, so let's work to getting it merged, as "Add support for HTTPS proxies (available to trio/asyncio)". Then follow up with the more awkward sync support.

Remaining here is...

* Documentation.
* Raise an error in the sync backend if TLS-in-TLS is attempted.

docs/proxies.md
```diff
@@ -51,10 +51,33 @@ proxy = httpcore.HTTPProxy(
 )
 ```
 
-## Proxy SSL and HTTP Versions
+## Proxy SSL
```
tomchristie, line 54: These docs are really nice, thanks! 💚

karosis88:
> If proxy_url is set to https and proxy_ssl_context is unset, then raise an exception.

Because most users do not interact with SSL contexts, we can allow this behavior to keep things as simple as possible.

When using the public HTTPS proxy, for example, the user must create the default SSL context and pass it to the HTTPProxy class without any modifications or configurations.

> If proxy_url is set to http and proxy_ssl_context is set, then raise an exception.

Agree

docs/proxies.md
```diff
@@ -51,10 +51,33 @@ proxy = httpcore.HTTPProxy(
...
+Also, when using HTTPS proxies, you may need to configure the SSL context, which you can do with the `proxy_ssl_context` argument.
+
+```python
+import ssl
+import httpcore
+
+proxy_ssl_context = ssl.create_default_context()
+proxy_ssl_context.check_hostname = False
+
+proxy = httpcore.HTTPProxy('http://127.0.0.1:8080/', proxy_ssl_context=proxy_ssl_context)
```
tomchristie, line 73:
```suggestion
proxy = httpcore.HTTPProxy('https://127.0.0.1:8080/', proxy_ssl_context=proxy_ssl_context)
```

CHANGELOG.md
```diff
@@ -4,8 +4,9 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## Unreleased
+## development
 
+- Add support for HTTPS proxies. (#745)
```
```suggestion
- Add support for HTTPS proxies. Currently only available for async. (#745)
```

CHANGELOG.md
```diff
@@ -4,8 +4,9 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## Unreleased
+## development
```
tomchristie, line 7:
```suggestion
## Unreleased
```

I don't recall which of these we prefer, but suggest we don't change it here,

tomchristie (approved):

tomchristie:
Great - No objections to this.

Welcome to merge once you're also happy with it @karosis88.

Issue #751, opened 2023-07-07
Use `pyproject.toml` instead of `setup.cfg` and `setup.py`


karosis88:
Discussed in: #748

Because `httpx` uses `pyproject.toml` for packaging, we should also use `pyproject.toml` for simplicity.
It would also be preferable to move all project configurations into pyproject toml, as httpx does.

It appears that we will need two PRs to make it much cleaner and more reviewable, one for packaging and one for configurations.

## Packaging part
The PR should look very similar to [this](https://github.com/encode/httpx/pull/2334/), so that we can easily migrate based on that PR.

## Configuration part
The configuration part is also simple; according to [this](https://github.com/encode/httpx/pull/2686), we can migrate quickly.

tomchristie:
Closed via #752

PR #754, opened 2023-07-11
Move configurations to `pyproject.toml`, use ruff

karosis88:
Issue: https://github.com/encode/httpcore/issues/751
Refs: https://github.com/encode/httpx/pull/2648 https://github.com/encode/httpx/pull/2680

Let's move all of our configurations to the `pyproject.toml` file so that we have a single file for our project metadata. I also added ruff as a dependency because `flake8` does not support `pyproject.toml`.


scripts/check
```diff
@@ -8,8 +8,6 @@ export SOURCE_FILES="httpcore tests"
 
 set -x
 
-${PREFIX}isort --check --diff --project=httpcore $SOURCE_FILES
-${PREFIX}black --exclude '/(_sync|sync_tests)/' --check --diff --target-version=py37 $SOURCE_FILES
```
zanieb, line 12: I don't think we want to remove black here

scripts/check
```diff
@@ -8,8 +8,6 @@ export SOURCE_FILES="httpcore tests"
...
-${PREFIX}flake8 $SOURCE_FILES
+${PREFIX}ruff check $SOURCE_FILES
```
zanieb, line 11: I think the `--show-source` option is nice here

scripts/lint
```diff
@@ -9,8 +9,7 @@ export SOURCE_FILES="httpcore tests"
 set -x
 
 ${PREFIX}autoflake --in-place --recursive --remove-all-unused-imports $SOURCE_FILES
-${PREFIX}isort --project=httpcore $SOURCE_FILES
-${PREFIX}black --target-version=py37 --exclude '/(_sync|sync_tests)/' $SOURCE_FILES
```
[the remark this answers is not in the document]
karosis88: Yes, we use ruff and black in httpx, but does it make sense?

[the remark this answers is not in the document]
karosis88: Okay according to https://beta.ruff.rs/docs/faq/#is-ruff-compatible-with-black let's use both
> Yes. Ruff is compatible with [Black](https://github.com/psf/black) out-of-the-box, as long as the line-length setting is consistent between the two.
As a project, Ruff is designed to be used alongside Black and, as such, will defer implementing stylistic lint rules that are obviated by autoformatting.

karosis88:
How did I break the `coverage` here? :-)

scripts/check
karosis88: Yes, thank you

pyproject.toml
```diff
@@ -66,3 +66,55 @@ path = "README.md"
...
+module = "hpack.*"
+ignore_missing_imports = true
+
+[tool.isort]
+profile = "black"
+combine_as_imports = true
+known_first_party = "httpcore,tests"
+known_third_party = "brotli,certifi,chardet,cryptography,h11,h2,hstspreload,pytest,rfc3986,setuptools,sniffio,trio,trustme,urllib3,uvicorn"
+skip = ["httpcore/_sync/,tests/_sync"]
+
```
zanieb, line 93: If `isort` is removed we should probably drop this section

```suggestion
```

Kludex (commented):
We also removed autoflake in other encode projects, since it becomes redundant as well.

scripts/lint
```diff
@@ -9,7 +9,7 @@ export SOURCE_FILES="httpcore tests"
 set -x
 
 ${PREFIX}autoflake --in-place --recursive --remove-all-unused-imports $SOURCE_FILES
```
zanieb, line 11: Needs removal

karosis88: Sure

zanieb (approved):

karosis88:
@tomchristie , did we break anything? Why does this test fail while the branch is up to date?

tomchristie:
> Why does this test fail while the branch is up to date?

Which test?

Kludex:
> > Why does this test fail while the branch is up to date?
> 
> Which test?

It was a previous issue from click. Now it's solved.

karosis88:
@tomchristie check [this job](https://github.com/encode/httpcore/actions/runs/5532888849/jobs/10095604921#step:7:66)

PR #757, opened 2023-07-14
Ensure HTTP/2 connections are gracefully closed on async cancellation

### LINKED ISSUE #756: Ensure HTTP/2 connections are gracefully closed on async cancellation
- [x] Initially raised as discussion #755 

The issue is that some HTTP/2 connections can fail and not be cleaned up, resulting in our connection being in a broken state.
When the asynchronous semaphore acquisition is cancelled, our `handle_request` method propagates the exception and never calls clean up.

The first step, as always, will be to write a failing test, but in this case that will be difficult because we need to mock the `AsyncSemaphore` object, more specifically, we want the acquisition method to be extremely slow.

Let's use `monkeypatching` to write a failing test with no complex changes, and then we can change that failing test if we want.

karpetrosyan:
Closes #756

# Checklist

# TODO

- [x] Add failing tests
- [x] Gracefully handle cancellation for unstarted http/2 requests 
- [x] Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Fix the unstarted http/2 connection cancellations. (#757)
 - Handle HTTP/1.1 half-closed connections gracefully. (#641)
 
 ## 0.17.3 (5th July 2023)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
-- Fix the unstarted http/2 connection cancellations. (#757)
+- Ensure that the cancellation of HTTP/2 connections does not break the state. (#757)
 - Handle HTTP/1.1 half-closed connections gracefully. (#641)
 
 ## 0.17.3 (5th July 2023)
--- a/httpcore/_async/http2.py
+++ b/httpcore/_async/http2.py
@@ -128,13 +128,15 @@ class AsyncHTTP2Connection(AsyncConnectionInterface):
                         await self._max_streams_semaphore.acquire()
                 except BaseException:
                     with AsyncShieldCancellation():
-                        await self._state_housekeeping()
+                        async with self._state_lock:
+                            await self._state_housekeeping()
                     raise
         try:
             await self._max_streams_semaphore.acquire()
         except BaseException:  # pragma: no cover
             with AsyncShieldCancellation():
-                await self._state_housekeeping()
+                async with self._state_lock:
+                    await self._state_housekeeping()
             raise
 
         try:
--- a/httpcore/_async/http2.py
+++ b/httpcore/_async/http2.py
@@ -123,10 +123,19 @@ class AsyncHTTP2Connection(AsyncConnectionInterface):
                 )
                 self._max_streams_semaphore = AsyncSemaphore(local_settings_max_streams)
 
-                for _ in range(local_settings_max_streams - self._max_streams):
-                    await self._max_streams_semaphore.acquire()
-
-        await self._max_streams_semaphore.acquire()
+                try:
+                    for _ in range(local_settings_max_streams - self._max_streams):
+                        await self._max_streams_semaphore.acquire()
+                except BaseException:
+                    with AsyncShieldCancellation():
+                        await self._state_housekeeping()
+                    raise
+        try:
+            await self._max_streams_semaphore.acquire()
+        except BaseException:  # pragma: no cover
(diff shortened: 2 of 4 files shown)


httpcore/_async/http2.py
```diff
@@ -123,10 +123,21 @@ async def handle_async_request(self, request: Request) -> Response:
...
+                        async with self._state_lock:
+                            await self._state_housekeeping()
+                    raise
+        try:
+            await self._max_streams_semaphore.acquire()
+        except BaseException:  # pragma: no cover
+            with AsyncShieldCancellation():
+                async with self._state_lock:
+                    await self._state_housekeeping()
+            raise
```
zanieb, line 140: Should `self._state_housekeeping` just be a context manager then you could simplify both of these cases to

```python
async with self._cleanup_on_error():
    ...  # do work
```

and the try/except logic can be handled in the context manager's yield.

I find this easier to read and it consolidates logic for the (slightly dangerous) capture of a `BaseException`.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Fix the unstarted http/2 connection cancellations. (#757)
```
zanieb, line 9: Perhaps

```suggestion
- Ensure HTTP/2 connections are closed on cancellation. (#757)
```

I think the title of this pull request is a little misleading as well :)

httpcore/_async/http2.py
karosis88: > I find this easier to read and it consolidates logic for the (slightly dangerous) capture of a BaseException.

We always catch the `BaseException` in `httpcore` for cleanup logic, so I believe it is safe.

> Should self._state_housekeeping just be a context manager then you could simplify both of these cases to

Can you provide an example of how it might look? Context managers are frequently used for try/finally logic rather than try/except logic, am I correct? Only when an exception occurs do we need to do housekeeping.

CHANGELOG.md
karosis88: Thanks! I changed it to be more descriptive.

httpcore/_async/http2.py
karosis88: Or do we really need that much extra complexity?

zanieb: I think `try/except` logic is common as well.

Here's a sketch of what I would do https://github.com/encode/httpcore/commit/178e7d1b8cd93a581ce23bea172bdea13a38d799

I find the intent clearer here — I'm not sure it's more complex than maintaining the repeated implementation for properly handling an exception with the cancellation shield and state lock.

karosis88: Let's keep it for another PR. `httpcore` now uses this type of cleanup everywhere, and if we want to replace common exception handling with context managers, we'll probably want to do it elsewhere as well.

karosis88: For example `_response_closed` call

```python
except BaseException as exc:
    with ShieldCancellation():
        with Trace("response_closed", logger, request) as trace:
            self._response_closed()
    raise exc
```

T-256 (commented):
could `(async )with self._state_lock:` move under `_state_housekeeping` function?

zanieb:
> could `(async )with self._state_lock:` move under `_state_housekeeping` function?

I agree, it's always used.

httpcore/_async/http2.py
zanieb: I don't quite think it follows that we'd want to do it elsewhere, the example you gave here is specific to a closed response whereas the other one is "some statement, if interrupted, needs us to clean up the state".

Anyway we don't need to go back and forth here. It really depends what style @tomchristie prefers for the project.

tomchristie:
At thanks! The failing test case is especially helpful here.

Pretty sure this resolution is overcomplicated, tho. What we actually want here is...

1. This existing `try...except` needs to be expanded...

https://github.com/encode/httpcore/blob/21e47a7481fbad3719417899113e1435a340b752/httpcore/_async/http2.py#L106-L113

...so that it instead covers *all* of this block...

https://github.com/encode/httpcore/blob/21e47a7481fbad3719417899113e1435a340b752/httpcore/_async/http2.py#L106-L127

2. This call to `acquire()`...

https://github.com/encode/httpcore/blob/21e47a7481fbad3719417899113e1435a340b752/httpcore/_async/http2.py#L129

Should also be surrounded with...

```python
try:
    await self._max_streams_semaphore.acquire()
 except BaseException as exc: 
     with AsyncShieldCancellation(): 
         await self.aclose() 
     raise exc
```

We could actually avoid (1) if we instead addressed the refactoring suggested in https://github.com/encode/httpcore/discussions/738 *or* refactored `self._send_connection_init(**kwargs)` so that our initial handshake waits for the settings response before we instantiate our `max_streams_semaphore`. But let's not be getting into that just now.

PR #758, opened 2023-07-15
refactor: simplify `exponential_backoff`

m9810223:
# Summary

# Checklist


tomchristie (commented):
😅 true.

could probably also do with a docstring.

m9810223:
let me add the docstring~

tomchristie (approved):

karosis88:
Do we need changeloog?

PR #696, opened 2023-05-22
Add sni hostname extension

karosis88:
closes #691

PR #773, opened 2023-08-08
Support `sni_hostname` extension with SOCKS proxy

Allgot:
* Support `sni_hostname` extension with SOCKS proxy.

* This is a complementary update of #696.
---------

Co-authored-by: Tom Christie <tom@tomchristie.com>

# Summary

A new extension, `sni_hostname` has enalbed users to specify the hostname which would be used during TLS handshake. However, the update was inconsistent, resulting in the incomplete extension. Currently, the `sni_hostname` extension are not handled if SOCKS proxy is in use. This PR would bridge this gap.

# Checklist

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: "actions/checkout@v3"
       - uses: "actions/setup-python@v4"
         with:
-          python-version: 3.7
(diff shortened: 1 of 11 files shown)


Allgot:
Sorry for the confusing PR. I created a new branch from the HEAD by mistake.

I would close this PR right now, then create new one properly. Sorry.

Issue #638, opened 2022-12-07
Document and cleanup `scripts`.


tomchristie:
Prompted by @agronholm on [our chat channel](https://gitter.im/encode/community?at=6380e60825ca105c88d16bec).

We should add a `README` to our `scripts` directory, explaining that we're using the GitHub "[scripts to pull them all](https://github.blog/2015-06-30-scripts-to-rule-them-all/)" pattern, and show how that ties in with our [`test`](https://github.com/encode/httpcore/blob/master/.github/workflows/test-suite.yml) and `publish` workflows.

That'd also be a good opportunity for a bit of clean-up.

It might be neatest for our scripts directory to strictly have a 1:1 mapping with the steps we have in our workflows.

Here's how our two workflows currently look...

```yaml
      - name: "Install dependencies"
        run: "scripts/install"
      - name: "Run linting checks"
        run: "scripts/check"
      - name: "Build package & docs"
        run: "scripts/build"
      - name: "Run tests"
        run: "scripts/test"
      - name: "Enforce coverage"
        run: "scripts/coverage"
```

```yaml
      - name: "Install dependencies"
        run: "scripts/install"
      - name: "Build package & docs"
        run: "scripts/build"
      - name: "Publish to PyPI & deploy docs"
        run: "scripts/publish"
```

And our scripts...

```shell
$ ls -1 ./scripts/
build
check
clean  # This doesn't exist in a workflow.
coverage  # Called into by `test` when running locally.
install
lint  # This doesn't exist in a workflow.
publish
test
unasync  # Called into by `check` and `lint`
```

Perhaps a neater approach would be...

* `scripts/install` - When running locally this should *always* start by removing any existing `venv`, `coverage` and `build` directories. We don't need `scripts/clean` then.
* `scripts/check` and `scripts/lint` - I'd suggest we just have `scripts/lint`. Run `scripts/lint --fix` to update files in-place.
* `scripts/unasync` - Should be a separate build stage. Run `scripts/unasync --fix` to update files in-place.

PR #779, opened 2023-08-10
Remove useless patterns from unasync.py

karosis88:
Some patterns became obsolete, but they were not deleted.

I also included a check for unused patterns.


karosis88:
Do we need a changelog?

tomchristie:
We don't, no.
It's a tooling change, not a user-facing behavioural change.

michaeloliverx (approved):

Kludex:
Should we move this to `scripts/`?

karosis88:
I think it's a job for https://github.com/encode/httpcore/issues/638

PR #794, opened 2023-09-02
Don't check all statuses when asquire connection

T-256:
# Summary

This improves performance by avoiding iterate over all members of `ConnectionPool._requests` when huge requests available in the pool.

# Checklist

### DIFF
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -154,9 +154,11 @@ class AsyncConnectionPool(AsyncRequestInterface):
 
         # If there are queued requests in front of us, then don't acquire a
         # connection. We handle requests strictly in order.
-        waiting = [s for s in self._requests if s.connection is None]
-        if waiting and waiting[0] is not status:
-            return False
+        for s in self._requests:
+            if s.connection is None:
+                if s is not status:
+                    return False
+                break
 
         # Reuse an existing connection if one is currently available.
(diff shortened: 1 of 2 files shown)


zanieb:
Hi @T-256 could you please start a discussion instead and describe the performance impact of this change?

T-256:
> Hi @T-256 could you please start a discussion instead and describe the performance impact of this change?

This PR is kind of neat code, So I think it doesn't need discussion.
Exmple:

```py
def test():
    iterator = range(10000)
    filtered = [x for x in iterator if x % 2000 == 0]  # here it iterates until 10000 even, that we don't need.
    if filtered and filtered[0] == 6000:  # and here only uses first member of list, so we can break it on first find.
        return
```
It's better to re-write it as:
```py
def test():
    iterator = range(10000)
    for x in iterator:
        if x % 2000 == 0:
            if x == 6000:
                return
            break
```
In first example we iterating over all members of range(0, 10000), but in the second example we won't iterate range(2000, 10000).

Another way here is use generator with cover of `StopIteration`, but I think current way is straighter forward way.

zanieb (approved):
While I agree this code makes more sense, code intended to change performance should generally include real world benchmarks indicating that it has the desired effect.

tomchristie:
I think we should close this off, pending a discussion that demos a valid motivation?

We've got clear guidelines, and they're helpful, so let's follow them.

PR #802, opened 2023-09-14
Handle asynchronous cancellations for HTTPConnection instances

### LINKED ISSUE #785: Handle asynchronous cancellations for HTTPConnection instances
- [x] Initially raised as [discussion #783](https://github.com/encode/httpcore/discussions/783)

The gist of the race condition is that when a web request is cancelled within a specific time window, it will degrade the state of the connection pool, leaving it unable to issue further requests. It also seems to create a deadlock.

It can be reproduced with the following code:

```python
import trio
import httpcore


TIMEOUT = {"read": 5, "write": 5, "connect": 5, "pool": 5}


async def send_request(client, seen_response):
    response = await client.request("GET", "http://127.0.0.1:8000", extensions={"timeout": TIMEOUT})

    # Print the first response that we see and set the "seen_response" flag.
    if not seen_response.is_set():
        print(response)
        seen_response.set()


async def main():
    async with httpcore.AsyncConnectionPool() as client:
        while 1:
                # This event is used to signal the first response we recieve.
                seen_response = trio.Event()

                # Kick off one hundred HTTP requests.
                for idx in range(100):
                    nursery.start_soon(send_request, client, seen_response)
                
                # As soon as we see a response we can cancel out of the nursery,
                # rather than allowing all tasks to complete.
                await seen_response.wait()
                nursery.cancel_scope.cancel()

trio.run(main)
```

This should produce a `httpcore.PoolTimeout` error on the second or third iteration, after which the loop enters into a deadlock.

karpetrosyan:
Motivated by https://github.com/encode/httpx/issues/947 🔥

Closes #785

This pull request just adds some logic to check if the connection was previously used or if it's just been created.
Now the example from #785 works fine(?).

It's very hard to explain where the problem actually is, but I think you can find the answer to that question in #785.

In simple words, the **client always thinks that there cannot be a connection that has not failed and is also not in the connecting process**. This is because every connection is attached to some request when it's created, and there are no cases when it can just hang in our connection pool except for asynchronous cancellations.

Now the `HTTPConnection` instances can be new if "._is_new" is set to true and failed if "._connect_failed" is set to true. Otherwise, if both of them are false, that means it's in the connecting process.

### DIFF
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -63,6 +63,7 @@ class AsyncHTTPConnection(AsyncConnectionInterface):
         self._connect_failed: bool = False
         self._request_lock = AsyncLock()
         self._socket_options = socket_options
+        self._is_new = True
 
     async def handle_async_request(self, request: Request) -> Response:
         if not self.can_handle_request(request.url.origin):
@@ -73,6 +74,7 @@ class AsyncHTTPConnection(AsyncConnectionInterface):
         async with self._request_lock:
             if self._connection is None:
                 try:
+                    self._is_new = False
                     stream = await self._connect(request)
 
                     ssl_object = stream.get_extra_info("ssl_object")
@@ -94,7 +96,7 @@ class AsyncHTTPConnection(AsyncConnectionInterface):
                             stream=stream,
                             keepalive_expiry=self._keepalive_expiry,
                         )
-                except Exception as exc:
+                except BaseException as exc:
                     self._connect_failed = True
                     raise exc
             elif not self._connection.is_available():
@@ -174,6 +176,8 @@ class AsyncHTTPConnection(AsyncConnectionInterface):
 
     def is_available(self) -> bool:
         if self._connection is None:
+            if self._is_new:
+                return True
             # If HTTP/2 support is enabled, and the resulting connection could
             # end up as HTTP/2 then we should indicate the connection as being
             # available to service multiple requests.
@@ -185,6 +189,8 @@ class AsyncHTTPConnection(AsyncConnectionInterface):
         return self._connection.is_available()
 
     def has_expired(self) -> bool:
+        if self._connect_failed:
+            return True
         if self._connection is None:
             return self._connect_failed
         return self._connection.has_expired()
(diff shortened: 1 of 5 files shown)


httpcore/_async/connection.py
```diff
@@ -94,7 +96,7 @@ async def handle_async_request(self, request: Request) -> Response:
                             stream=stream,
                             keepalive_expiry=self._keepalive_expiry,
                         )
-                except Exception as exc:
+                except BaseException as exc:
```
karosis88, line 99: This is to be sure that it works when we cancel the asynchronous task.

httpcore/_sync/connection.py
```diff
@@ -201,7 +205,9 @@ def is_closed(self) -> bool:
 
     def info(self) -> str:
         if self._connection is None:
-            return "CONNECTION FAILED" if self._connect_failed else "CONNECTING"
+            if self._is_new:
+                return "NEW CONNECTION"
+            return "CONNECTINION FAILED" if self._connect_failed else "CONNECTING"
```
zanieb, line 210:
```suggestion
            return "CONNECTION FAILED" if self._connect_failed else "CONNECTING"
```

httpcore/_async/connection.py
```diff
@@ -201,7 +205,9 @@ def is_closed(self) -> bool:
 
     def info(self) -> str:
         if self._connection is None:
-            return "CONNECTION FAILED" if self._connect_failed else "CONNECTING"
+            if self._is_new:
+                return "NEW CONNECTION"
+            return "CONNECTINION FAILED" if self._connect_failed else "CONNECTING"
```
zanieb, line 210:
```suggestion
            return "CONNECTION FAILED" if self._connect_failed else "CONNECTING"
```

httpcore/_sync/connection.py
karosis88: Thanks

karosis88:
Alternatively, we can add states for HTTPConnection instances (NEW, CONNECTING, FAILED), but I believe we should avoid that because we already have connecting states in HTTP11Connection instances.

karpetrosyan:
Now the connection has a state of connecting, even if it was just created and is not trying to connect somewhere. This pull request simply adds the new state for such connections.

T-256:
Duo to #785 closed, I think you can close this PR too.

T-256:
Now #785 closed, I think you can close this PR too.

PR #803, opened 2023-09-15
Skip the http2 cancellation tests if they fail.

### LINKED ISSUE #756: Ensure HTTP/2 connections are gracefully closed on async cancellation

karosis88:
# Summary

Now tests randomly fail, which in some places makes development hard (see #756). 
Local testing can randomly fail, and GitHub actions can also fail. 

For example, 3.11 tests fail, causing other actions to be interrupted.
Let's just skip these tests for now and remove the "@pytest.mark.xfail" flag in the pull request, which will resolve #756

# Checklist

### DIFF
--- a/tests/test_cancellations.py
+++ b/tests/test_cancellations.py
@@ -171,6 +171,7 @@ async def test_h11_timeout_during_response():
         assert conn.is_closed()
 
 
+@pytest.mark.xfail
 @pytest.mark.anyio
 async def test_h2_timeout_during_handshake():
     """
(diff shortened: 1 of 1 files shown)


tomchristie:
Yep. Flaky async cancellation tests *should* be marked `xfail`. 👍
And *this is okay*.

tomchristie (approved):

PR #807, opened 2023-09-17
Add support for Python 3.12

hugovk:
# Summary

The [second Python 3.12 release candidate](https://discuss.python.org/t/python-3-12-0rc2-final-release-candidate-released/33105?u=hugovk) is out! :rocket:

> ## Call to action
> 
> We strongly encourage maintainers of third-party Python projects to prepare their projects for 3.12 compatibilities during this phase, and where necessary publish Python 3.12 wheels on PyPI to be ready for the final release of 3.12.0.

Python 3.12.0 final is due out in two weeks: [2023-10-02](https://peps.python.org/pep-0693/).

See also https://dev.to/hugovk/help-test-python-312-beta-1508/

# Checklist

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -14,7 +14,7 @@ jobs:
       name: deploy
 
     steps:
-      - uses: "actions/checkout@v3"
+      - uses: "actions/checkout@v4"
       - uses: "actions/setup-python@v4"
         with:
           python-version: 3.8
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -14,13 +14,14 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
 
     steps:
-      - uses: "actions/checkout@v3"
+      - uses: "actions/checkout@v4"
       - uses: "actions/setup-python@v4"
         with:
           python-version: "${{ matrix.python-version }}"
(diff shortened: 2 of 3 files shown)


tomchristie (approved):
😊

karosis88:
Did you forget a changelog?

tomchristie:
Okay let me slow this down a little and be more precise.

First up, we should change the title of this PR...

tomchristie:
> Did you forget a changelog?

Good thing to check, thanks. No we don't need a CHANGELOG entry here.
We're not adding 3.12 support, we're including it in our CI tests.

hugovk:
This PR also added the Trove classifier for 3.12, which could be seen as adding 3.12 support?

I think it's a good time to add it: the final 3.12 RC3 is due out today, the full release in two weeks, and there should be no changes between now and then, and no ABI changes allowed.

tomchristie:
> This PR also added the Trove classifier for 3.12, which could be seen as adding 3.12 support?

Ah thats a good point, thanks @hugovk.

Yes a CHANGELOG entry noting that our package now explicitly indicates 3.12 support would be valid.

Related... anyone up for an equivalent pull request to `httpx`?

hugovk:
> > This PR also added the Trove classifier for 3.12, which could be seen as adding 3.12 support?
> 
> Ah thats a good point, thanks @hugovk.
> 
> Yes a CHANGELOG entry noting that our package now explicitly indicates 3.12 support would be valid.

Here we go! https://github.com/encode/httpcore/pull/810

> Related... anyone up for an equivalent pull request to `httpx`?

And here we go again! https://github.com/encode/httpx/pull/2854

PR #809, opened 2023-09-19
Version 1.0

tomchristie:
Proposal for `httpcore` version 1.0.

Our API is already stable, and I don't think we've any blockers that ought to prevent us issuing a 1.0 release.

The one change that I *am* suggesting in this pull request is that we switch to a more minimal set of default dependancies, by making the async case fully optional. We've already [done the work on making these packages not strictly required](https://github.com/encode/httpcore/pull/639). This proposal builds on that work by no longer installing `anyio`/`sniffio`/`trio` by default.

---

**TODO**:

- [x] Update CHANGELOG.md

---

*See also https://github.com/encode/httpx/issues/947*.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,11 +42,9 @@ socks = [
 ]
 trio = [
     "trio>=0.22.0,<0.23.0",
-    "sniffio==1.*",
 ]
 asyncio = [
     "anyio>=4.0,<5.0",
-    "sniffio==1.*",
 ]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,6 +60,7 @@ include = [
     "/httpcore",
     "/CHANGELOG.md",
     "/README.md",
+    "/tests"
 ]
 
 [tool.hatch.metadata.hooks.fancy-pypi-readme]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ socks = [
     "socksio==1.*",
 ]
 trio = [
-    "trio>=0.21.0,<0.23.0",
+    "trio>=0.21.0,<0.22.0",
     "sniffio==1.*",
 ]
 asyncio = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ classifiers = [
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
     "Topic :: Internet :: WWW/HTTP",
 ]
 dependencies = [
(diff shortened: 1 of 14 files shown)


karosis88:
Do we want https://github.com/encode/httpcore/discussions/800 after this release?

tomchristie:
Have followed up against that discussion.

karosis88 (commented):
We can change the error message in https://github.com/encode/httpcore/pull/809/files#diff-5bb436aadeb1bfe2f41078fe8d1a4db589eb3405df644e07ce5e26c7749da824R40 to something like `Running under trio requires the 'httpcore' to be installed with the 'trio' extra.`  because when the user has trio installed, the current message doesn't make any sense.

zanieb:
> because when the user has trio installed, the current message doesn't make any sense

Why would that error be displayed when trio is installed?

karosis88:
> > because when the user has trio installed, the current message doesn't make any sense
> 
> Why would that error be displayed when trio is installed?

When the trio is installed but the sniffio isn't

See: https://github.com/encode/httpcore/pull/809/files#diff-5bb436aadeb1bfe2f41078fe8d1a4db589eb3405df644e07ce5e26c7749da824R20-R24

tomchristie:
> We can change the error message

Resolved

pyproject.toml
```diff
@@ -42,6 +40,14 @@ http2 = [
 socks = [
     "socksio==1.*",
 ]
+trio = [
+    "trio>=0.21.0,<0.22.0",
```
karosis88, line 44: Is this necessary?

```suggestion
    "trio>=0.21.0",
```

karosis88 (approved):

pyproject.toml
agronholm: This depends on the AnyIO version used. AnyIO 4 will only work with Trio 0.22 or later while AnyIO 3 will only work with 0.21 or earlier. AnyIO itself has the `trio` extra which will constraint the version appropriately.

tomchristie:
I think we're good to go here... I'll plan on releasing this tomorrow.

karosis88:
Do we want to include #799 to 1.0.0?

tomchristie:
Ok.

karpetrosyan:
Release date :cold_face:

tomchristie:
Oops. (thanks)

agronholm:
On that note, how are you creating your GitHub releases? I don't see any automation.

tomchristie:
Copy the release notes. Yup some buttons do get pressed through the GitHub UI.
From that, the PyPI releases are automated from a GitHub action.

agronholm:
In case you're interested, here's how I handle things: https://github.com/agronholm/anyio/blob/master/.github/workflows/publish.yml
Pushing a tag makes the automation copy the release notes matching the tag, and make a GitHub release as well as publishing the sdist and wheels to PyPI.

PR #811, opened 2023-09-19
Handle async cancelled error explicitly

T-256:
# Summary

Discussed in https://github.com/encode/httpcore/discussions/805

# Checklist


Tunglies:
We'd need change `httpcore/_async/http11.py` and `httpcore/_sync/http11.py`. To ensure them match, run `./scripts/unasync`.

Tunglies (changes requested):
We'd need to change in both `httpcore/_async` and `httpcore/_sync` and ensure them match.

karosis88:
Is this solution respects such cases? 
> because the user may catch the keyboardinterrupt in his code and continue to use httpcore.ConnectionPool.

T-256:
> Is this solution respects such cases?
> 
> >

No, #812 will respect.

karosis88:
> No, #812 will respect.

Do we want to marge this PR when we know it will break programs that handle `BaseExceptions` like `KeyboardInterrupt`?

T-256:
> Do we want to marge this PR when we know it will break programs that handle `BaseExceptions` like `KeyboardInterrupt`?

Yes. because still some of cleanups doesn't handle `BaseException` in codebase:

- https://github.com/encode/httpcore/blob/94ffb33cf378e41662c90615458db9e42b43cad5/httpcore/_async/connection.py#L97-L100

- https://github.com/encode/httpcore/blob/94ffb33cf378e41662c90615458db9e42b43cad5/httpcore/_async/http2.py#L441-L453

- https://github.com/encode/httpcore/blob/94ffb33cf378e41662c90615458db9e42b43cad5/httpcore/_async/http2.py#L470-L482

- https://github.com/encode/httpcore/blob/94ffb33cf378e41662c90615458db9e42b43cad5/httpcore/_async/socks_proxy.py#L294-L297

- https://github.com/encode/httpcore/blob/94ffb33cf378e41662c90615458db9e42b43cad5/httpcore/_backends/anyio.py#L76-L79

- https://github.com/encode/httpcore/blob/94ffb33cf378e41662c90615458db9e42b43cad5/httpcore/_backends/trio.py#L78-L81

- plus, same things in `_sync` part.

And also notice few `BaseException` was introduced in https://github.com/encode/httpcore/pull/726, so before that we were not supporting `KeyboardInterrupt`:

> * [First](https://github.com/encode/httpcore/pull/726/files#diff-4a314fb3c706f5b1f148fe26dffa6ef7ab62eaaa4217b90ff5fd286b73d0087bR106), _Before_: no error handling, _After_: `BaseException` handling.
> * [Second](https://github.com/encode/httpcore/pull/726/files#diff-4a314fb3c706f5b1f148fe26dffa6ef7ab62eaaa4217b90ff5fd286b73d0087bR163), _Before_: `Exception` handling, _After_: `BaseException` handling.

httpcore/_synchronization.py
```diff
@@ -6,16 +6,25 @@
 
 from ._exceptions import ExceptionMapping, PoolTimeout, map_exceptions
 
+
+EXCEPTION_OR_CANCELLED = (Exception,)
+
```
T-256, line 11:
```suggestion
EXCEPTION_OR_CANCELLED = (Exception,)

```

httpcore/_synchronization.py
```diff
@@ -1,21 +1,32 @@
 import threading
 from types import TracebackType
-from typing import Optional, Type
+from typing import Optional, Type, Tuple
```
T-256, line 3:
```suggestion
from typing import Optional, Tuple, Type
```

T-256:
Should we consider convert exception handlings in https://github.com/encode/httpcore/pull/811#issuecomment-1727720109 to `EXCEPTION_OR_CANCELLED`?

httpcore/_synchronization.py
```diff
@@ -1,21 +1,37 @@
 import threading
 from types import TracebackType
-from typing import Optional, Type
+from typing import Optional, Tuple, Type, Union
 
 import sniffio
 
 from ._exceptions import ExceptionMapping, PoolTimeout, map_exceptions
 
+
```
T-256, line 9:
```suggestion
```

httpcore/_synchronization.py
```diff
@@ -1,21 +1,32 @@
 import threading
 from types import TracebackType
-from typing import Optional, Type
+from typing import Optional, Tuple, Type, Union
```
T-256, line 3:
```suggestion
from typing import Optional, Tuple, Type
```

httpcore/_synchronization.py
```diff
@@ -1,21 +1,32 @@
...
 
 try:
     import anyio
+
+    try:
+        import asyncio
+
+        EXCEPTION_OR_CANCELLED += (asyncio.CancelledError,)
+    except ImportError:
+        pass
```
T-256, line 29:
```suggestion
    except ImportError:  # pragma: nocover
        pass
```

PR #823, opened 2023-10-11
Ensure pool timeout is applied, even after attempts leading to ConnectionNotAvailable

### LINKED ISSUE #822: Pool timeout doesn't work
Discussion: #814

The pool timeout exception is now only raised when we are waiting for a connection from a pool and the pool is always full.

## Problem

When the response releases the connection, all the requests in the queue can take the same connection.
But in the AsyncHTTP11Connection class, when the `handle_async_request` is called, it checks if his state is correct. If the state is not NEW or IDLE, that means that the other request has stolen that connection, and it raises ConnectionNotAvailable.

Like so:
https://github.com/encode/httpcore/blob/8780c9cf11ccd5d0a5ddc307bd2b1ed8a682a030/httpcore/_async/http11.py#L78-L84

Because the queued request has this cycle, the program may hang indefinitely.

- waiting for a connection 
- gets the connection that is already in use (but the connection state has not been changed yet, so it thinks he is the only one who uses that connection)
- catches the `ConnectionNotAvailable` exception and tries again with the same pool timeout.

There is a code that can reproduce this issue (credits to @valsteen).

### Server

```python
import asyncio
from fastapi import FastAPI

app = FastAPI()

@app.get("/")
async def root():
    await asyncio.sleep(2)
```

### Client

```python
import time
import trio
from httpcore import AsyncConnectionPool

TIMEOUTS = {"read": 5, "write": 5, "pool": 5, "connect": 5}


async def read():
    start = time.time()
    try:
        response = await http.request(
            "GET", "http://127.0.0.1:8000", extensions={"timeout": TIMEOUTS}
        )
        if response.status // 100 != 2:
            raise RuntimeError
    except Exception as ex:
        print(f"failed: {ex.__class__}")

    print(f"time: {time.time()-start}")


async def main():
        for _ in range(100):
            nursery.start_soon(read)


if __name__ == "__main__":
    http = AsyncConnectionPool(
        max_connections=3,
        max_keepalive_connections=3,
    )
    trio.run(main)
```

valsteen:
# Summary

fixes #822 

This ensures the total time spent trying to acquire a connection is considered.

- both sync and async versions are fixed
- the first commit `add failing tests` contains a failing test case demonstrating the issue. The second commit fixes it
- as before, a pool timeout of `0` will succeed if a connection is immediately available

# Checklist

- [x] I've updated the documentation accordingly –  _the documentation does not need change_


karpetrosyan:
I believe that changes such as "quote to double quote" are unnecessary in this pull request.
We also want to keep things as simple as possible, so the time module should be removed along with the async_open_nursery function.

The implementation makes sense, but the test cases are overly complicated; we could probably write them much more simply.

karpetrosyan:
Let's just use the previous test case. 
This test case is complicated enough to be ignored. 
Also, I am not sure if these kinds of milisecond timeouts are reliable.

I have tried to write a simple test for these cases, but it seems like we should avoid that. 
The implementation makes sense to me; let's just add a changelog, and I think it should be ready.

valsteen:
@karpetrosyan I understand, let me check if it can be isolated/simplified

valsteen:
tests reverted and Changelog added

karpetrosyan:
This is great. Thank you @valsteen!

karpetrosyan (approved):

PR #846, opened 2023-11-07
Fix float("inf") timeouts

karpetrosyan:
Resolves #845

The issue is that our `AsyncEvent.wait` supports `float('inf')` timeout but `Event.wait` doesn't.

Simple demonstration:

Works
```python
import asyncio
from httpcore._synchronization import AsyncEvent

async def background(event: AsyncEvent):
    await asyncio.sleep(2)
    event.set()

async def main():
    event = AsyncEvent()
    asyncio.create_task(background(event))
    await event.wait(timeout=float("inf"))
    print('done')

asyncio.run(main())
```

Doesn't work
```python
import threading
import time
from httpcore._synchronization import Event

def background(event: Event):
    time.sleep(2) 
    event.set()

def main():
    event = Event()
    threading.Thread(target=background, args=[event]).start()
    event.wait(timeout=float("inf"))
    print('done')

main()  
```

Error:

```
Traceback (most recent call last):
  File "/home/test/programs/gitprojects/httpcore/tust.py", line 15, in <module>
    main()
  File "/home/test/programs/gitprojects/httpcore/tust.py", line 12, in main
    event.wait(timeout=float("inf"))
  File "/home/test/programs/gitprojects/httpcore/httpcore/_synchronization.py", line 229, in wait
    if not self._event.wait(timeout=timeout):
  File "/usr/lib/python3.10/threading.py", line 607, in wait
    signaled = self._cond.wait(timeout)
  File "/usr/lib/python3.10/threading.py", line 324, in wait
    gotit = waiter.acquire(True, timeout)
OverflowError: timestamp too large to convert to C _PyTime_t
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+
+- Fix `float("inf")` timeouts in `Event.wait` function. (#846)
+
 ## 1.0.1 (November 3rd, 2023)
 
 - Fix pool timeout to account for the total time spent retrying. (#823)
--- a/httpcore/_synchronization.py
+++ b/httpcore/_synchronization.py
@@ -226,6 +226,8 @@ class Event:
         self._event.set()
 
     def wait(self, timeout: Optional[float] = None) -> None:
+        if timeout == float("inf"):  # pragma: no cover
+            timeout = None
         if not self._event.wait(timeout=timeout):
             raise PoolTimeout()  # pragma: nocover
 


tomchristie:
How would you feel about us starting by just reverting #823 for a 1.0.2 release?

I feel like I'd like a slower review onto #822 / #823 (which isn't critical).

karpetrosyan:
> How would you feel about us starting by just reverting https://github.com/encode/httpcore/pull/823 for a 1.0.2 release?

This means that we would not support pool timeout, which I believe is critical.

Also, look at #814. The previous behavior was very unstable and unpredictable.

tomchristie:
So...

We have a stable version with a minor issue that some timeouts can take longer than expected in some circumstances.
We've "upgraded" that to a more complex version that introduces an overflow error.

My strong preference in a case like that is to immediately roll back the change, re-release, and then make sure we're slowing down on future changes.

I'm up for re-reviewing #822 and #823 (plus this change) but I'd much rather be able to do that in a way that's not time-pressured.

karpetrosyan:
It's up to you. 

> My strong preference in a case like that is to immediately roll back the change, re-release, and then make sure we're slowing down on future changes.

I would agree if the issue was not small enough to be fixed immediately in the next release.

tomchristie (approved):
👍 Okay, yep. You've probs got a good call on this one.

Would you also like to roll up a 1.0.2 release for this?

(thanks)

karpetrosyan:
> Would you also like to roll up a 1.0.2 release for this?

 Yes I would

PR #849, opened 2023-11-22
Fix bug with async socks proxy and tracing.

cono:
# Summary

If you are using socks proxy, asynchronous client and tracing, it fails with the following error:

```
  File "/opt/.../3rdparty_python_httpcore/site-packages/httpcore/_async/socks_proxy.py", line 231, in handle_async_request
    with Trace("connect_tcp", logger, request, kwargs) as trace:
  File "/opt/.../3rdparty_python_httpcore/site-packages/httpcore/_trace.py", line 50, in __enter__
    self.trace(f"{self.name}.started", info)
  File "/opt/.../3rdparty_python_httpcore/site-packages/httpcore/_trace.py", line 33, in trace
    raise TypeError(
TypeError: If you are using a synchronous interface, the callback of the `trace` extension should be a normal function instead of an asynchronous function.
```

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 1.0.2 (November 10th, 2023) 
+## 1.0.3 (November 22th, 2023)
+
+- Fix copy paste bug with socks proxy and tracing. (#849)
+
+## 1.0.2 (November 10th, 2023)
 
 - Fix `float("inf")` timeouts in `Event.wait` function. (#846)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
-- Fix copy paste bug with socks proxy and tracing. (#849)
+- Fix trace extension when used with socks proxy. (#849)
 
 ## 1.0.2 (November 10th, 2023)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 1.0.3 (November 22th, 2023)
+## Unreleased
 
 - Fix copy paste bug with socks proxy and tracing. (#849)
 
--- a/httpcore/__init__.py
+++ b/httpcore/__init__.py
@@ -130,7 +130,7 @@ __all__ = [
     "WriteError",
 ]
 
-__version__ = "1.0.2"
+__version__ = "1.0.3"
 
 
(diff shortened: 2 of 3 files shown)


karpetrosyan:
How can I reproduce this issue locally?

cono:
@karpetrosyan use async interface and socks5 proxy.

karpetrosyan:
Yes, I see the problem. 
We appear to have accidentally used the synchronous context manager in the asynchronous code.
Thank you :pray:

cono:
@karpetrosyan added

zanieb (commented):
Can you add test coverage for this?

I'm also not sure we should bump the version here, that's usually done in a separate pull request.

cono:
I wonder why person who added this code haven't covered this with tests.
So will leave it for you guys.
We workarounded this in production anyway, as your release process is not fast.

CHANGELOG.md
```diff
@@ -4,7 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 1.0.2 (November 10th, 2023) 
+## 1.0.3 (November 22th, 2023)
```
karpetrosyan, line 7:
```suggestion
## Unreleased
```

httpcore/__init__.py
```diff
@@ -130,7 +130,7 @@ def __init__(self, *args, **kwargs):  # type: ignore
     "WriteError",
 ]
 
-__version__ = "1.0.2"
+__version__ = "1.0.3"
```
karpetrosyan, line 133:
```suggestion
__version__ = "1.0.2"
```

karpetrosyan:
> I wonder why person who added this code haven't covered this with tests.

The code is actually covered, at least, because we have 100% code coverage.

> So will leave it for you guys.

If you have found a bug that has not been tested and you have considered opening the pull request to fix that bug, you **MUST** add a test case, especially when it was requested by the maintainer.

There are rules for how this project should be maintained. 
You cannot do just a part of the pull request and let the rest of the work be done by others.
If you had considered opening the pull request, you should either do it fully or don't do it at all.

Anyway, I think we don't need a test case for this pull request. Maybe we could add a rule to **unasync.py** so we can avoid this kind of situation.

If @zanieb is okay with that, we can skip the test case for this change.

Thanks for this pull request @cono :pray:

cono:
Guys, you have a bug in a library which crash application in particular circumstances, and isntead of accepting 2 line changes, you making discussion for 2 days about nothing. It's kinda like a joke to me.
Do whatever you want. "MUST" is too strong for me, I don't owe you anything. Good luck

CHANGELOG.md
```diff
@@ -4,7 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## 1.0.2 (November 10th, 2023) 
+## Unreleased
+
+- Fix copy paste bug with socks proxy and tracing. (#849)
```
```suggestion
- Fix trace extension when used with socks proxy. (#849)
```

tomchristie (approved):

PR #856, opened 2023-12-04
Remove expensive run time type casting

bdraco:
# Summary

Guard casting type subscripting with TYPE_CHECKING to avoid runtime construction each time `_receive_event` is called which is unexpectedly expensive

```python
import h11
from typing import TYPE_CHECKING, Union, Type, cast
import timeit


def test_with_type_checking():
    # mypy fails to narrow the type in the above if statement above
    event = None
    if TYPE_CHECKING:  # pragma: no cover
        event = cast(Union[h11.Event, Type[h11.PAUSED]], event)  # pragma: no cover


def test_without_type_checking():
    event = None
    event = Union[h11.Event, Type[h11.PAUSED]], event


without_type_checking = timeit.timeit(test_without_type_checking)
with_type_checking = timeit.timeit(test_with_type_checking)


print(f"Time for without: {without_type_checking}")
print(f"Time for with: {with_type_checking}")

```

```
Time for without: 0.37302987500152085
Time for with: 0.02093412500107661
```

Screenshot 2023-12-04 at 11 15 24 AM
Screenshot 2023-12-04 at 11 15 14 AM

# Checklist

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,0 @@
-[report]
-exclude_lines =
-    # TYPE_CHECKING blocks are never executed during pytest run
-    if TYPE_CHECKING:
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,4 @@
+[report]
+exclude_lines =
+    # TYPE_CHECKING blocks are never executed during pytest run
+    if TYPE_CHECKING:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -14,10 +14,10 @@ jobs:
       name: deploy
 
     steps:
-      - uses: "actions/checkout@v3"
+      - uses: "actions/checkout@v4"
       - uses: "actions/setup-python@v4"
         with:
-          python-version: 3.7
+          python-version: 3.8
       - name: "Install dependencies"
         run: "scripts/install"
       - name: "Build package & docs"
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -14,13 +14,14 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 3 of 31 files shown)


T-256:
Seems reasonable! how can I reproduce?

bdraco:
I only saw it because it came up in the py-spy watching Home Assistant request for an enphase envoy. It shows up in a callgrind as well using cProfile.

bdraco:
Added a micro benchmark above

tomchristie (commented):
Yep, smart change. 👍
For review I `grep`'ed this and ensured that you've applied this to all the `cast` cases consistently.

An alternative here would be to *remove* the cast and use a `type: ignore`, should we consider that instead as a simpler alternative? (Or not, also okay?)

bdraco:
> Yep, smart change. 👍 For review I `grep`'ed this and ensured that you've applied this to all the `cast` cases consistently.
> 
> An alternative here would be to _remove_ the cast and use a `type: ignore`, should we consider that instead as a simpler alternative? (Or not, also okay?)

Happy to do either as its really a matter of preference. 

I'm used to guarding with `if TYPE_CHECKING` since thats what Home Assistant prefers for these [cases](https://github.com/home-assistant/core/blob/a8ca73a7dd033de0cc290e21675af168e4365c7c/homeassistant/core.py#L530), but some projects prefer ignores.

T-256:
> but some projects prefer ignores.

I do think this repo also prefers ignores over casting.
I'm unsure why `cast` added to this line but it's worth to use `type: ignore` instead at there.

T-256 (approved):
Thank you!

Pliner (approved):

bdraco:
Thanks

Issue #830, opened 2023-10-24
AsyncConnectionPool hangs future connections sometimes when a connection get cancelled.


fusiyuan2010:
In `async def handle_async_request(self, request: Request) -> Response:` there are such lines:
```
        async with self._pool_lock:
            self._requests.append(status)
            await self._close_expired_connections()
            await self._attempt_to_acquire_connection(status)
```

There is a probability that the coroutine got canceled after `self._requests.append(status)` but before `await self._attempt_to_acquire_connection`.  Since these few lines are not protected against canceling nor having cleaning up on exception, it may leave a status stayed in _requests forever, blocking all future `connection = await status.wait_for_connection(timeout=timeout)` which is a few lines after.

karpetrosyan:
Hi @fusiyuan2010!

Can you give a minimum example of how we can reproduce this issue?

fusiyuan2010:
@karpetrosyan Unfortunately, it's hard to give a working example that can easily reproduce the issue since the probability of this happening is low.  I can describe the scenario of how this can happen:

I have a typical web server serving requests. In the request handler,  I use this global AsyncConnectionPool object, which is shared across all requests, to call another service.  The current request handler can fail while calling HTTP to upstream service, for whatever reason but not related to this HTTP call itself, for example:
```
@app.get("/somea_api")
async def main():
    req = make_upstream_request("http://upstream-service")
    task_http_call = asyncio.create_task(global_client_pool.handle_async_request(req))
    main_task = asyncio.create_task(other_tasks())
    await asyncio.gather(task_http_call, main_task)
```
If main_task throws an exception at the exact time the handle_async_request function runs in the self._close_expired_connections(), the task_http_call is canceled with the current request left in the `self._requests` forever. This leaves this global pool in a bad state that all future calls to _attempt_to_acquire_connection will return False immediately.

karpetrosyan:
We currently have issues with our async cancellation support; also, catching each await and handling cacnel to avoid breaking the client state does not make sense to me. 
Perhaps we could shield all connection pool activities to ensure that the connection pool is never broken :thinking: .

See also: https://github.com/encode/httpcore/issues/785, https://github.com/encode/httpcore/pull/757

baconismycopilot:
I ran into this issue using httpx and finally reached the root cause, which is httpcore. [This snippet reliably recreates this issue. ](https://github.com/encode/httpx/issues/1171#issuecomment-1864841152)

rattrayalex:
[...]

PR #880, opened 2024-02-06
Safe async cancellations.

### LINKED ISSUE #785: Handle asynchronous cancellations for HTTPConnection instances

### LINKED ISSUE #861: Add `__repr__` for connection pools
- [ ] Initially raised as discussion https://github.com/encode/httpx/discussions/2966#discussioncomment-7921948

For example (proposed by @tomchristie):

```py
>>> client = httpx.Client()
>>> client.transport
<httpx.HTTPTransport [Connections: 0 active, 0 keep-alive. Requests: 0 active, 0 queued.]>
```

Ideally, this will be followed by a public API to access this information programatically.

(Note that I am filing this issue as requested, but do not expect to use it myself, as I am interested primarily in the programmatic access).

cc @robertcraigie

tomchristie:
Closes https://github.com/encode/httpcore/issues/830
Closes #785
Closes #861
Closes https://github.com/encode/httpx/issues/1171

Let's talk about what's going on here.

We've got an issue with handling async cancellations correctly, which needs some re-working in order to comprehensively resolve it. We're somewhat in contrast to either `urllib3` (sync) or `aiohttp` (async) here, because we're having to get both the thread-safe and the task-safe-plus-also-support-cancellations cases correct.

Currently we're at thread-safe plus task-safe, but missing correct handling of also allowing external cancellations at any point.

So then...

---

This pull request reworks the handling of the connection pool state. The state is a list of connections, a list of requests, and an association between each request and the connection that has been selected to handle it.

The fundamental change in this pull request is that the management of the pool state has been re-worked so that it does not include I/O.

https://github.com/encode/httpcore/blob/cda040d4f6f01b73c97fbf8cd767cff21e820b83/httpcore/_async/connection_pool.py#L228-L299

This ensures that updating the state of the connection pool is always atomic.

---

As part of resolving this, also adds connection pool `__repr__` to demonstate the state more clearly.

The look like so...

```python
<httpcore.ConnectionPool [Requests: 10 active, 3 queued | Connections: 2 active, idle]>
```

### DIFF
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -208,11 +208,12 @@ class AsyncConnectionPool(AsyncRequestInterface):
                 else:
                     break
 
-        except BaseException:
+        except BaseException as exc:
             async with self._pool_lock:
                 self._requests.remove(pool_request)
                 closing = self._assign_requests_to_connections()
             await self._close_connections(closing)
+            raise exc from None
 
         assert isinstance(response.stream, AsyncIterable)
         return Response(
@@ -258,8 +259,8 @@ class AsyncConnectionPool(AsyncRequestInterface):
         queued_requests = [
             request for request in self._requests if request.connection is None
         ]
-        for request in queued_requests:
-            origin = request.url.origin
+        for pool_request in queued_requests:
+            origin = pool_request.request.url.origin
             avilable_connections = [
                 connection
                 for connection in self._connections
@@ -268,12 +269,12 @@ class AsyncConnectionPool(AsyncRequestInterface):
             if avilable_connections:
                 # log: "reusing existing connection"
                 connection = avilable_connections[0]
-                request.assign_to_connection(connection)
+                pool_request.assign_to_connection(connection)
             elif len(self._connections) < self._max_connections:
                 # log: "creating new connection"
                 connection = self.create_connection(origin)
                 self._connections.append(connection)
-                request.assign_to_connection(connection)
+                pool_request.assign_to_connection(connection)
             else:
                 idle_connections = [
                     connection
@@ -287,7 +288,7 @@ class AsyncConnectionPool(AsyncRequestInterface):
                 # log: "creating new connection"
                 connection = self.create_connection(origin)
                 self._connections.append(connection)
-                request.assign_to_connection(connection)
+                pool_request.assign_to_connection(connection)
 
         return closing_connections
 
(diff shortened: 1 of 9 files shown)


httpcore/_async/connection_pool.py
```diff
@@ -220,116 +169,149 @@ async def handle_async_request(self, request: Request) -> Response:
...
             headers=response.headers,
-            content=ConnectionPoolByteStream(response.stream, self, status),
+            content=PoolByteStream(
+                stream=response.stream, pool_request=pool_request, pool=self
+            ),
             extensions=response.extensions,
         )
 
-    async def response_closed(self, status: RequestStatus) -> None:
+    def _assign_requests_to_connections(self) -> List[AsyncConnectionInterface]:
```
tomchristie, line 227: This is where we mange the state of the connection pool, entirely within a non-I/O block. The `async` case cannot have cancellations or context-switches midway through the state management. The `sync` case is explicitly guarded with a thread lock.

Review is probably best made against the module as a whole, rather than the diff.

https://github.com/encode/httpcore/blob/f91789652485d5ec7651458bc20a9164ff56dbd6/httpcore/_async/connection_pool.py

rattrayalex:
FYI, a user reports that this branch does indeed fix issues they saw in the OpenAI Python client library: https://github.com/openai/openai-python/issues/1059#issuecomment-1936333824

tomchristie:
Fantastic, thanks for the feedback.

karpetrosyan:
Amazing work @tomchristie, this PR rsolves many problems..

karpetrosyan (approved):
I guess we also need a changelog here

PR #882, opened 2024-02-11
Support connection `Upgrade` and `CONNECT`.

### LINKED ISSUE #872: Support connection `Upgrade` and `CONNECT`.
- [x] Initially raised as discussion #871

In cases of `CONNECT` and `Upgrade` requests, it is necessary to handle the incoming data (`h11.Connection.trailing_data`).

https://h11.readthedocs.io/en/latest/api.html?highlight=trailing_data#switching-protocols

However, accessing `trailing_data` in HTTPCore requires a very lengthy property access.

```python
with httpcore.ConnectionPool() as http:
    with http.stream("GET", url, headers=headers) as response:
        # Get the trailing data.
        trailing_data, _ = response.stream._stream._connection._h11_state.trailing_data
```

Making this accessible through the public API would be beneficial for developers using the `"network_stream"`.

Related HTTPCore documentation: https://www.encode.io/httpcore/extensions/#network_stream

Here is what tomchristie suggested in the discussion. https://github.com/encode/httpcore/discussions/871#discussioncomment-8223419

> We could then either...
> 
> * Make the  available through a publicly documented response extension.`trailing_data`
> * Ensure that the  _is_ returned by the network stream, on the first .`trailing_data``.read()`
> 
> The second one of these is neatest from the user-perspective.
> 
> We'd probably implement that as a proxy class onto the underlying network stream, that's able to additionally deal with pushing the trailing data in the event back onto the stream...`h11`
> 
> ```python
> # This kinda thing...
> class UpgradeNetworkStream():
>     def __init__(self, leading_data, network_stream):
>         # We need to push any data that's already been read back onto the stream.
>         self._leading_data = leading_data
>         self._network_stream = network_stream
> 
>     def read(...)
>         if self._leading_data:
>             initial = self._leading_data
>             self._leading_data = b''
>             return initial
>         else:
>             return self._network_stream.read()
> ```

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/httpcore/issues/872">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/httpcore/issues/872/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

MtkN1:
# Summary

- Fixes #872

# Checklist

### DIFF
--- a/httpcore/_async/http11.py
+++ b/httpcore/_async/http11.py
@@ -115,6 +115,12 @@ class AsyncHTTP11Connection(AsyncConnectionInterface):
                     headers,
                 )
 
+            # FIXME: encode/httpcore#872
+            if (status == 101) or (
+                (request.method == b"CONNECT") and (200 <= status < 300)
+            ):
+                raise AssertionError("protocol switches")
+
             return Response(
                 status=status,
                 headers=headers,
--- a/httpcore/_sync/http11.py
+++ b/httpcore/_sync/http11.py
@@ -115,6 +115,12 @@ class HTTP11Connection(ConnectionInterface):
                     headers,
                 )
 
+            # FIXME: encode/httpcore#872
+            if (status == 101) or (
+                (request.method == b"CONNECT") and (200 <= status < 300)
+            ):
+                raise AssertionError("protocol switches")
+
             return Response(
                 status=status,
                 headers=headers,


MtkN1:
I have created a draft pull request to begin the work!
First, `Upgrade`/`CONNECT` (incorrectly) raises an exception.

My understanding is that the following work needs to be done...

1. Implement a network stream that can handle trailing data like this advice (https://github.com/encode/httpcore/discussions/871#discussioncomment-8223419)
2. Generates its network stream at the step that raised the exception
4. Switch variables to be stored in `network_stream`.
    https://github.com/encode/httpcore/blob/79fa6bf0dfcf3820d1ae7e52a2d268f33022c5a4/httpcore/_async/http11.py#L125

@tomchristie  I would appreciate any advice you can give me!

tomchristie:
Ah fantastic thanks, shall we start with a test?...

1. There's an existing `Upgrade` test here, that could be replicated into a new `test_upgrade_with_trailing_data` test case.

https://github.com/encode/httpcore/blob/79fa6bf0dfcf3820d1ae7e52a2d268f33022c5a4/tests/_async/test_http11.py#L238
 
I think the next step after that will be...

2. We probably want to change `receive_response_headers` so that it also returns the `trailing_data` as part of the response info, include the trailing data in the raised exception.

And then...

3. We'll want an `HTTPUpgradeNetworkStream(SyncNetworkStream)` class. This have an `__init__` method taking a network stream and trailing data. Most of the methods should delegate to the underlying network stream, except `read` which should initially return the trailing data. We'll return that in the extensions for the `Upgrade`/`CONNECT` cases.

MtkN1:
Thank you for your guidance 🙏 I'll give it a try.

MtkN1:
Work completed, I marked PR as ready. Could you please review this?

- Added `HTTP11UpgradeStream`.
  - The `read()` method allows trailing data to be received.
  - Other methods bypass the original NetworkStream.
- For `Upgrade` and `CONNECT`, switch the `network_stream` extension to `HTTP11UpgradeStream`.
- Added tests. 100% coverage.
- I confirm that this fix resolves my discussion (https://github.com/encode/httpcore/discussions/871#discussion-6123741) and httpx-ws issue (https://github.com/frankie567/httpx-ws/issues/7#issue-1463219741).
- I am not adding documentation because this change would be the behavior the user expects.

tests/_async/test_http11.py
```diff
@@ -265,10 +265,65 @@ async def test_http11_upgrade_connection():
...
+
+            assert network_stream._leading_data != b""
+            content = await network_stream.read(max_bytes=3)
+            assert content == b"foo"
+            content = await network_stream.read(max_bytes=3)
+            assert content == b"bar"
+
+            assert network_stream._leading_data == b""
+            content = await network_stream.read(max_bytes=3)
+            assert content == b"baz"
```
tomchristie, line 318: Test is great.

Let's not assert on the private state here, the read/content is enough.

tests/_async/test_http11.py
```diff
@@ -265,10 +265,65 @@ async def test_http11_upgrade_connection():
         ) as response:
             assert response.status == 101
             network_stream = response.extensions["network_stream"]
+            assert network_stream._leading_data == b""
```
tomchristie, line 268:
```suggestion
```

tomchristie (commented):
This one is a fantastic fix, thanks. ✨
Minor suggestion wrt. the tests.

---

Later follow-ups we can consider...

* Should all HTTP requests include this behaviour or only `Upgrade`/`CONNECT`?
* Should all HTTP requests return `network_stream` or only `Upgrade`/`CONNECT`?
* What might a nice `__repr__` be here?

(I'm just jotting these down here for reference)

httpcore/_async/http11.py
```diff
@@ -340,3 +353,34 @@ async def aclose(self) -> None:
             self._closed = True
             async with Trace("response_closed", logger, self._request):
                 await self._connection._response_closed()
+
+
+class HTTP11UpgradeStream(AsyncNetworkStream):
```
T-256, line 358: - is this only for `Upgrade`?
- is this only for http1.1?
- how about adding `Async` prefix
- Perhaps could move it to `httpcore._backends` into `sync` and `auto` submodules.

T-256, line 358:
```suggestion
class AsyncHTTP11TrailingStream(AsyncNetworkStream):
```

tests/_async/test_http11.py
MtkN1: Understood. I have modified it as suggested!

httpcore/_async/http11.py
MtkN1: Should I prefix it with Async?  The `HTTP11ConnectionByteStream` class in the same `http11.py` file has no prefix.
A similar pattern can be seen in the utility classes of other files.

tomchristie: Yep. The `scripts/unasync` script will strip that portion in the corresponding `_sync` code.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+
+- Support connection Upgrade and CONNECT. (#882)
```
```suggestion
- Fix support for connection Upgrade and CONNECT when some data in the stream has been read. (#882)
```

httpcore/_async/http11.py
MtkN1: Added. `scripts/unasync` is great all around!

tests/_async/test_http11.py
```diff
@@ -269,6 +269,57 @@ async def test_http11_upgrade_connection():
...
+            network_stream = response.extensions["network_stream"]
+
+            content = await network_stream.read(max_bytes=3)
+            assert content == b"foo"
+            content = await network_stream.read(max_bytes=3)
+            assert content == b"bar"
+            content = await network_stream.read(max_bytes=3)
+            assert content == b"baz"
+
+            # Lazy tests for HTTP11UpgradeStream
```
T-256, line 316:
```suggestion
            # Lazy tests for AsyncHTTP11UpgradeStream
```

T-256 (approved):
LGTM, Thank You!

tests/_async/test_http11.py
MtkN1: Oops, thank you!

tomchristie (approved):

MtkN1:
Thank you for your approval 🚀 It is a pleasure to contribute to this project.

PR #884, opened 2024-02-12
[DEV] Update requirements.txt

T-256:
# Summary
- Updated `werkzeug` pinned issue
- Dropped `types-certifi`, Also see https://github.com/encode/httpx/pull/3015
- Assigned `twine` version

# Checklist


requirements.txt
```diff
@@ -10,15 +10,14 @@ jinja2==3.1.3
 
 # Packaging
 build==1.0.3
-twine
+twine==4.0.2
```
T-256, line 13:
```suggestion
twine==5.0.0
```

tomchristie (approved):

PR #885, opened 2024-02-12
Group dependencies on dependabot updates

T-256:
# Summary
Ref: https://github.com/encode/httpx/pull/3054
# Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,10 @@
-
 version: 2
 updates:
   - package-ecosystem: "pip"
     directory: "/"
     schedule:
       interval: "monthly"
+    groups:
+      python-packages:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,7 +1,10 @@
...
 version: 2
 updates:
   - package-ecosystem: "pip"
     directory: "/"
     schedule:
       interval: "monthly"
+    groups:
+      python-packages:
+        patterns:
+          - "*"
```
T-256, line 10:
```suggestion
          - "*"
        exclude-patterns:
          - "werkzeug"
```

Pliner (approved):

tomchristie:
Thanks @T-256, @Pliner

PR #887, opened 2024-02-13
Ruff linter: Unexcluded `_sync` folders, Used default `line-length`

T-256:
# Summary
Same as what's done on httpx: https://github.com/encode/httpx/pull/2922

# Checklist


pyproject.toml
```diff
@@ -96,26 +96,27 @@ filterwarnings = ["error"]
...
+    "httpcore/*",
+    "tests/*",
 ]
-include = ["httpcore/*", "tests/*"]
 
 [tool.ruff]
 exclude = [
     "httpcore/_sync",
     "tests/_sync",
 ]
```
T-256, line 111: Why these folders excluded?

T-256: https://github.com/encode/httpcore/pull/887/commits/2e95f26a2b787063549f5ad6042099cf1ac7b3de

tomchristie: Because they're auto-generated... they mirror the async codebase.

unasync.py
```diff
@@ -5,33 +5,36 @@
...
+    ("handle_async_request", "handle_request"),
+    ("aclose", "close"),
+    ("aiter_stream", "iter_stream"),
+    ("aread", "read"),
+    ("asynccontextmanager", "contextmanager"),
+    ("__aenter__", "__enter__"),
+    ("__aexit__", "__exit__"),
+    ("__aiter__", "__iter__"),
+    ("@pytest.mark.anyio", "# unasync anyio"),
+    ("@pytest.mark.trio", "# unasync trio"),
```
tomchristie, line 28: Let's not change these, keep the PR focused.

unasync.py
```diff
@@ -5,33 +5,36 @@
...
-    ('__aenter__', '__enter__'),
-    ('__aexit__', '__exit__'),
-    ('__aiter__', '__iter__'),
-    ('@pytest.mark.anyio', ''),
-    ('@pytest.mark.trio', ''),
-    ('AutoBackend', 'SyncBackend'),
+    (
+        "from .._backends.auto import AutoBackend",
+        "from .._backends.sync import SyncBackend",
+    ),
```
T-256, line 11:
```suggestion
    ('from .._backends.auto import AutoBackend', 'from .._backends.sync import SyncBackend'),
```

pyproject.toml
```diff
@@ -96,22 +96,26 @@ filterwarnings = ["error"]
...
-include = ["httpcore/*", "tests/*"]
-
-[tool.ruff]
-exclude = [
-    "httpcore/_sync",
-    "tests/_sync",
+include = [
+    "httpcore/*",
+    "tests/*",
```
T-256, line 104:
```suggestion
    "scripts/*",
    "tests/*",
```

pyproject.toml
```diff
@@ -96,26 +96,26 @@ filterwarnings = ["error"]
...
-max-line-length = 120
-
-[tool.ruff.isort]
+[tool.ruff.lint.isort]
 combine-as-imports = true
+
+[tool.ruff.lint.per-file-ignores]
+# needed for unasync.py
+"httpcore/_sync/*" = ["I001"]
+"tests/_sync/*" = ["I001"]
```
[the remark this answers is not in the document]
karpetrosyan: Do we really need to avoid excluding sync directories? When we need to format/lint them? they are auto generated.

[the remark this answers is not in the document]
T-256: They are pre-generated. it means currently, they are already available in sdist, wheels and repository.
It doesn't look good that readers navigate through unformatted sources after installation.
For example, after install the `httpcore`, I'll navigate to `_sync` folder's contents and open a single file, then I'll see linter/formatter warnings in my IDE.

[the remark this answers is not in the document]
T-256: They are pre-generated. it means currently, they are already available in sdist, wheels and repository.
It doesn't look good that readers navigate through unformatted sources after installation.
For example, after installation, I'll navigate to `site-packages/httpcore` then in `_sync` folder's contents I open a single file, then I'll see linter/formatter warnings in my IDE.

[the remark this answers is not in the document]
T-256: When it's possible to have them, why excluding them?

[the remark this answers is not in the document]
T-256: > we don't know how many additional rules may be broken when introducing a new feature. We need add new rules that break sync files into the configuration every time, which makes the maintenance process more difficult.

The codebase is almost stable, I think the only violation rule would be import sorting violation. by checking `unasync.py` script I don't except new violation happening when adding new feature other than I001 (and F811 which can be in-line ignoring this without changing `pyproject.toml`)

IMO, we could give it a try, since always doable in future to revert this change.

PR #888, opened 2024-02-14
Add `target` request extension

karpetrosyan:
Refs: https://github.com/encode/httpx/discussions/2949#discussioncomment-8466736

This PR adds a target extension, which can be used to force HTTP target to a specific bytes-encoded string.

Check out https://developer.mozilla.org/en-US/docs/Web/HTTP/Messages#request_line to read more about HTTP message structure and what HTTP target is.


karpetrosyan:
I thought it is more natural to have this kind of extension logic in the httpcore package + all the transports will support it.

tomchristie:
How does this (does this?) effect the proxies?

tomchristie:
So here's where I was expecting we'd implement this...

https://github.com/encode/httpx/blob/3faa4a8f2e0d406167b913bbfd3afab087662d03/httpx/_transports/default.py#L222-L227

https://github.com/encode/httpx/blob/3faa4a8f2e0d406167b913bbfd3afab087662d03/httpx/_transports/default.py#L362-L367

Thoughts?

karpetrosyan:
I don't see any differences, If you like to see this at the httpx layer, I can close this one and open a new PR.

But, is there anything wrong with this one? Passing the target extension directly to the H11 target makes more sense to me.

tomchristie:
> But, is there anything wrong with this one?

You're probably right that it's sensible to implement this directly in `httpcore`, since all the other extensions we support are also there. Good call.

I think the neatest way is implementing this elsewhere...

https://github.com/encode/httpcore/blob/908013ca0d70afd44d75cb18e28b94aca5983611/httpcore/_models.py#L354

```python
if 'target' in extensions:
    self.url = URL(
        scheme=self.url.scheme,
        host=self.url.host,
        port=self.url.port,
        target=extensions['target'],
    )
```

That'll update the target once, and be passed through everywhere.

Reasonable?

karpetrosyan:
> Reasonable?

Much neater, thanks.

docs/extensions.md
```diff
@@ -166,6 +166,22 @@ response = httpcore.request(
 )
 ```
 
+### `"target"`
+
+The target that is used as the HTTP target instead of the URL path.
```
tomchristie, line 171:
```suggestion
The target that is used as [the HTTP target instead of the URL path](https://datatracker.ietf.org/doc/html/rfc2616#section-5.1.2).

This enables support constructing requests that would otherwise be unsupported. In particular...

* Forward proxy requests using an absolute URI.
* Tunneling proxy requests using `CONNECT` with hostname as the target.
* Server-wide `OPTIONS *` requests.
```

tomchristie:
Looks good, thanks.

I'd suggest also updating the `CONNECT` requests docs...

https://github.com/encode/httpcore/blob/908013ca0d70afd44d75cb18e28b94aca5983611/docs/extensions.md?plain=1#L212-L219

Since we can now express this with...

```python
url = "http://127.0.0.1:8080"
extensions = {"target: "http://www.example.com"}
with httpcore.stream("CONNECT", url, extensions=extensions) as response: 
```

Then, yep let's get this in.

tomchristie (approved):
Made a couple of additional docs changes on top of this.
Fantastic addition, thanks @karpetrosyan. 🎉

PR #895, opened 2024-03-01
Fix Ruff v0.3.0 violations

T-256:
Needed for #894

### DIFF
--- a/scripts/check
+++ b/scripts/check
@@ -8,7 +8,7 @@ export SOURCE_FILES="httpcore tests"
 
 set -x
 
-${PREFIX}ruff check --show-source $SOURCE_FILES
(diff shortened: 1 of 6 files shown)


scripts/check
```diff
@@ -8,7 +8,7 @@ export SOURCE_FILES="httpcore tests"
 
 set -x
 
-${PREFIX}ruff check --show-source $SOURCE_FILES
 ${PREFIX}ruff format $SOURCE_FILES --diff
 ${PREFIX}mypy $SOURCE_FILES
+${PREFIX}ruff check $SOURCE_FILES
```
T-256, line 13: chosen to be same as httpx

tomchristie (approved):

PR #901, opened 2024-03-19
really drop python<=3.7 support

kloczek:
# Summary
Filter all code over `pyupgrade --py38`.

# Checklist


tomchristie:
Let's close this in it's current state.

PR #920, opened 2024-06-09
Fix the error in generating the idle connection list

zhanghanyun:
Fix the error in generating the idle connection list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
+- Fix generate the idle connection list. (#920)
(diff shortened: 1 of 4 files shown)


tomchristie:
We currently have a test case `test_connection_pool_with_no_keepalive_connections_allowed` which deals with the `max_keepalive_connections=0` case.

Is it worth including `test_connection_pool_with_one_keepalive_connections_allowed` and `test_connection_pool_with_two_keepalive_connections_allowed` test cases? Would those test cases validate this change?

tomchristie:
Okay, let's just take this as-is... it's self-evidently an improvement.
If you could add an entry to [the CHANGELOG](https://github.com/encode/httpcore/blob/master/CHANGELOG.md) and then let's get this merged.

PR #922, opened 2024-06-10
Use `asyncio` synchronization instead of `anyio`

MarkusSintonen:
# Summary

First PR in series of optimizations to improve performance of `httpcore` and even reach performance levels of `aiohttp`.
Related discussion https://github.com/encode/httpx/issues/3215#issuecomment-2157885121

**Previously:**
old1

**With PR:**
new1

Request latency is more stable and the overall duration of the included benchmark improves by 3.5x.

# Checklist


tomchristie:
@MarkusSintonen Thanks so much for your work on this. Could we isolate the benchmarking into a separate PR?

MarkusSintonen:
> @MarkusSintonen Thanks so much for your work on this. Could we isolate the benchmarking into a separate PR?

Added here https://github.com/encode/httpcore/pull/923

httpcore/__init__.py
```diff
@@ -130,7 +130,7 @@ def __init__(self, *args, **kwargs):  # type: ignore
     "WriteError",
 ]
 
-__version__ = "1.0.5"
+__version__ = "1.0.6"
```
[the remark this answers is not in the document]
MarkusSintonen: Done

httpcore/_synchronization.py
```diff
@@ -83,10 +118,7 @@ async def __aexit__(
         exc_value: Optional[BaseException] = None,
         traceback: Optional[TracebackType] = None,
     ) -> None:
-        if self._backend == "trio":
-            self._trio_lock.release()
-        elif self._backend == "asyncio":
-            self._anyio_lock.release()
+        cast(_LockProto, self._lock).release()
```
T-256, line 121: use `#type: ignore` to prevent runtime calls.

httpcore/_synchronization.py
```diff
@@ -18,15 +30,30 @@
...
     try:
         import sniffio
     except ImportError:  # pragma: nocover
-        environment = "asyncio"
+        environment: AsyncBackend = "asyncio"
     else:
-        environment = sniffio.current_async_library()
+        environment = cast(AsyncBackend, sniffio.current_async_library())
 
     if environment not in ("asyncio", "trio"):  # pragma: nocover
```
T-256, line 58: is it worth to renaming `environment` to `backend` now?

httpcore/_synchronization.py
```diff
@@ -53,28 +90,26 @@ class AsyncLock:
...
-        if self._backend == "trio":
-            self._trio_lock = trio.Lock()
-        elif self._backend == "asyncio":
-            self._anyio_lock = anyio.Lock()
+        if current_async_backend() == "trio":
+            lock: _LockProto = trio.Lock()
+        else:
+            # Note: asyncio.Lock has better performance characteristics than anyio.Lock
+            # https://github.com/encode/httpx/issues/3215
+            lock = asyncio.Lock()
```
T-256, line 105:
```suggestion
            lock: _LockProto = asyncio.Lock()
```

httpcore/_synchronization.py
```diff
@@ -112,117 +144,110 @@ def __exit__(
...
-        self._backend = current_async_library()
+        self._backend = current_async_backend()
         if self._backend == "trio":
-            self._trio_event = trio.Event()
-        elif self._backend == "asyncio":
-            self._anyio_event = anyio.Event()
+            event: _EventProto = trio.Event()
+        else:
+            # Note: asyncio.Event has better performance characteristics than anyio.Event
+            event = asyncio.Event()
```
T-256, line 159:
```suggestion
            event: _EventProto = asyncio.Event()
```

httpcore/_synchronization.py
```diff
@@ -112,117 +144,110 @@ def __exit__(
...
-        if not self._backend:
-            self.setup()
-
-        if self._backend == "trio":
-            self._trio_event.set()
-        elif self._backend == "asyncio":
-            self._anyio_event.set()
+        if (event := self._event) is None:
+            event = self.setup()
+        event.set()
```
T-256, line 166: same as `AsyncLock.__aenter__`

httpcore/_synchronization.py
```diff
@@ -112,117 +144,110 @@ def __exit__(
...
+    @staticmethod
+    async def _asyncio_shield(
+        shielded: Callable[[], Coroutine[Any, Any, None]],
+    ) -> None:  # pragma: nocover
+        inner_task = asyncio.create_task(shielded())
+        try:
+            await asyncio.shield(inner_task)
+        except (asyncio.exceptions.CancelledError, asyncio.CancelledError):
+            # Let the inner_task to complete as it was shielded from the cancellation
+            await inner_task
```
T-256, line 250: No need instantiation, so we can convert to function then:
```suggestion
async def async_cancel_shield(shielded: Callable[[], Coroutine[Any, Any, None]]) -> None:
    if current_async_backend() == "trio":
        with trio.CancelScope(shield=True):
            await shielded()
    elif sys.version_info < (3, 11):  # pragma: nocover
        with anyio_shield(shield=True):
            await shielded()
    else:
        inner_task = asyncio.create_task(shielded())
        try:
            await asyncio.shield(inner_task)
        except (asyncio.exceptions.CancelledError, asyncio.CancelledError):
            # Let the inner_task to complete as it was shielded from the cancellation
            await inner_task
```

T-256 (commented):
Great,
Also, is this mean we no more need `anyio` dependency for `httpx[asyncio]` installation on `Python>=3.11` ?

httpcore/_synchronization.py
MarkusSintonen: Done, also needed to add the function to the `unasync` utility.

httpcore/_synchronization.py
MarkusSintonen: Done

httpcore/_synchronization.py
MarkusSintonen: Done

httpcore/_synchronization.py
MarkusSintonen: Done

httpcore/_synchronization.py
MarkusSintonen: Done

httpcore/_synchronization.py
MarkusSintonen: Done

httpcore/_synchronization.py
```diff
@@ -18,15 +30,30 @@
...
     # Determine if we're running under trio or asyncio.
     # See https://sniffio.readthedocs.io/en/latest/
     try:
         import sniffio
     except ImportError:  # pragma: nocover
-        environment = "asyncio"
+        environment: AsyncBackend = "asyncio"
     else:
-        environment = sniffio.current_async_library()
+        environment = cast(AsyncBackend, sniffio.current_async_library())
```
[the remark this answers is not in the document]
MarkusSintonen: Done

httpcore/_synchronization.py
```diff
@@ -53,28 +90,26 @@ class AsyncLock:
...
-            self.setup()
-
-        if self._backend == "trio":
-            await self._trio_lock.acquire()
-        elif self._backend == "asyncio":
-            await self._anyio_lock.acquire()
-
+        if (lock := self._lock) is None:
+            lock = self.setup()
+        await lock.acquire()
```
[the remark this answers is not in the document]
MarkusSintonen: Hmm yeah, I used that although all the mypy ignores this requires looks uglier to me

httpcore/_synchronization.py
T-256, line 1: > Actually I fixed the code now so that anyio is not required either in Python<3.11

Then, you could also cleanup unused variable `anyio` on this file.

T-256 (approved):
LGTM Thank You @MarkusSintonen for this.

T-256:
> Yes thats right, once we also bring back the native asyncio based network to `_backends`.

I'll wait you to create a PR from your [commit](https://github.com/encode/httpcore/commit/b29b655718d368b63426ffb1127721e0d2f5ff51) to track and review it separately.

httpcore/_async/connection_pool.py
```diff
@@ -299,11 +299,16 @@ def _assign_requests_to_connections(self) -> List[AsyncConnectionInterface]:
         return closing_connections
 
     async def _close_connections(self, closing: List[AsyncConnectionInterface]) -> None:
+        if not closing:
+            return
```
tomchristie, line 303: Could you explain this change to me?

httpcore/_synchronization.py
```diff
@@ -112,117 +128,95 @@ def __exit__(
...
+        with trio.CancelScope(shield=True):
+            await shielded()
+    else:
+        inner_task = asyncio.create_task(shielded())
+        retry = False
+        while True:
+            try:
+                await asyncio.shield(inner_task)
+                break
+            except asyncio.CancelledError:
```
tomchristie, line 213: Why are we not just using a standard shield here?

```python
inner_task = asyncio.create_task(shielded())
await asyncio.shield(inner_task)
```

httpcore/_synchronization.py
MarkusSintonen: Its still needed there for the check in the file as `anyio` needs to be present due to the networking backend

httpcore/_async/connection_pool.py
MarkusSintonen: If its an empty list then we dont need to go through `await async_cancel_shield` below, right? If that was the question

httpcore/_synchronization.py
MarkusSintonen: Its to wait for the `inner_task` to finish (unless something is repeatedly trying to cancel this). But if we dont need to wait for the given task to finish then it is not needed and we can use just the simple shield

tomchristie:
This PR highlights to me that the approach to shielding here is wrong.
We're using cancellation shielding solely to deal with shielding connection closes.
Rather than use shielding what we actually ought to be doing is...

* Unconditionally marking the connection state as closed. (Doesn't require I/O, can't be cancelled.)

And then...

* Performing the network stream close. (If we perform a cancellation operation then it's reasonable that SSL connections might not be "gracefully" closed.)

This is a bit fiddly to get right, needs some careful attention to detail.

MarkusSintonen:
@tomchristie do you mean the previous contextmanager based shielding also had issues? Requiring additional reordering of closing handlers? If it didnt have issues then we could still use the anyio CancelScope just for shielding? But it means the anyio dependency cant be fully removed.

MarkusSintonen:
Or yeah did you mean the non-IO related state management should be extracted out from the async closing when shielding with asyncio.shield. That's probably right

tomchristie:
@MarkusSintonen I've attempted to answer my intent here through implementation. See #927. (Currently draft, and approach would need to be applied also to the HTTP/2 implementation, and the closing behaviour in the connection pool)

MarkusSintonen:
> @MarkusSintonen I've attempted to answer my intent here through implementation. See #927. (Currently draft, and approach would need to be applied also to the HTTP/2 implementation, and the closing behaviour in the connection pool)

Big thanks! Now I understand, so yeah it was indeed about "separating" the sync state management and the async IO part. Ie the streams aclose is the cancellable part and the state management being separate from async flow can not be cancelled.

agronholm:
Would you mind testing my latest PR branch (https://github.com/agronholm/anyio/pull/761)? In my own testing, it about halved the performance overhead of `anyio.Lock` on asyncio, and was actually *faster* than asyncio's own native `Lock` class.

MarkusSintonen:
> Would you mind testing my latest PR branch ([agronholm/anyio#761](https://github.com/agronholm/anyio/pull/761))? In my own testing, it about halved the performance overhead of `anyio.Lock` on asyncio, and was actually _faster_ than asyncio's own native `Lock` class.

Nice, thank you! I can try sometime next week.

MarkusSintonen:
> Would you mind testing my latest PR branch (https://github.com/agronholm/anyio/pull/761)? In my own testing, it about halved the performance overhead of anyio.Lock on asyncio, and was actually faster than asyncio's own native Lock class.

@agronholm I run the tests and it is indeed better with the fix. But there seems there is still some overhead with the implementation when comparing to asyncio:

Baseline (Anyio 4.4.0):
baseline

Anyio [PR](https://github.com/agronholm/anyio/pull/761):
branch

Asyncio:
asyncio

(Above tests were run with httpcore optimizations from other pending PRs to not count in the issues from httpcore itself)

agronholm:
> > Would you mind testing my latest PR branch ([agronholm/anyio#761](https://github.com/agronholm/anyio/pull/761))? In my own testing, it about halved the performance overhead of anyio.Lock on asyncio, and was actually faster than asyncio's own native Lock class.
> 
> @agronholm I run the tests and it is indeed better with the fix. But seems there is still some overhead with the implementation when comparing to asyncio:
> 
> Baseline (Anyio 4.4.0): baseline
> 
> Anyio [PR](https://github.com/agronholm/anyio/pull/761): branch
> 
> Asyncio: asyncio
> 
> (Above tests were run with httpcore optimizations from other pending PRs to not count in the issues from httpcore itself)

Have you identified where this overhead is coming from? Can we eliminate synchronization as the cause after that PR?

agronholm:
If you run the benchmarks after replacing the locks and semaphores in the `asyncio` implementation with the corresponding AnyIO primitives, that should shed some light on the issue.

MarkusSintonen:
> If you run the benchmarks after replacing the locks and semaphores in the `asyncio` implementation with the corresponding AnyIO primitives, that should shed some light on the issue.

It seems to be the `anyio.Lock`. When I just replace it with `asyncio.Lock` the issue disappears. (HTTP2 also utilizes the `Semaphore` but I haven't tested that here.)

agronholm:
In the AnyIO variant, the acquisition of an unowned lock would be slower because this operation contains a yield point unlike in the asyncio variant. Would you mind testing against AnyIO lock after commenting out the checkpoint calls around lines 1690-1697 in `backends/_asyncio.py`?

MarkusSintonen:
> In the AnyIO variant, the acquisition of an unowned lock would be slower because this operation contains a yield point unlike in the asyncio variant. Would you mind testing against AnyIO lock after commenting out the checkpoint calls around lines 1690-1697 in `backends/_asyncio.py`?

Indeed, it is the `await AsyncIOBackend.cancel_shielded_checkpoint()` (at line 1694)

agronholm:
Yeah, so the problem here is that this yield point is needed to maintain the same semantics as with Trio. This is Trio's `Lock.acquire()`:
```py
    @enable_ki_protection
    async def acquire(self) -> None:
        """Acquire the lock, blocking if necessary."""
        await trio.lowlevel.checkpoint_if_cancelled()
        try:
            self.acquire_nowait()
        except trio.WouldBlock:
            # NOTE: it's important that the contended acquire path is just
            # "_lot.park()", because that's how Condition.wait() acquires the
            # lock as well.
            await self._lot.park()
        else:
            await trio.lowlevel.cancel_shielded_checkpoint()
```
If there wasn't a yield point, then this could become a busy-wait loop:
```py
async def loop(lock):
    while True:
        async with lock:
            print("I'm holding the lock")
```

MarkusSintonen:
> If there wasn't a yield point, then this could become a busy-wait loop:

So this is needed for the `CancelScope` logic in AnyIO? What if nothing is using the `CancelScope` in the caller stack? Is the great penalty avoidable if user is not actually using `CancelScope`s?

agronholm:
No, the `checkpoint_if_cancelled()` call handles `CancelScope` support. The `cancel_shielded_checkpoint()` there prevents the busy-wait loop which will happen if you run the above loop with `asyncio` primitives, as they weren't designed with the right principles.

agronholm:
To summarize, asyncio's incorrect design makes it faster here.

agronholm:
> Ok I see thanks for the explanation! So in essence `anyio.Lock` can be used to work around the issue (do Python core devs know about it? :D). `httpcore` doesnt utilize the `_synchronization.AsyncLock` in such ways so it seems its safe to replace the `anyio.Lock` with `asyncio.Lock` to not pay the penalty.

Only in the `asyncio` backend. In the `AnyIO` backend, you would need to use AnyIO primitives, lest your render the whole backend pointless.

rattrayalex:
What is the status of this PR? Should it be reviewed+merged, closed, or something else?

agronholm:
> What is the status of this PR? Should it be reviewed+merged, closed, or something else?

Perhaps we should see how https://github.com/agronholm/anyio/pull/761 affects the real-world performance before jumping the gun?

MarkusSintonen:
> > What is the status of this PR? Should it be reviewed+merged, closed, or something else?
> 
> Perhaps we should see how [agronholm/anyio#761](https://github.com/agronholm/anyio/pull/761) affects the real-world performance before jumping the gun?

Also posting here, I rerun the benchmarks with above fix from AnyIO using `fast_acquire=True` for Lock (and Semaphore) in httpcore. The results look pretty good with AnyIO fix and `fast_acquire=True` usage.

First results with all optimization PRs applied (to not count other httpcore issues in).

With `fast_acquire=False`:
Näyttökuva 2024-09-02 kello 20 15 55

With `fast_acquire=True`:
Näyttökuva 2024-09-02 kello 20 16 46

Lastly results without all optimization PRs applied (how it would look against httpcore master, with its other issues).

With `fast_acquire=False`:
Näyttökuva 2024-09-02 kello 20 22 11

With `fast_acquire=True`:
Näyttökuva 2024-09-02 kello 20 20 50

MarkusSintonen:
Should we keep AnyIO but instead use the `fast_acquire=True` mode in `_synchronization.py`? Cc @tomchristie

tomchristie:
> Should we keep AnyIO but instead use the fast_acquire=True mode in _synchronization.py?

Exercise caution. What does the pull request for that look like? That'll be a useful highlighter.

MarkusSintonen:
> Exercise caution. What does the pull request for that look like? That'll be a useful highlighter.

Here is the PR https://github.com/encode/httpcore/pull/953/files

agronholm:
I see a slight issue there, not among the changes but still: AnyIO 4.5.0 requires trio >= 0.26.1 in its `trio` extra, but those requirements limit it to < 0.26.0.

rattrayalex:
@tomchristie should this remain closed? Not sure if `fast_acquire` in anyio has proven sufficient here?

agronholm:
> @tomchristie should this remain closed? Not sure if `fast_acquire` in anyio has proven sufficient here?

Do you have evidence to the contrary? All the benchmarks I've seen that contain all the relevant performance improvements place httpcore's performance on the same level as aiohttp.

rattrayalex:
I don't, just wasn't aware of any data either way once that was released into the wild, so wanted to check -- sounds like it does indeed make sense to close!

agronholm:
It looks like after @graingert merged the `fast_acquire` [PR](https://github.com/encode/httpcore/pull/953), @tomchristie reverted it. I don't understand the rationale behind the reversion, but I commented on the PR so hopefully we will hear more.

PR #923, opened 2024-06-10
Add benchmarking script

MarkusSintonen:
# Summary

Adds benchmarking script comparing `httpcore` against another popular http library `aiohttp`. Related to performance issues `httpcore` has https://github.com/encode/httpx/issues/3215.

This runs 500 requests with 20 requests concurrently. The test is repeated 10 times. Test server has simulated latency of 10ms with 2KB response. Simulating a good network connection (eg in data center conditions).

# Checklist

### DIFF
--- a/tests/benchmark/server.py
+++ b/tests/benchmark/server.py
@@ -1,0 +1,39 @@
+import asyncio
+
+import uvicorn
+
+PORT = 1234
+RESP = b"a" * 2000
+SLEEP = 0.01
+
+
+async def app(scope, receive, send):
+    assert scope["type"] == "http"
+    assert scope["path"] == "/req"
+    assert not (await receive()).get("more_body", False)
+
+    await asyncio.sleep(SLEEP)
+    await send(
+        {
+            "type": "http.response.start",
+            "status": 200,
+            "headers": [[b"content-type", b"text/plain"]],
+        }
+    )
+    await send(
+        {
+            "type": "http.response.body",
+            "body": RESP,
+        }
+    )
+
+
+if __name__ == "__main__":
+    uvicorn.run(
+        app,
+        port=PORT,
+        log_level="error",
+        # Keep warmed up connections alive during the test to have consistent results across test runs.
+        # This avoids timing differences with connections getting closed and reopened in the background.
+        timeout_keep_alive=100,
+    )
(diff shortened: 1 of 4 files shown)


tests/benchmark/server.py
```diff
@@ -0,0 +1,22 @@
+import asyncio
+
+from aiohttp import web
```
T-256, line 3: Could you use another library here? e.g. FastAPI, to avoid any possible internal optimization between aiohttp's client and server.

MarkusSintonen: Hmm it shouldn't make any difference as its running in a separate process as a normal HTTP server. I wouldn't want to include FastAPI into the benchmark as it is a pretty thick framework.

MarkusSintonen: Switched to using `uvicorn` (which FastAPI also uses in background). It makes no difference when rerunning benchmarks.

T-256 (approved):

tomchristie:
Great thanks. Is there any sense in us also having the same benchmark for the threaded case?
Eg. Comparison between `requests` and `httpx`.

MarkusSintonen:
> Great thanks. Is there any sense in us also having the same benchmark for the threaded case? Eg. Comparison between `requests` and `httpx`.

It might make sense for checking the pool optimization with the sync case

MarkusSintonen:
@tomchristie I added now here the sync benchmark using `requests` library as a baseline. Rerun the benchmark for the connection pool optimization https://github.com/encode/httpcore/pull/924 and updated the description there. There results look pretty promising for the sync-case. With that `httpcore` reaches and is even faster than `requests`.

karpetrosyan:
I think the most important part of performance improvement is writing effective benchmarks. So, I have two questions here:

- Should we run these two benchmarks in one process, or would we get more precision if we isolate them?
- Should we use requests or urllib3? I suppose httpcore is more akin to an alternative to urllib3 rather than requests :d

MarkusSintonen:
> * Should we run these two benchmarks in one process, or would we get more precision if we isolate them?

It doesnt actually work very well when running in a single process as then the benchmark also includes the server side. The sync benchmark especially breaks as the async server wont be able to process the requests as the sync IO is blocking everything. So it would need also to use some threaded sync server which starts to get complex to my taste just for the benchmarking.

> * Should we use requests or urllib3? I suppose httpcore is more akin to an alternative to urllib3 rather than requests :d

Good point, I think it makes sense to use `urllib3` as a baseline. I switched now to `urllib3` and updated the benchmark results in the connection pool PR.

tomchristie:
Okay, yes.

tomchristie (approved):

PR #924, opened 2024-06-10
Optimize connection pool implementation

MarkusSintonen:
# Summary

Second PR in series of optimizations to improve performance of `httpcore` and even reach performance levels of `aiohttp` (and `urllib3` library).
Related discussion https://github.com/encode/httpx/issues/3215#issuecomment-2157885121

Optimizes the connection pool by reducing the time complexity of the idling connections checks. Also it no longer checks the sockets readable status on every pool operation. Pools `has_expired` check uses socket polling via `is_readable` check which is relatively expensive. So now the polling is done using smudged intervals (smudging to avoid all polls being done at the exactly same time). This still should have relatively low chance of encountering broken keep alive connections when connection is picked up from the pool.

**Async** previously:
old1

**Async** with PR:
new2

Async request latency is not so stable yet (as this doesn't include https://github.com/encode/httpcore/pull/922) but the overall duration of the [benchmark](https://github.com/encode/httpcore/pull/923) improves by 7.5x. (The difference diminishes when server latency increases over 100ms or so.)

**Sync** previously:
sync_old

**Sync** with PR:
sync_new

Sync request latency improves by x2.5. With this `httpcore` has same performance as `urllib3` library.

# Checklist

### DIFF
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -238,6 +238,7 @@ class AsyncConnectionPool(AsyncRequestInterface):
         those connections to be handled seperately.
         """
         closing_connections = []
+        idling_count = 0
 
         # First we handle cleaning up any connections that are closed,
         # have expired their keep-alive, or surplus idle connections.
@@ -249,26 +250,24 @@ class AsyncConnectionPool(AsyncRequestInterface):
                 # log: "closing expired connection"
                 self._connections.remove(connection)
                 closing_connections.append(connection)
-            elif (
-                connection.is_idle()
-                and len([connection.is_idle() for connection in self._connections])
-                > self._max_keepalive_connections
-            ):
+            elif connection.is_idle():
+                if idling_count < self._max_keepalive_connections:
+                    idling_count += 1
+                    continue
                 # log: "closing idle connection"
                 self._connections.remove(connection)
                 closing_connections.append(connection)
 
         # Assign queued requests to connections.
-        queued_requests = [request for request in self._requests if request.is_queued()]
-        for pool_request in queued_requests:
+        for pool_request in list(self._requests):
+            if not pool_request.is_queued():
+                continue
+
             origin = pool_request.request.url.origin
             available_connections = [
                 connection
                 for connection in self._connections
                 if connection.can_handle_request(origin) and connection.is_available()
-            ]
-            idle_connections = [
-                connection for connection in self._connections if connection.is_idle()
             ]
 
             # There are three cases for how we may be able to handle the request:
@@ -286,15 +285,18 @@ class AsyncConnectionPool(AsyncRequestInterface):
                 connection = self.create_connection(origin)
                 self._connections.append(connection)
                 pool_request.assign_to_connection(connection)
-            elif idle_connections:
-                # log: "closing idle connection"
-                connection = idle_connections[0]
-                self._connections.remove(connection)
-                closing_connections.append(connection)
-                # log: "creating new connection"
-                connection = self.create_connection(origin)
-                self._connections.append(connection)
-                pool_request.assign_to_connection(connection)
+            else:
+                idling_connection = next(
+                    (c for c in self._connections if c.is_idle()), None
+                )
+                if idling_connection is not None:
+                    # log: "closing idle connection"
+                    self._connections.remove(idling_connection)
+                    closing_connections.append(idling_connection)
+                    # log: "creating new connection"
+                    new_connection = self.create_connection(origin)
+                    self._connections.append(new_connection)
+                    pool_request.assign_to_connection(new_connection)
 
         return closing_connections
 
--- a/httpcore/_async/http11.py
+++ b/httpcore/_async/http11.py
@@ -1,5 +1,6 @@
 import enum
 import logging
+import random
 import ssl
 import time
 from types import TracebackType
@@ -56,10 +57,12 @@ class AsyncHTTP11Connection(AsyncConnectionInterface):
         origin: Origin,
         stream: AsyncNetworkStream,
         keepalive_expiry: Optional[float] = None,
+        socket_poll_interval_between: Tuple[float, float] = (1, 3),
     ) -> None:
         self._origin = origin
         self._network_stream = stream
-        self._keepalive_expiry: Optional[float] = keepalive_expiry
+        self._keepalive_expiry = keepalive_expiry
+        self._socket_poll_interval_between = socket_poll_interval_between
         self._expire_at: Optional[float] = None
         self._state = HTTPConnectionState.NEW
         self._state_lock = AsyncLock()
@@ -68,6 +71,8 @@ class AsyncHTTP11Connection(AsyncConnectionInterface):
             our_role=h11.CLIENT,
             max_incomplete_event_size=self.MAX_INCOMPLETE_EVENT_SIZE,
         )
+        # Assuming we were just connected
+        self._network_stream_used_at = time.monotonic()
 
     async def handle_async_request(self, request: Request) -> Response:
         if not self.can_handle_request(request.url.origin):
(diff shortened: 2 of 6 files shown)


httpcore/_async/connection_pool.py
```diff
@@ -238,24 +238,27 @@ def _assign_requests_to_connections(self) -> List[AsyncConnectionInterface]:
         those connections to be handled seperately.
         """
         closing_connections = []
+        idling_connections = {c for c in self._connections if c.is_idle()}
```
karpetrosyan, line 241: It seems like we use this collection just for tracking the count of idle connections. Maybe it should just be an integer for simplicity?

MarkusSintonen: With integer we would need to recheck all the connections again. We can not just decrement below in the if-elif branches as we could end up going negative etc

MarkusSintonen: Heh sorry, you are right. We can just check eg the expired one if it was also idle one and decrement!

karpetrosyan: Does that part significantly change the performance? Did you run any tests? Maybe we are gaining the same performance boost by only using polling?

MarkusSintonen: When I checked with `pyinstrument` it showed this exact spot as a hot spot spending time iterating the connections in outer and inner loops. Ill rerun the benchmark to check it again

MarkusSintonen: Anyways I pushed now the integer fix

karpetrosyan:
I have similar results without this PR. Am I doing something wrong? 
I am using benchmarks script from [your another PR](https://github.com/encode/httpcore/pull/923)

Here is what I got on my machine without this PR:
image

MarkusSintonen:
> I have similar results without this PR. Am I doing something wrong?

What kind of system are you on and python version?

httpcore/_async/connection_pool.py
MarkusSintonen: Without the fix in `_assign_requests_to_connections` the performance starts to deteriorate when the amount of connections in the pool increases (which is not surprising as there is the loop in the loop).

With 20 concurrency (sync):
sync_20

With 40 concurrency (sync):
sync_40

httpcore/_async/connection_pool.py
```diff
@@ -238,28 +238,27 @@ def _assign_requests_to_connections(self) -> List[AsyncConnectionInterface]:
...
             elif (
-                connection.is_idle()
-                and len(idling_connections) > self._max_keepalive_connections
+                connection.is_idle() and idling_count > self._max_keepalive_connections
             ):
                 # log: "closing idle connection"
                 self._connections.remove(connection)
-                idling_connections.discard(connection)
                 closing_connections.append(connection)
+                idling_count -= 1
```
T-256, line 261: Just for myself interest: Could we have *single-loop incremental* idling count instead of *double-loop decremental*?
```python
        closing_connections = []
        idling_count = 0

        # First we handle cleaning up any connections that are closed,
        # have expired their keep-alive, or surplus idle connections.
        for connection in list(self._connections):
            if connection.is_closed():
                # log: "removing closed connection"
                self._connections.remove(connection)
            elif connection.has_expired():
                # log: "closing expired connection"
                self._connections.remove(connection)
                closing_connections.append(connection)
            elif connection.is_idle():
                if idling_count > self._max_keepalive_connections:
                    # log: "closing idle connection"
                    self._connections.remove(connection)
                    closing_connections.append(connection)
                else:
                    idling_count += 1

```

MarkusSintonen: We could simplify even more if we recreate the `self._connections` and not `remove` from it:
```python
connections = []
closing_connections = []
idling_count = 0

# First we handle cleaning up any connections that are closed,
# have expired their keep-alive, or surplus idle connections.
for connection in self._connections:
    if connection.is_closed():
        # log: "removing closed connection"
        pass
    elif connection.has_expired():
        # log: "closing expired connection"
        closing_connections.append(connection)
    elif connection.is_idle():
        if idling_count > self._max_keepalive_connections:
            # log: "closing idle connection"
            closing_connections.append(connection)
        else:
            idling_count += 1
            connections.append(connection)

self._connections = connections
```

T-256: I also concern about this new change, because `idling_count` will became _dynamic_ through iteration while previously it was _static_ out of iteration. However, we can test it.


> ```python
> self._connections = connections
> ```

I prefer to avoid re-assigning at here 🤷‍♂️

T-256: Ok, I [did test](https://github.com/T-256/httpcore/commit/07c066074e8a5a87eaf39334170cd57bd817cb06) and failed:
https://github.com/T-256/httpcore/actions/runs/9482448452/job/26127347871

MarkusSintonen: Good catch, thanks!

T-256: And fixed by [this change](https://github.com/T-256/httpcore/commit/ffb789a4f692a8aafaea7a15abc5e4648295465f). now tests passing :)

MarkusSintonen: Thanks thats just great, I applied your diff!

httpcore/_async/http11.py
```diff
@@ -281,16 +288,28 @@ def is_available(self) -> bool:
...
+        if (now - self._network_stream_used_at) > self._socket_poll_interval():
+            self._network_stream_used_at = now
+            server_disconnected = (
+                self._state == HTTPConnectionState.IDLE
+                and self._network_stream.get_extra_info("is_readable")
+            )
+            if server_disconnected:
+                return True
+
+        return False
```
T-256, line 307: Actually, I'm not happy with _interval_ calculating.

Could we improve `is_readable` instead? FWIW I noticed anyio and sync backend use `httpcore._utils.is_socket_readable` while trio backend uses its own, @MarkusSintonen is there any benchmark difference when you switch backend to trio?

For improving `is_readable` in `get_extra_info`:
- Always assume it is readable and turn it to false by specified events such as received close socket from server.
- Use synchronized Event on readability status change.

I didn't go deep to search these cases are possible or not. They are only my opinions 🤷‍♂️

T-256 (approved):
Except that _interval_ mechanism (perhaps could be excluded from this PR), it's LGTM, Thank You!

httpcore/_async/http11.py
MarkusSintonen: Trio also suffers badly from the constantly happening socket polling.

Without intervalled socket polling:
old_trio

With intervalled socket polling:
new_trio

So in `trio` its over 5x slower when constantly doing the socket polling.

MarkusSintonen: > For improving is_readable in get_extra_info:
> Always assume it is readable and turn it to false by specified events such as received close socket from server.

Im not aware of anyway to to get events about socket getting closed. As far as I know the only way to know it is to use the socket. 🤔  But I agree the `is_readable` could be better so its not so heavy weight. We could make it just a flag based so in networking side we just set some boolean flag when we detect a network error on usage. This has a downside that we have greater probability of giving our already broken connections from the pool. But as far as I know this is how its usually done.

MarkusSintonen:
> (perhaps could be excluded from this PR)

That was the beef of the PR as the constantly happening socket polling is the most expensive thing in the whole pooling implementation 😅

httpcore/_async/http11.py
MarkusSintonen: Other options would be to move the intervalled polling into the network backends side. Or get even more elaborate and run the socket polling in a specific interval via `loop.call_later` which run the actual poll via `loop.run_in_executor` to avoid any possible nonasync socket IO in the async land. Gets easily hairy 😄

tomchristie:
Let's get this split into two seperate PRs please, so we can look at each in isolation and determine if it's a benefit.

MarkusSintonen:
@tomchristie / @T-256
I have now split the PR into two. Ill close this one.

PR #927, opened 2024-06-13
Remove shielding from cancellation process.

tomchristie:
For handling async-cancellations we're currently shielding `close` operations, in order to ensure we don't end up with a connection in an inconsistent state when cancellations are used.

A better approach is to ensure that the state changes for this case are handled synchronously (so that cancellations won't propagate into those). While the network close remains unshielded async (it's okay for cancellations to propagate into those).

Reasoning about the state when cancellations occur is a bit fiddly, tho I think we can apply this same style of approach all the way through to remove the need for async cancellation-shielding. (Closing state is applied first synchronously. Network close operations are then attempted, and may propagate cancellation.)

Refs #922


httpcore/_async/http11.py
```diff
@@ -344,15 +353,16 @@ async def __aiter__(self) -> AsyncIterator[bytes]:
...
 
     async def aclose(self) -> None:
-        if not self._closed:
-            self._closed = True
-            async with Trace("response_closed", logger, self._request):
-                await self._connection._response_closed()
+        async with Trace("response_closed", logger, self._request, kwargs={}):
+            if not self._closed:
+                self._closed = True
+                if self._connection._connection_should_close():
```
MarkusSintonen, line 364: Can the await inside the `Trace.__aenter__` cause issues here? Ie can there be a cancellation with the await inside it that would cause a skip of the `self._closed`-checks and `self._connection._connection_should_close()`?

httpcore/.DS_Store
T-256, line 1: junk?

T-256 (commented):
There are also in _connection_pool_ and _http2_

rattrayalex:
@tomchristie do you intend to take this over the line, or would it be helpful for someone else to try to take it over?

(for context, users of `openai-python` are [reporting](https://github.com/openai/openai-python/issues/1596) unusably slow performance, which seems related to https://github.com/encode/httpx/issues/3215)

httpcore/.DS_Store
graingert, line 1: probably don't want this file

httpcore/_async/http11.py
graingert: yes this is a problem, It's also a problem if the trace function raises, I think you'll need to do this:

```python
# synchronization.py
aclose_forcefully = anyio.aclose_forcefully
def close_forcefully(sock):
    sock.close()
```

```python
    async def aclose(self):
        entered_cmgr = False
        try:
            async with Trace("response_closed", logger, self._request, kwargs={}):
                entered_cmgr = True
                if not self._closed:
                    self._closed = True
                    if self._connection._connection_should_close():
                        await self._connection.aclose()
        except BaseException:
            if entered_cmgr:
                raise
            if not self._closed:
                self._closed = True
                if self._connection._connection_should_close():
                    await aclose_forcefully(self._connection)
            raise

```

graingert (changes requested):

httpcore/_async/http11.py
```diff
@@ -137,9 +141,8 @@ async def handle_async_request(self, request: Request) -> Response:
                 },
             )
         except BaseException as exc:
-            with AsyncShieldCancellation():
-                async with Trace("response_closed", logger, request) as trace:
-                    await self._response_closed()
+            if self._connection_should_close():
+                await self._network_stream.aclose()
```
graingert, line 145: this isn't tracing anymore

rattrayalex:
Any update here?

T-256:
Bump

PR #928, opened 2024-06-13
Connection pool optimization: reduce socket polling with expiry checks

MarkusSintonen:
# Summary

(Split from original PR here https://github.com/encode/httpcore/pull/924)

Connection pool implementation is heavily doing socket polling via `AsyncConnectionInterface.has_expired()`. This check in turn does `AsyncNetworkStream.get_extra_info("is_readable")` which finally does `select.poll()`. This check quickly starts to add up when doing many requests through `httpcore` with great concurrency. Many connection sockets in the pool are constantly polled over and over again when doing requests and also when request closes.

This PR instead proposes on doing the socket polling only with certain smudged intervals. This reduces greatly the amount of socket related work this had to do previously. (Smudging is applied to polling timing to avoid many connections doing socket polling at once when connections are being opened in close time proximity).

[Benchmarks](https://github.com/encode/httpcore/tree/master/tests/benchmark) show how the heavily done socket polling causes over 4.5x overhead.

**In [master](https://github.com/encode/httpcore/commits/da86ca4e22e64b078cc8a574f43c8ad2564e2e27/) with async requests:**
old

**PR with async:**
new_socket

Similar results are obtained with synchronous code. There socket polling causes over 2x overhead (most probably the overhead is lower due to sync concurrency using threads where the socket polling IO is spread to threads).

**In master with sync requests:**
old_sync

**PR with sync:**
new_sync_socket

Trio-based backend is also affected by this. There the overhead is at similar levels of about 5x as was seen [here](https://github.com/encode/httpcore/pull/924#discussion_r1636879771).

# Checklist


httpcore/_async/http11.py
```diff
@@ -281,16 +288,28 @@ def is_available(self) -> bool:
...
-        server_disconnected = (
-            self._state == HTTPConnectionState.IDLE
-            and self._network_stream.get_extra_info("is_readable")
-        )
+        # Checking the readable status is relatively expensive so check it at a lower frequency.
+        if (now - self._network_stream_used_at) > self._socket_poll_interval():
+            self._network_stream_used_at = now
+            server_disconnected = (
+                self._state == HTTPConnectionState.IDLE
+                and self._network_stream.get_extra_info("is_readable")
```
MarkusSintonen, line 302: I'm still slightly worried about what this check does behind the scenes. Does the socket poll cause IO so in async code it disturbs other requests being processed as this is doing the IO in non-async code? Maybe its more fine when we reduce the amount of polling happening.

MarkusSintonen: Another option would be to use this style of approach for checking aliveness. So preemptively trying to load data from the stream into a buffer. https://github.com/encode/httpx/issues/2983#issuecomment-1843756808

MarkusSintonen: Or yet another option would be to do this similarly as in `urllib3` so we would do the socket poll check when picking up the connection from the pool and not as part of the pool maintenance loop. Like here [is_connection_dropped](https://github.com/urllib3/urllib3/blob/main/src/urllib3/connectionpool.py#L290) which uses [wait_for_read](https://github.com/urllib3/urllib3/blob/da410581b6b3df73da976b5ce5eb20a4bd030437/src/urllib3/connection.py#L260) doing the [pool](https://github.com/urllib3/urllib3/blob/da410581b6b3df73da976b5ce5eb20a4bd030437/src/urllib3/util/wait.py#L70) similarly as in `httpcore`

MarkusSintonen:
@T-256 I fixed this PR to drop the interval based socket polling approach as you suggested [here](https://github.com/encode/httpcore/pull/924#discussion_r1636844294).

Instead this is now using a very similar approach as in `urllib3` where the socket health is validated at the connection usage time when its taken from the pool. In oppose of polling sockets heavily for all the connections in the pool every single time requests are done and closed.

httpcore/_async/http11.py
```diff
@@ -77,6 +77,16 @@ async def handle_async_request(self, request: Request) -> Response:
             )
 
         async with self._state_lock:
+            # If the HTTP connection is idle but the socket is readable, then the
+            # only valid state is that the socket is about to return b"", indicating
+            # a server-initiated disconnect.
+            server_disconnected = (
+                self._state == HTTPConnectionState.IDLE
+                and self._network_stream.get_extra_info("is_readable")
```
MarkusSintonen, line 85: Should this check be also done in `AsyncHTTP2Connection` similarly as here?

httpcore/_async/http11.py
```diff
@@ -77,6 +77,16 @@ async def handle_async_request(self, request: Request) -> Response:
...
+            )
+            if server_disconnected:
+                raise ConnectionNotAvailable()
```
MarkusSintonen, line 88: Should this check be also done in `AsyncHTTP2Connection` similarly as here?

httpcore/_async/http11.py
```diff
@@ -77,6 +79,17 @@ async def handle_async_request(self, request: Request) -> Response:
             )
 
         async with self._state_lock:
+            # If the HTTP connection is idle but the socket is readable, then the
+            # only valid state is that the socket is about to return b"", indicating
+            # a server-initiated disconnect.
+            server_disconnected = (
+                self._state == HTTPConnectionState.IDLE
+                and self._network_stream.get_extra_info("is_readable")
+            )
```
MarkusSintonen, line 88: Should this check be also done in `AsyncHTTP2Connection` similarly as here?

httpcore/_exceptions.py
```diff
@@ -19,6 +19,10 @@ class ConnectionNotAvailable(Exception):
     pass
 
 
+class ServerDisconnectedError(Exception):
```
T-256, line 22: It could be subclass of `ConnectionNotAvailable` since there connection won't be available after server closed it. this would prevent from [this change](https://github.com/encode/httpcore/pull/928/files#diff-2c49ff24c06fb1f90e1cba7d18757d80f3197a92106e04ba77179497461d69aa).

Also, we need to add it in docs.

httpcore/_async/http11.py
```diff
@@ -77,6 +79,20 @@ async def handle_async_request(self, request: Request) -> Response:
...
+            if server_disconnected:
+                self._state = HTTPConnectionState.SERVER_DISCONNECTED
+                raise ServerDisconnectedError()
+
```
T-256, line 95: most of users make subclass of `handle_async_request` to customize their _Client_ instances. I think it could be better for us to keep it cleaner and more readable. 
Could you export this change into priavte method `_raise_for_state` which have no returns and only raises if possible.

tests/_sync/test_http11.py
```diff
@@ -1,6 +1,9 @@
+import typing
+
 import pytest
 
 import httpcore
+from httpcore._exceptions import ServerDisconnectedError
```
T-256, line 6:
```suggestion
```
Don't import from private modules

tests/_sync/test_http11.py
```diff
@@ -168,6 +171,47 @@ def test_http11_connection_handles_one_active_request():
...
+        assert conn.is_idle() and not conn.has_expired()
+        stream.mock_is_readable = True  # Simulate connection breakage
+
+        with pytest.raises(ServerDisconnectedError):
+            conn.request("GET", "https://example.com/")
+        assert conn.has_expired() and not conn.is_idle()
+
+        with pytest.raises(ServerDisconnectedError):
+            conn.request("GET", "https://example.com/")
+        assert conn.has_expired() and not conn.is_idle()
```
T-256, line 211:
```suggestion
        with pytest.raises(httpcore.ServerDisconnectedError):
            conn.request("GET", "https://example.com/")
        assert conn.has_expired() and not conn.is_idle()
```

T-256 (changes requested):
Less overheads than _interval_ mechanism :)
btw here are few review comments:

T-256:
@MarkusSintonen Do you mind/working to send PR for re-adding `asyncio` backend and drop `anyio` backend? I'm also have interest on those works.

MarkusSintonen:
> @MarkusSintonen Do you mind/working to send PR for re-adding `asyncio` backend and drop `anyio` backend? I'm also have interest on those works.

Sure, I was wishing on getting the synchronization PR first approved/merged before opening it. But I can open it already.

httpcore/_exceptions.py
MarkusSintonen: Done, made it a subclass of the existing one. Should we have a subclass also for the previous state related error?

> Also, we need to add it in docs.

I can not see existing one mentioned in the docs. Probably due to the exception being handled by the pool. I added docstrings instead to the classes about this.

MarkusSintonen: Im now wondering do we even need the new `ServerDisconnectedError`. Because this is also internal to the pooling mechanism like the previous one.

httpcore/_async/http11.py
MarkusSintonen: Done, moved it into `_update_state` function as we also need to update the state as part of raising. (Also for http2 to have it consistent)

tests/_sync/test_http11.py
MarkusSintonen: Fixed

tests/_sync/test_http11.py
MarkusSintonen: Fixed (the latter one was actually for coverage but I changed to logic so that we dont need the extra assert step here for coverage)

httpcore/_sync/http11.py
```diff
@@ -142,6 +138,29 @@ def handle_request(self, request: Request) -> Response:
...
 
+    def _update_state(self) -> None:
+        with self._state_lock:
+            # If the HTTP connection is idle but the socket is readable, then the
+            # only valid state is that the socket is about to return b"", indicating
+            # a server-initiated disconnect.
+            server_disconnected = (
+                self._state == HTTPConnectionState.IDLE
+                and self._network_stream.get_extra_info("is_readable")
+            )
```
MarkusSintonen, line 149: Should this check be also done in `AsyncHTTP2Connection` similarly as here?

httpcore/_async/connection_pool.py
```diff
@@ -286,15 +283,19 @@ def _assign_requests_to_connections(self) -> List[AsyncConnectionInterface]:
...
-                connection = idle_connections[0]
-                self._connections.remove(connection)
-                closing_connections.append(connection)
-                # log: "creating new connection"
-                connection = self.create_connection(origin)
-                self._connections.append(connection)
-                pool_request.assign_to_connection(connection)
+            else:
+                purged_connection = next(
+                    (c for c in self._connections if c.is_idle() or c.has_expired()),
```
MarkusSintonen, line 288: I think this should also consider expired connections here for removal. Because A) clock may have moved forward so that some are now expired here B) expired ones are the ones that maybe have been disconnected by the server (they are not idle)

httpcore/_async/http11.py
```diff
@@ -142,6 +138,29 @@ async def handle_async_request(self, request: Request) -> Response:
...
 
+    async def _update_state(self) -> None:
+        async with self._state_lock:
+            # If the HTTP connection is idle but the socket is readable, then the
+            # only valid state is that the socket is about to return b"", indicating
+            # a server-initiated disconnect.
+            server_disconnected = (
+                self._state == HTTPConnectionState.IDLE
+                and self._network_stream.get_extra_info("is_readable")
+            )
```
MarkusSintonen, line 149: Should this check be also done in `AsyncHTTP2Connection` similarly as here?

T-256: I think yes, but cc @tomchristie who knows better about http2

T-256 (approved):

MarkusSintonen:
> @MarkusSintonen Do you mind/working to send PR for re-adding `asyncio` backend and drop `anyio` backend? I'm also have interest on those works.

Its now here: https://github.com/encode/httpcore/pull/930

httpcore/_async/connection_pool.py
```diff
@@ -267,15 +267,12 @@ def _assign_requests_to_connections(self) -> List[AsyncConnectionInterface]:
                 for connection in self._connections
                 if connection.can_handle_request(origin) and connection.is_available()
             ]
```
eth3lbert, line 269: It seems like we could just find the first available one here?

```python
            available_connections = next(
                (
                    connection
                    for connection in self._connections
                    if connection.can_handle_request(origin)
                    and connection.is_available()
                ),
                None,
            )
           ...
           if available_connections:
               connection = available_connections
               ...
```

httpcore/_sync/connection_pool.py
```diff
@@ -267,15 +267,12 @@ def _assign_requests_to_connections(self) -> List[ConnectionInterface]:
                 for connection in self._connections
                 if connection.can_handle_request(origin) and connection.is_available()
             ]
```
eth3lbert, line 269: ditto.

httpcore/_async/connection_pool.py
eth3lbert: Oh, I just noticed that this has been split into separate PRs. 😅

PR #929, opened 2024-06-13
Connection pool optimization: reduce connection maintenance loops complexity

MarkusSintonen:
# Summary

This is a small low hanging fruit for optimizing and simplifying the connection maintenance loops in the connection pool. This work is based on suggestion by @T-256 [here](https://github.com/encode/httpcore/pull/924#discussion_r1636465142) (thanks!). Previously performance of `httpcore` degraded as connection count increased in the pool. As the maintenance loops had to do more and more work iterating and reiterating the connections.

This optimization brings the performance of `httpcore` to same level as `urllib3` with sync-usage.

The benchmarks below include the [socket polling fix](https://github.com/encode/httpcore/pull/928). (The socket polling problem makes request processing latency so highly varying that it overshadows everything.)

**Previously with sync (with optimized socket polling):**
new_expiry_sync

**PR with sync:**
sync_new

As it can be seen the performance gets to exact same level as `urllib3`. There is almost no overhead from `httpcore` related  request processing anymore.

# Checklist

### DIFF
--- a/httpcore/_sync/connection_pool.py
+++ b/httpcore/_sync/connection_pool.py
@@ -1,7 +1,14 @@
 import ssl
 import sys
 from types import TracebackType
-from typing import Iterable, Iterator, Iterable, List, Optional, Type
+from typing import (
+    Iterable,
+    Iterator,
+    Iterable,
+    List,
+    Optional,
+    Type,
+)
 
 from .._backends.sync import SyncBackend
 from .._backends.base import SOCKET_OPTION, NetworkBackend
@@ -238,6 +245,7 @@ class ConnectionPool(RequestInterface):
         those connections to be handled seperately.
         """
         closing_connections = []
+        idling_count = 0
 
         # First we handle cleaning up any connections that are closed,
         # have expired their keep-alive, or surplus idle connections.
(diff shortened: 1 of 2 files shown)


T-256 (approved):

MarkLux:
wonder if there are any further progress for this PR, our team is waiting for this fix @MarkusSintonen @T-256

MarkusSintonen:
@MarkLux I don't know where the author has gone :/

httpcore/_sync/connection_pool.py
```diff
@@ -249,52 +257,56 @@ def _assign_requests_to_connections(self) -> List[ConnectionInterface]:
...
+                    idling_count += 1
+                    continue
                 # log: "closing idle connection"
                 self._connections.remove(connection)
                 closing_connections.append(connection)
 
         # Assign queued requests to connections.
-        queued_requests = [request for request in self._requests if request.is_queued()]
-        for pool_request in queued_requests:
+        for pool_request in list(self._requests):
```
Mrreadiness, line 269: Do we need a copy of the list here?

httpcore/_sync/connection_pool.py
```diff
@@ -249,52 +257,56 @@ def _assign_requests_to_connections(self) -> List[ConnectionInterface]:
...
-                closing_connections.append(connection)
-                # log: "creating new connection"
-                connection = self.create_connection(origin)
-                self._connections.append(connection)
-                pool_request.assign_to_connection(connection)
+            else:
+                idling_connection = next(
+                    (c for c in self._connections if c.is_idle()), None
+                )
+                if idling_connection is not None:
```
Mrreadiness, line 302: Should we continue iterating over `self._requests` if there is no available or idling connection? Can we `break` in such cases?

Mrreadiness (commented):
Hi @MarkusSintonen! Thank you for your pr, I hope to see it merged soon.
Recently I've faced a performance issue in the pool, in case of a growing number of requests awaiting execution. It looks like your pr will partially solve such issues, I've marked a couple of places with potential improvements.  
Also, have you thought about changing the `List` of `_requests` to something more optimal for queue tasks, for example, `deque`? It could reduce CPU usage for `remove` operations, especially in cases of huge queue.

PR #930, opened 2024-06-16
Add `asyncio` backend

MarkusSintonen:
# Summary

Adds back native `asyncio` backend based on commit where it was removed [here](https://github.com/encode/httpcore/blob/85a1f9ecab09e65d4299b5e2676964c43ac2c02d/httpcore/_backends/asyncio.py). Adjusts it to work with the new interfaces. This also adds some additional integration testing.

Doesn't make the new backend the default.

AnyIO has considerable overhead so this allows again using the native backend. Also this is going to allow removal of the AnyIO as the dependency (when the synchronization [PR](https://github.com/encode/httpcore/pull/922) goes in also).

Overhead is about 1.45x here (the synchronization has a lot more overhead in the other PR).

**Previously (is master):**
old

**Now:**
asyncio

Related discussion https://github.com/encode/httpx/issues/3215#issuecomment-2155630018.

# Checklist

### DIFF
--- a/docs/network-backends.md
+++ b/docs/network-backends.md
@@ -73,8 +73,29 @@ while True:
 
 If we're working with an `async` codebase, then we need to select a different backend.
 
-The `httpcore.AnyIOBackend` is suitable for usage if you're running under `asyncio`. This is a networking backend implemented using [the `anyio` package](https://anyio.readthedocs.io/en/3.x/).
+These `async` network backends are available:
+- `httpcore.AsyncioBackend` This networking backend is implemented using Pythons native `asyncio`.
+- `httpcore.AnyIOBackend` This is implemented using [the `anyio` package](https://anyio.readthedocs.io/en/3.x/).
+- `httpcore.TrioBackend` This is implemented using [`trio`](https://trio.readthedocs.io/en/stable/).
 
+Currently by default `AnyIOBackend` is used when running with `asyncio` (this may change).
+`TrioBackend` is used by default when running with `trio`.
+
+Using `httpcore.AsyncioBackend`:
+```python
+import httpcore
+import asyncio
+
+async def main():
+    network_backend = httpcore.AsyncioBackend()
+    async with httpcore.AsyncConnectionPool(network_backend=network_backend) as http:
+        response = await http.request('GET', 'https://www.example.com')
+        print(response)
+
+asyncio.run(main())
+```
+
+Using `httpcore.AnyIOBackend`:
 ```python
 import httpcore
 import asyncio
--- a/docs/network-backends.md
+++ b/docs/network-backends.md
@@ -74,20 +74,20 @@ while True:
 If we're working with an `async` codebase, then we need to select a different backend.
 
 These `async` network backends are available:
-- `httpcore.AsyncioBackend` This networking backend is implemented using Pythons native `asyncio`.
+- `httpcore.AsyncIOBackend` This networking backend is implemented using Pythons native `asyncio`.
 - `httpcore.AnyIOBackend` This is implemented using [the `anyio` package](https://anyio.readthedocs.io/en/3.x/).
 - `httpcore.TrioBackend` This is implemented using [`trio`](https://trio.readthedocs.io/en/stable/).
 
 Currently by default `AnyIOBackend` is used when running with `asyncio` (this may change).
 `TrioBackend` is used by default when running with `trio`.
 
-Using `httpcore.AsyncioBackend`:
+Using `httpcore.AsyncIOBackend`:
 ```python
 import httpcore
 import asyncio
 
 async def main():
-    network_backend = httpcore.AsyncioBackend()
+    network_backend = httpcore.AsyncIOBackend()
     async with httpcore.AsyncConnectionPool(network_backend=network_backend) as http:
         response = await http.request('GET', 'https://www.example.com')
         print(response)
--- a/httpcore/_backends/asyncio.py
+++ b/httpcore/_backends/asyncio.py
@@ -1,0 +1,227 @@
+import asyncio
+import socket
+import ssl
+from typing import Any, Dict, Iterable, Optional, Type
+
+from .._exceptions import (
+    BrokenSocketError,
+    ConnectError,
+    ConnectTimeout,
+    ReadError,
+    ReadTimeout,
+    WriteError,
+    WriteTimeout,
+    map_exceptions,
+)
+from .._utils import is_socket_readable
+from .base import SOCKET_OPTION, AsyncNetworkBackend, AsyncNetworkStream
+
+
+class AsyncIOStream(AsyncNetworkStream):
+    def __init__(
+        self, stream_reader: asyncio.StreamReader, stream_writer: asyncio.StreamWriter
+    ):
+        self._stream_reader = stream_reader
+        self._stream_writer = stream_writer
+        self._read_lock = asyncio.Lock()
+        self._write_lock = asyncio.Lock()
+        self._inner: Optional[AsyncIOStream] = None
+
+    async def start_tls(
+        self,
+        ssl_context: ssl.SSLContext,
+        server_hostname: Optional[str] = None,
+        timeout: Optional[float] = None,
+    ) -> AsyncNetworkStream:
+        loop = asyncio.get_event_loop()
+
+        stream_reader = asyncio.StreamReader()
+        protocol = asyncio.StreamReaderProtocol(stream_reader)
+
+        exc_map: Dict[Type[Exception], Type[Exception]] = {
+            asyncio.TimeoutError: ConnectTimeout,
+            OSError: ConnectError,
+        }
+        with map_exceptions(exc_map):
+            transport_ssl = await asyncio.wait_for(
+                loop.start_tls(
+                    self._stream_writer.transport,
+                    protocol,
+                    ssl_context,
+                    server_hostname=server_hostname,
+                ),
+                timeout,
+            )
+        if transport_ssl is None:
+            # https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.start_tls
+            raise ConnectError("Transport closed while starting TLS")  # pragma: nocover
+
+        # Initialize the protocol, so it is made aware of being tied to
+        # a TLS connection.
+        # See: https://github.com/encode/httpx/issues/859
+        protocol.connection_made(transport_ssl)
+
+        stream_writer = asyncio.StreamWriter(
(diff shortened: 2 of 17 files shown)


tests/conftest.py
```diff
@@ -0,0 +1,77 @@
...
+    if backend_name == "trio":
+        AutoBackend.set_default_backend(None)
+    else:
+        use_anyio = options.pop("httpcore_use_anyio", False)
+        AutoBackend.set_default_backend(AnyIOBackend if use_anyio else AsyncioBackend)
+
+    return backend_name, options
+
+
+class Server(uvicorn.Server):
```
MarkusSintonen, line 34: Seems the `httpbin` fixture doesnt work well for pool related testing as it always closes the connections immeaditely...

tomchristie:
In order to land this PR I'd suggest adding an `asyncio` native backend *without any other changes*.

The default backend should not change as part of the pull request, and the existing `anyio` backend should not be modified. I'd expect the only modified code files to be `__init__.py` and `httpcore/_backends/asyncio.py`.

The documentation for https://www.encode.io/httpcore/network-backends/#async-network-backends should also be modified to include the native `asyncio` backend.

(Minimal incremental changes ftw)

MarkusSintonen:
> The default backend should not change as part of the pull request

Yep that I didnt change as you previously suggested

MarkusSintonen:
> In order to land this PR I'd suggest adding an `asyncio` native backend _without any other changes_.
> 
> The default backend should not change as part of the pull request, and the existing `anyio` backend should not be modified. I'd expect the only modified code files to be `__init__.py` and `httpcore/_backends/asyncio.py`.

This is now done, so there is only the new backend and export. + The tests I used add the full coverage and verify it works

MarkusSintonen:
> The documentation for https://www.encode.io/httpcore/network-backends/#async-network-backends should also be modified to include the native `asyncio` backend, demo'ing it's usage with

Added this

MarkusSintonen:
Asyncio-backend performs also better when all other current optimization PRs are applied (benchmark params REPEATS=100 REQUESTS=50 CONCURRENCY=20):

**With AnyIO-backend:**
anyio

**With asyncio-backend:**
asyncio

tomchristie:
The code & docs changes look reasonable.
Not obvious that the tests cover the functionality added?

MarkusSintonen:
> Not obvious that the tests cover the functionality added?

Previously the `@pytest.mark.anyio` runned the tests as parametrized by `asyncio-anyio` and `trio` (which is the default). But now via the `conftest` we specify the test parametrization to also include the plain `asyncio` test mode (as documented in AnyIO fixture docs https://anyio.readthedocs.io/en/stable/testing.html#specifying-the-backends-to-run-on).

With this and the added integration testing (eg UDS, socket options) we get close to 100% coverage for the new network backend. (Actually we get even more coverage for anyio/trio backends so there is some no longer needed `nocoverage`)

httpcore/_backends/asyncio.py
```diff
@@ -0,0 +1,228 @@
...
+        return sock
+
+    @property
+    def _sslobj(self) -> Optional[ssl.SSLObject]:
+        transport = self._stream_writer.transport
+        sslobj: Optional[ssl.SSLObject] = transport.get_extra_info("ssl_object")
+        return sslobj
+
+
+class AsyncioBackend(AsyncNetworkBackend):
```
tomchristie, line 167:
```suggestion
class AsyncIOBackend(AsyncNetworkBackend):
```

?

tomchristie (commented):
Looks great, thanks so much.

Naming... should we use `AsyncIOStream` and `AsyncIOBackend` or `AsyncioStream` and `AsyncioBackend`?

bdraco:
Thanks for doing this.  I think this will close out the discussion I opened https://github.com/encode/httpcore/discussions/893

MarkusSintonen:
> Looks great, thanks so much.
> 
> Naming... should we use `AsyncIOStream` and `AsyncIOBackend` or `AsyncioStream` and `AsyncioBackend`?

No problem! Its now renamed to `AsyncIOBackend`. Thanks

docs/network-backends.md
```diff
@@ -73,8 +73,29 @@ while True:
 
 If we're working with an `async` codebase, then we need to select a different backend.
 
-The `httpcore.AnyIOBackend` is suitable for usage if you're running under `asyncio`. This is a networking backend implemented using [the `anyio` package](https://anyio.readthedocs.io/en/3.x/).
+These `async` network backends are available:
+- `httpcore.AsyncioBackend` This networking backend is implemented using Pythons native `asyncio`.
```
tomchristie, line 77:
```suggestion
- `httpcore.AsyncIOBackend` This networking backend is implemented using Pythons native `asyncio`.
```

docs/network-backends.md
```diff
@@ -73,8 +73,29 @@ while True:
...
+- `httpcore.AnyIOBackend` This is implemented using [the `anyio` package](https://anyio.readthedocs.io/en/3.x/).
+- `httpcore.TrioBackend` This is implemented using [`trio`](https://trio.readthedocs.io/en/stable/).
 
+Currently by default `AnyIOBackend` is used when running with `asyncio` (this may change).
+`TrioBackend` is used by default when running with `trio`.
+
+Using `httpcore.AsyncioBackend`:
```
tomchristie, line 84:
```suggestion
Using `httpcore.AsyncIOBackend`:
```

docs/network-backends.md
```diff
@@ -73,8 +73,29 @@ while True:
...
+```python
+import httpcore
+import asyncio
+
+async def main():
+    network_backend = httpcore.AsyncioBackend()
```
tomchristie, line 90:
```suggestion
    network_backend = httpcore.AsyncIOBackend()
```

tomchristie:
Fantastic, appreciate your time on this @MarkusSintonen.
Following `SEMVER` we'll want this in a `1.1.0` release. (New functionality)

We could either get #957 released first then merge this, or include this now and bump the version.

MarkusSintonen:
@tomchristie no problem! Would it be possible to include also the asyncio-based synchronization changes as part of this?

httpcore/_backends/asyncio.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+class AsyncIOStream(AsyncNetworkStream):
+    def __init__(
+        self, stream_reader: asyncio.StreamReader, stream_writer: asyncio.StreamWriter
+    ):
+        self._stream_reader = stream_reader
+        self._stream_writer = stream_writer
+        self._read_lock = asyncio.Lock()
+        self._write_lock = asyncio.Lock()
```
tomchristie, line 26: We don't use locking at the stream layer, let's remove these.

(I think we had locking at this layer in the past, the locking is now handled at the `HTTP11Connection`/`HTTP2Connection` layer)

tomchristie, line 26:
```suggestion
```

httpcore/_backends/asyncio.py
```diff
@@ -0,0 +1,228 @@
...
+                    if "resume_reading" in str(exc):
+                        # Python's asyncio has a bug that can occur when a
+                        # connection has been closed, while it is paused.
+                        # See: https://github.com/encode/httpx/issues/1213
+                        #
+                        # Returning an empty byte-string to indicate connection
+                        # close will eventually raise an httpcore.RemoteProtocolError
+                        # to the user when this goes through our HTTP parsing layer.
+                        return b""
+                    raise
```
tomchristie, line 95:
```suggestion
        with map_exceptions(exc_map):
            try:
                return await asyncio.wait_for(
                    self._stream_reader.read(max_bytes), timeout
                )
            except AttributeError as exc:  # pragma: nocover
                if "resume_reading" in str(exc):
                    # Python's asyncio has a bug that can occur when a
                    # connection has been closed, while it is paused.
                    # See: https://github.com/encode/httpx/issues/1213
                    #
                    # Returning an empty byte-string to indicate connection
                    # close will eventually raise an httpcore.RemoteProtocolError
                    # to the user when this goes through our HTTP parsing layer.
                    return b""
                raise
```

httpcore/_backends/asyncio.py
```diff
@@ -0,0 +1,228 @@
...
+            return
+
+        exc_map: Dict[Type[Exception], Type[Exception]] = {
+            asyncio.TimeoutError: WriteTimeout,
+            OSError: WriteError,
+        }
+        async with self._write_lock:
+            with map_exceptions(exc_map):
+                self._stream_writer.write(data)
+                return await asyncio.wait_for(self._stream_writer.drain(), timeout)
```
tomchristie, line 108:
```suggestion
        with map_exceptions(exc_map):
            self._stream_writer.write(data)
            return await asyncio.wait_for(self._stream_writer.drain(), timeout)
```

httpcore/_backends/asyncio.py
```diff
@@ -0,0 +1,228 @@
...
+            try:
+                self._stream_writer.close()
+                if is_ssl:
+                    # Give the connection a chance to write any data in the buffer,
+                    # and then forcibly tear down the SSL connection.
+                    await asyncio.sleep(0)
+                    self._stream_writer.transport.abort()
+                await self._stream_writer.wait_closed()
+            except OSError:  # pragma: nocover
+                pass
```
tomchristie, line 137:
```suggestion
        try:
            self._stream_writer.close()
            if is_ssl:
                # Give the connection a chance to write any data in the buffer,
                # and then forcibly tear down the SSL connection.
                await asyncio.sleep(0)
                self._stream_writer.transport.abort()
            await self._stream_writer.wait_closed()
        except OSError:  # pragma: nocover
            pass
```

tomchristie:
> Would it be possible to include also the asyncio-based synchronization changes as part of this?

Thanks for the ask, let's not do that. Mixing too many different concerns.
This is a self-contained PR and we can deal with it in isolation.

httpcore/_backends/asyncio.py
MarkusSintonen: Done

httpcore/_backends/asyncio.py
MarkusSintonen: Done

httpcore/_backends/asyncio.py
MarkusSintonen: Done

httpcore/_backends/asyncio.py
MarkusSintonen: Done

MarkusSintonen: Done, they are now gone

PR #944, opened 2024-08-19
fix(h2_connection_reset): add stream reset when client cancel the req…

### LINKED ISSUE #943: Missing client stream reset in h2 connection
### Discussed in https://github.com/encode/httpcore/discussions/941

<div type='discussions-op-text'>

<sup>Originally posted by **MarkLux** August  9, 2024</sup>
**Background**

1. Using Http2Connection with a high concurrency stream num (e.g over 400 stream_ids reusing single connection) 
2. Each stream is waiting for SSE events (assume that there are many  server events to receive for each stream)
3. Some of the streams cancelled by client (Not ALL)

**Phenomenon**
1. The Cancelled streams will be closed
2. But, the Server is still sending SSE events for that stream (because server doesn't receive any signal to indicate the stream is closed, and the connection is still alive because  other streams are using)
3. The rest streams on this connection, will be blocked in this loop:

```
# http2.py, class AsyncHTTP2Connection
# ...
async def _receive_stream_event(
        self, request: Request, stream_id: int
    ) -> typing.Union[
        h2.events.ResponseReceived, h2.events.DataReceived, h2.events.StreamEnded
    ]:
        while not self._events.get(stream_id): 
            # REST STREAM ARE BLOCKED HERE
            # for each loop, the _receive_events give an events that belongs to other stream (those we closed, but server didn't know, keep sending)
            await self._receive_events(request, stream_id)
        event = self._events[stream_id].pop(0)
        if isinstance(event, h2.events.StreamReset):
            raise RemoteProtocolError(event)
        return event
```

we made some experiments and the following logs were observed

img_v3_02dj_93f91287-7a04-4c02-8089-1bb14c957c5g

**Solution**

reset the stream when client cancelled the stream, here is a demo (have made experiments to prove it works)

```
# http2.py, class AsyncHTTP2Connection

async def _response_closed(self, stream_id: int) -> None:
        await self._max_streams_semaphore.release()
        del self._events[stream_id]
        # ADD
        stream = self._h2_state._get_stream_by_id(stream_id=stream_id)
        if stream and not stream.closed:
            # send a RstStream with Cancel Code 
            self._h2_state.reset_stream(stream_id=stream_id, error_code=8)
        # ADD END
        async with self._state_lock:
            if self._connection_terminated and not self._events:
                await self.aclose()
```

----

We referred some h2 libraries of other language, and found that the 'client stream reset when cancelled' behavior were both implemented.

e.g.1: [go net http2 library](https://github.com/golang/net/blob/master/http2/transport.go#L1643)
e.g.2: [rust h2 library](https://github.com/hyperium/h2/blob/master/src/proto/streams/streams.rs#L1503)
</div>

MarkLux:
# Summary

add the missing connection reset of h2 connection when client cancel the request.

fix #943

in discussion with #941

### DIFF
--- a/httpcore/_async/http2.py
+++ b/httpcore/_async/http2.py
@@ -401,6 +401,9 @@ class AsyncHTTP2Connection(AsyncConnectionInterface):
                     await self._max_streams_semaphore.acquire()
                     self._max_streams -= 1
 
+    async def _reset_steam(self, stream_id: int, error_code: int) -> None:
+        self._h2_state.reset_stream(stream_id=stream_id, error_code=error_code)
+
     async def _response_closed(self, stream_id: int) -> None:
         await self._max_streams_semaphore.release()
         del self._events[stream_id]
@@ -578,6 +581,12 @@ class HTTP2ConnectionByteStream:
             # we want to close the response (and possibly the connection)
             # before raising that exception.
             with AsyncShieldCancellation():
+                # need send cancel frame when the exception is not from remote peer.
+                if not isinstance(exc, RemoteProtocolError):
+                    await self._connection._reset_steam(
+                        stream_id=self._stream_id,
+                        error_code=h2.settings.ErrorCodes.CANCEL,
+                    )
                 await self.aclose()
             raise exc
 
--- a/httpcore/_sync/http2.py
+++ b/httpcore/_sync/http2.py
@@ -401,6 +401,9 @@ class HTTP2Connection(ConnectionInterface):
                     self._max_streams_semaphore.acquire()
                     self._max_streams -= 1
 
+    def _reset_steam(self, stream_id: int, error_code: int) -> None:
+        self._h2_state.reset_stream(stream_id=stream_id, error_code=error_code)
+
     def _response_closed(self, stream_id: int) -> None:
         self._max_streams_semaphore.release()
         del self._events[stream_id]
@@ -578,6 +581,12 @@ class HTTP2ConnectionByteStream:
             # we want to close the response (and possibly the connection)
             # before raising that exception.
             with ShieldCancellation():
+                # need send cancel frame when the exception is not from remote peer.
+                if not isinstance(exc, RemoteProtocolError):
+                    self._connection._reset_steam(
+                        stream_id=self._stream_id,
+                        error_code=h2.settings.ErrorCodes.CANCEL,
+                    )
                 self.close()
             raise exc
 


httpcore/_async/http2.py
```diff
@@ -404,6 +404,14 @@ async def _receive_remote_settings_change(self, event: h2.events.Event) -> None:
     async def _response_closed(self, stream_id: int) -> None:
         await self._max_streams_semaphore.release()
         del self._events[stream_id]
+
+        stream = self._h2_state._get_stream_by_id(stream_id=stream_id)
```
tomchristie, line 408: Is the usage of private API on the `H2Connection` instance avoidable?

tomchristie (commented):
Thanks, seems like a good start. Make sure to run `scripts/unasync.py` to also apply the changes to the `httpcore/_sync/http2.py` module.

https://github.com/encode/httpcore/actions/runs/10464774808/job/28978912844#step:5:26

(Related... maybe we could change the build logs when `unasync --check` fails, so that there's clearer feedback to the developer.)

httpcore/_async/http2.py
MarkLux: yeah, I think the `stream_reset` should only be called in `aclose`(where the request is cancelled by client), thus we no longer need to check the stream closed here. 

Testing code now, an improved version will be pushed later

httpcore/_async/http2.py
```diff
@@ -578,12 +582,18 @@ async def __aiter__(self) -> typing.AsyncIterator[bytes]:
             # we want to close the response (and possibly the connection)
             # before raising that exception.
             with AsyncShieldCancellation():
-                await self.aclose()
+                # close the stream with cancel
+                await self.aclose(cancel_stream=isinstance(exc, asyncio.exceptions.CancelledError))
```
MarkLux, line 586: @tomchristie If we don't want to using `_h2_state._get_stream_by_id()` to check if the stream closed before send rst stream, we need to figure out if the stream closed by 'cancel' action.

thus we have to judge the exc type here, and due to the multi backend impls (asyncio, trio, etc.), the condition judgements can be very complex with many `isinstance` calls , it's ugly and hard to maintain afterwards : ( 

so maybe it's more appropriate to just adding a stream closed condition before sending rst stream with CANCEL ? (the only concern is that we called the protected method of h2)

MarkLux: ignore this one, found solution by referring the [golang impl](https://github.com/golang/net/blob/4542a42604cd159f1adb93c58368079ae37b3bf6/http2/transport.go#L1639)

judge if the error caused by remote, that's all we need to do : )

httpcore/_async/http2.py
MarkLux: @tomchristie Done. Could you check it again?

MarkLux:
sorry to bother but we are really in a hurry to fix this one, could you have a review again? @tomchristie

MarkLux:
https://github.com/langchain-ai/httpcore/pull/5 

approved by langchain community now, shall we have any further progress on this one? @tomchristie

httpcore/_sync/http2.py
```diff
@@ -572,6 +575,12 @@ def __iter__(self) -> typing.Iterator[bytes]:
             # we want to close the response (and possibly the connection)
             # before raising that exception.
             with ShieldCancellation():
+                # need send cancel frame when the exception is not from remote peer.
+                if not isinstance(exc, RemoteProtocolError):
```
tomchristie, line 579: > need send cancel frame when the exception is not from remote peer.

It's not obvious to me that `not isinstance(exc, RemoteProtocolError)` is correct/sufficient here.

Should this instead be something like *just wrapping the yield*?...

```python
try:
    yield chunk
except Exception as exc:
    self._connection._reset_steam(
        stream_id=self._stream_id,
        error_code=h2.settings.ErrorCodes.CANCEL,
    )
    raise exc
```

tomchristie:
> approved by langchain community now, shall we have any further progress on this one?

Thanks @MarkLux - apologies for the delay it's a bit of a complex one to review.

Thoughts on my review comment?

mabrowning:
This PR or something like it is still needed. Even with the latest version of httpcore, if the client closes the response before the server is done, the client does not send any `RST_STREAM` frames; the server will continue to generate and stream data and the client will ignore it.

httpcore/_sync/http2.py
mabrowning: `BaseException` covers `asyncio.CancelledError` as well, which should also reset the stream.

httpcore/_sync/http2.py
```diff
@@ -397,6 +397,9 @@ def _receive_remote_settings_change(self, event: h2.events.Event) -> None:
                     self._max_streams_semaphore.acquire()
                     self._max_streams -= 1
 
+    def _reset_steam(self, stream_id: int, error_code: int) -> None:
+        self._h2_state.reset_stream(stream_id=stream_id, error_code=error_code)
```
mabrowning, line 401: After updating the local `h2_state`, we should write the updated bytes on the wire, otherwise its reliant on some (possibly much) later write (like initiating a new request or a keep-alive frame) sending the reset.

PR #953, opened 2024-09-19
Use AnyIO with `fast_acquire` synchronization 

MarkusSintonen:
# Summary

Related to https://github.com/agronholm/anyio/pull/761 and https://github.com/encode/httpcore/pull/922#issuecomment-2325139599

This uses AnyIOs `fast_acquire`-mode in its synchronization primitives (instead of replacing AnyIO with vanilla asyncio). Which gives a big performance boost (see above comment lint).

# Checklist

### DIFF
--- a/httpcore/_async/http2.py
+++ b/httpcore/_async/http2.py
@@ -67,10 +67,12 @@ class AsyncHTTP2Connection(AsyncConnectionInterface):
         # Mapping from stream ID to response stream events.
         self._events: dict[
             int,
-            h2.events.ResponseReceived
-            | h2.events.DataReceived
-            | h2.events.StreamEnded
-            | h2.events.StreamReset,
+            list[
+                h2.events.ResponseReceived
+                | h2.events.DataReceived
+                | h2.events.StreamEnded
+                | h2.events.StreamReset,
+            ],
         ] = {}
 
         # Connection terminated events are stored as state since
@@ -102,9 +104,11 @@ class AsyncHTTP2Connection(AsyncConnectionInterface):
         async with self._init_lock:
             if not self._sent_connection_init:
                 try:
-                    kwargs = {"request": request}
-                    async with Trace("send_connection_init", logger, request, kwargs):
-                        await self._send_connection_init(**kwargs)
+                    sci_kwargs = {"request": request}
+                    async with Trace(
+                        "send_connection_init", logger, request, sci_kwargs
+                    ):
+                        await self._send_connection_init(**sci_kwargs)
                 except BaseException as exc:
                     with AsyncShieldCancellation():
                         await self.aclose()
@@ -293,6 +297,7 @@ class AsyncHTTP2Connection(AsyncConnectionInterface):
 
         status_code = 200
         headers = []
+        assert event.headers is not None
         for k, v in event.headers:
             if k == b":status":
                 status_code = int(v.decode("ascii", errors="ignore"))
@@ -310,6 +315,8 @@ class AsyncHTTP2Connection(AsyncConnectionInterface):
         while True:
             event = await self._receive_stream_event(request, stream_id)
             if isinstance(event, h2.events.DataReceived):
+                assert event.flow_controlled_length is not None
+                assert event.data is not None
                 amount = event.flow_controlled_length
                 self._h2_state.acknowledge_received_data(amount, stream_id)
                 await self._write_outgoing_data(request)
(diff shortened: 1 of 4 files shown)


graingert (approved):

graingert:
nice this looks good!

This feature was added to anyio specifically for httpcore, so I'm going to merge it even though I'm the only review

tomchristie:
@graingert No. Please don't do that on this PR.

This was my latest comment regarding this... https://github.com/encode/httpcore/pull/922#issuecomment-2360556934

Here's a comment that'd make me wary of jumping onto `fast_acquire=True`... https://github.com/agronholm/anyio/pull/761#pullrequestreview-2279500760

graingert:
Happy for a revert, does https://github.com/encode/httpcore/pull/927 remove a lock on the async code path and therefore make this PR redundant?

tomchristie:
I'm not talking about #927 or directing my attention there at the moment.

tomchristie:
Revert PR at #1002.

Nivyaal-zenity:
When will the tag related to this change here and on httpx will be relaesed?

ofek:
I'm confused about what happened here; what was the technical issue?

zanieb:
@ofek https://github.com/encode/httpcore/pull/953#issuecomment-2758988306 / https://github.com/agronholm/anyio/pull/761#pullrequestreview-2279500760

ofek:
Ah I missed the second comment, thanks! I don't do enough `async` in Python to know what that means but I'll take a deeper dive eventually because such performance improvements always intrigue me.

agronholm:
> @graingert No. Please don't do that on this PR.
> 
> This was my latest comment regarding this... [#922 (comment)](https://github.com/encode/httpcore/pull/922#issuecomment-2360556934)
> 
> Here's a comment that'd make me wary of jumping onto `fast_acquire=True`... [agronholm/anyio#761 (review)](https://github.com/agronholm/anyio/pull/761#pullrequestreview-2279500760)

Can you explain this to me, as it didn't make a lick of sense to me? Was this about `fast_acquire=True` skipping the checkpoint? `asyncio.Lock` never had that checkpoint to begin with, so what do you suppose is gained by reverting the commit?

lovelydinosaur:
>  what do you suppose is gained by reverting the commit?

@agronholm So. The priority at the time was bumping a release from 1.0.7 to 1.0.8 in order to resolve https://github.com/encode/httpcore/pull/1005.

Mixing in other concerns introduces risk, rather than just *prioritising the necessary fix for Python 3.14.0a6*.

---

Also I've been doing some background work on a significantly simplified implementation. I've got a connection pool approach that's smaller, more readable, and easier to reason about. It also happens to result in faster & more stable response characteristics under load. (significant improvement vs. current httpx, marginal improvement vs. aiohttpx)

I'm not ready to share the work publicly yet, tho will update shortly.

agronholm:
> > what do you suppose is gained by reverting the commit?
> 
> @agronholm So. The priority at the time was bumping a release from 1.0.7 to 1.0.8 in order to resolve #1005.
> 
> Mixing in other concerns introduces risk, rather than just _prioritising the necessary fix for Python 3.14.0a6_.
> 
> Also I've been doing some background work on a significantly simplified implementation. I've got a connection pool approach that's smaller, more readable, and easier to reason about. It also happens to result in faster & more stable response characteristics under load. (significant improvement vs. current httpx, marginal improvement vs. aiohttpx)
> 
> I'm not ready to share the work publicly yet, tho will update shortly.

OK, thanks for the quick response. I was previously left with the impression that you had reservations about `fast_acquire`, but perhaps I misunderstood.

lovelydinosaur:
> I was previously left with the impression that you had reservations about fast_acquire, but perhaps I misunderstood.

I don't really know enough about `fast_acquire` to take an informed stance on that.

(Incidentally, thanks so much for your work on `anyio`, and helping bring structured concurrency patterns to Python's asyncio.)

agronholm:
Yeah, so the summary is that with `fast_acquire=True`, AnyIO locks work pretty much the same way as asyncio locks (hence the performance parity). It does that by skiping a checkpoint, which on Trio is considered faux pas where everything with `async with`, `async for` or `await` should go through a checkpoint, but on asyncio there are no such considerations. Basically asyncio cuts corners and is faster because of that. `fast_acquire=True` just cuts corners the same way `asyncio.Lock` does.

PR #956, opened 2024-09-27
Update trio dependency.

tomchristie:
Relax `trio` dependency pinning upper bound to `<1.0`.

Trio's versioning has been extremely careful.
I think we can feel confident enough with this.

Prompted by https://github.com/encode/httpcore/pull/922#issuecomment-2361111584 - thanks @agronholm.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ socks = [
     "socksio==1.*",
 ]
 trio = [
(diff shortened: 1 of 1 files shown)


tomchristie:
The alternative would be to push up to 0.27.x.

I'd also be happy with that if prefered.

gvbgduh (approved):
Lgtm

PR #972, opened 2024-10-25
default port for the socks5h scheme

mateuszlitwin:
# Summary

This change is needed to unblock https://github.com/encode/httpx/pull/3178 and support `socks5h` scheme (server-side hostname resolution socks5 proxy). Support for the `socks5h` is needed to ensure global environment variables like `all_proxy="socks5h://..."` can be used reliably with different programs (some of which are httpx based).

# Checklist

### DIFF
--- a/httpcore/_models.py
+++ b/httpcore/_models.py
@@ -284,6 +284,7 @@ class URL:
             b"ws": 80,
             b"wss": 443,
             b"socks5": 1080,
+            b"socks5h": 1080,
         }[self.scheme]
         return Origin(
             scheme=self.scheme, host=self.host, port=self.port or default_port
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -43,11 +43,15 @@ def test_url_cannot_include_unicode_strings():
 
 def test_url_origin_socks5():
     url = httpcore.URL("socks5://127.0.0.1")
-    assert url.origin == httpcore.Origin(scheme=b"socks5", host=b"127.0.0.1", port=1080)
+    assert url.origin == httpcore.Origin(
+        scheme=b"socks5", host=b"127.0.0.1", port=1080
+    )
     assert str(url.origin) == "socks5://127.0.0.1:1080"
 
     url = httpcore.URL("socks5h://127.0.0.1")
-    assert url.origin == httpcore.Origin(scheme=b"socks5h", host=b"127.0.0.1", port=1080)
+    assert url.origin == httpcore.Origin(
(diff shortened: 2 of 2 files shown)


mateuszlitwin:
Can add tests sync and async tests for `httpcore.SOCKSProxy` if needed.

tomchristie:
Thanks. Do you know implementation would need to change in establishing the connection in `socks_proxy.py`?

mateuszlitwin:
I don't think so:

- I tested `test_socks_proxy.py` with socks5h scheme and they work.
- I tested https://github.com/encode/httpx/pull/3178 with e2e on real proxies.

tomchristie:
We do already use "proxy resolves the hostname" behaviour, so our SOCKS support is "sock5h".

https://superuser.com/questions/1762341/does-chromium-not-support-socks5h-with-the-h-in-the-end

It's not obvious that we should add a synonym to the protocol name. (Staying in line with chromium's behaviour here seems reasonable.)

mateuszlitwin:
It is true that `httpx` already does proxy side hostname resolution, but other tools do not. For instance, `curl` does not for the `socks5` scheme. This makes it impossible to set global env e.g. `all_proxy="socks5h://127.0.0.1:1080` to correctly support proxing in multiple tools.

tomchristie:
Okay yep.
Looks like there's a linter failure needs resolving...

```
+ ruff format httpcore tests --diff
1 file would be reformatted, 37 files already formatted
--- tests/test_models.py
+++ tests/test_models.py
@@ -47,7 +47,9 @@
     assert str(url.origin) == "socks5://[12](https://github.com/encode/httpcore/actions/runs/11522347142/job/32078024585?pr=972#step:5:13)7.0.0.1:1080"
 
     url = httpcore.URL("socks5h://127.0.0.1")
-    assert url.origin == httpcore.Origin(scheme=b"socks5h", host=b"127.0.0.1", port=1080)
+    assert url.origin == httpcore.Origin(
+        scheme=b"socks5h", host=b"127.0.0.1", port=1080
+    )
     assert str(url.origin) == "socks5h://127.0.0.1:1080"
```

mateuszlitwin:
I think you are looking at the wrong revision?

tomchristie (approved):
Okay, yep.
Thanks! 🙏🏼

mateuszlitwin:
@tomchristie Could you revisit https://github.com/encode/httpx/pull/3178 after merging this PR? I think this might have been the blocker.

PR #986, opened 2024-12-18
Fix ConnectionPool state after async cancellation

grynko:
# Summary

The PR fixes `ConnectionPool` state after exceptions/async cancellations.

# Root cause analysis

`ConnectionPool` removes only connections which are:
1. closed
2. idle
3. expired

At the same time, some implementations of `AsyncConnectionInterface`, e.g. `AsyncHTTPConnection` don't make actual connection on construction, but are left in semi-valid state with `_connection = None`. Thus, `is_closed()`/`is_idle`/`has_expired()` return `False` before actual connection is established or failed during the call to `AsyncHTTPConnection.handle_async_request`.

Let's take a look at `ConnectionPool.handle_async_request` method:
```python
try:
    ...
        with self._optional_thread_lock:
            # [POINT 1] the `AsyncHTTPConnection` may be created to handle the request:
            closing = self._assign_requests_to_connections()
        await self._close_connections(closing)

        connection = await pool_request.wait_for_connection(timeout=timeout)

        try:
            # [POINT 2]  the `AsyncHTTPConnection` may create an actual connection:
            response = await connection.handle_async_request(
                pool_request.request
            )
        ...
```
Please, pay attention, that if the request task is being cancelled between `[POINT 1]` and `[POINT 2]`, the Schrödinger connection (that is neither opened, nor closed) stays in the pool forever. In the long run (that depends on the load and pool max size), all new requests are refused by the pool, raising `PoolTimeout` for any request.

Please, see the `test_connection_pool_cancellation_during_waiting_for_connection()` for a reproducible example.

# Solution

Introduce new method of `AsyncConnectionInterface` - `is_connected()`. It's not just `not is_closed()`, but returns `False` for such semi-constructed connections, allowing the `ConnectionPool` to collect them.

# Checklist


drozdvadym:
Thanks for working on this! We’ve encountered similar issues with connections getting stuck in the pool after async cancellations, so it’s great to see this being addressed. This fix will be helpful—I'm looking forward to it getting merged!

PR #999, opened 2025-03-26
fix ci

graingert:
# Summary

# Checklist


tomchristie:
What's the background here?... What changed to break things?

graingert:
h2 typing changed and among other things h2.events.ResponseReceived etc went from Any to             h2.events.ResponseReceived

tomchristie (approved):

PR #1000, opened 2025-03-26
Fix max_keepalive_connections

blazewicz:
# Summary

This commit fixes a bug in both sync and async connection pools where idle connections were dropped from the pool even when max_keepalive_connections limit has not been reached. This happened because the check compared this setting's value to the total number of connections, not only of the idle ones.

# Checklist


tests/_async/test_connection_pool.py
```diff
@@ -79,28 +79,33 @@ async def test_connection_pool_with_keepalive():
         )
 
         # Sending a request to a different origin will not reuse the existing IDLE connection.
-        async with pool.stream("GET", "http://example.com/") as response:
+        async with pool.stream("GET", "http://example.com/") as response_1, pool.stream(
+            "GET", "http://example.com/"
+        ) as response_2:
             info = [repr(c) for c in pool.connections]
             assert info == [
                 "<AsyncHTTPConnection ['https://example.com:443', HTTP/1.1, IDLE, Request Count: 2]>",
```
blazewicz, line 87: This connection is dropped without the fix.

tests/_sync/test_connection_pool.py
```diff
@@ -79,28 +79,33 @@ def test_connection_pool_with_keepalive():
         )
 
         # Sending a request to a different origin will not reuse the existing IDLE connection.
-        with pool.stream("GET", "http://example.com/") as response:
+        with pool.stream("GET", "http://example.com/") as response_1, pool.stream(
+            "GET", "http://example.com/"
+        ) as response_2:
             info = [repr(c) for c in pool.connections]
             assert info == [
                 "<HTTPConnection ['https://example.com:443', HTTP/1.1, IDLE, Request Count: 2]>",
```
blazewicz, line 87: This connection is dropped without the fix.

zanieb (approved):
This does seem like a fairly straightforward oversight in the implementation.

lovelydinosaur:
Ah yeah apologies... this *is* a nice simple fix, and worth getting in.

springmeyer:
Will this be released soon in a formal release? Perhaps a 1.0.10?

PR #965, opened 2024-10-09
Move benchmark to scripts.

T-256:
# Summary

- closes #964

Issue #964, opened 2024-10-09
Drop benchmarking requirements from CI.


tomchristie:
We don't run the benchmarking in CI, and I'd prefer that our standard CI runs are as sharp as possible.

Perhaps just move them directly into the benchmarking script, to help keep that standard setup minimal.

PR #1001, opened 2025-03-26
ci: test Python 3.13 and 3.14.0-alpha - 3.14

JohnVillalovos:
Also stop testing Python 3.8 as it is End-of-Life as of 2024-10-07. https://devguide.python.org/versions/

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14.0-alpha - 3.14"]
 
     steps:
       - uses: "actions/checkout@v4"
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14.0-alpha - 3.14"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
 
     steps:
       - uses: "actions/checkout@v4"


tomchristie:
Okay, shall we see about getting these passing?
The behaviour change is identified in https://github.com/encode/httpcore/discussions/995

One resolution here would be to change this line...

https://github.com/encode/httpcore/blob/a1735520e3826ccc861cdadf3e692abfbb19ac6a/httpcore/__init__.py#L138

To add an exclusion for `SOCKET_OPTION`.

JohnVillalovos:
> Okay, shall we see about getting these passing? The behaviour change is identified in #995
> 
> One resolution here would be to change this line...
> 
> https://github.com/encode/httpcore/blob/a1735520e3826ccc861cdadf3e692abfbb19ac6a/httpcore/__init__.py#L138
> 
> To add an exclusion for `SOCKET_OPTION`.

First will likely need #999 or something similar merged to get the other parts of the CI to pass.

.github/workflows/test-suite.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14.0-alpha - 3.14"]
```
hugovk, line 17: We can simplify this (`allow-prereleases: true` is set below):

```suggestion
        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
```

hugovk:
The next failure in the yak stack can be fixed by setting the `PYO3_USE_ABI3_FORWARD_COMPATIBILITY: 1` env var to allow PYO3 to run on Python 3.14 pre-releases, for example:

* https://github.com/hugovk/httpcore/commit/6ef32471c4c1998d25953098b6935b48fea767d3
* https://github.com/hugovk/httpcore/actions/runs/14191922956/job/39758340950

(`FORCE_COLOR: 1` is optional and adds colour to the CI output, which makes errors easier to find in long logs.)

Then the next failure is because it's trying to build Pillow (a dependency of a dependency) from source because doesn't have wheels yet for 3.14, and most likely they'll be available in the 1st July release (the first one after beta):

> The headers or library files could not be found for jpeg,
> a required dependency when compiling Pillow from source.

* https://github.com/hugovk/httpcore/actions/runs/14191922956/job/39758340950

This can be worked around by temporarily installing libjpeg8-dev, something like this (it could be restricted to just the 3.14 build):

* https://github.com/hugovk/httpcore/commit/bd7d194727db2e357a964344f1a91a1687871caf

Pillow then builds:

* https://github.com/hugovk/httpcore/actions/runs/14194081780/job/39765107948

And that gets us to the `typing.Union` failure from:

* https://github.com/encode/httpcore/discussions/995

tomchristie:
@hugovk Rather than resolve those weirdly specific alpha buglettes, I'd suggest we drop the "benchmarking requirements" as suggested in https://github.com/encode/httpcore/issues/964. They're breaking our builds.

hugovk:
Good idea, there's an approved PR for that issue awaiting merge: https://github.com/encode/httpcore/pull/965.

JohnVillalovos:
I'm going to assume this is out-dated.

PR #1002, opened 2025-03-27
Revert "Use AnyIO with `fast_acquire` synchronization "

tomchristie:
Reverts encode/httpcore#953

### DIFF
--- a/httpcore/_synchronization.py
+++ b/httpcore/_synchronization.py
@@ -65,7 +65,7 @@ class AsyncLock:
         if self._backend == "trio":
             self._trio_lock = trio.Lock()
         elif self._backend == "asyncio":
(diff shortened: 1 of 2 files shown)


MarkusSintonen:
Why?

methane:
> There's smarter simpler approaches to connection pooling: tho it'll need a more careful engineering approach.

How much time would it take to implement that smarter approach? The performance issues with httpx are significant and affecting its popularity. https://github.com/openai/openai-python/issues/1596

MarkusSintonen:
> How much time would it take to implement that smarter approach? The performance issues with httpx are significant and affecting its popularity. [openai/openai-python#1596](https://github.com/openai/openai-python/issues/1596)

Httpx/httpcore projects are pretty much dead. Better to just get rid of them. Fixes are not coming (any time in near future)

Issue #1015, opened 2025-06-06
Timeouts are not applied during socks5 initialization.


Brranton:
I encountered a problem when using socks5 proxy.
Sometimes the server would not get a response during socks5 initialization, causing the thread to hang when using httpx.Client.
Based on this I checked the initialization of socks5 connection. And found that timeouts are not applied during socks5 initialization.

HansBrende:
I've encountered exactly the same issue when using the `socks5h` proxy and httpx/httpcore/openai.

The request hung for 25 minutes before finally failing with:

> Traceback (most recent call last):
  File ".../python3.12/site-packages/httpx/_transports/default.py", line 101, in map_httpcore_exceptions
    yield
  File ".../python3.12/site-packages/httpx/_transports/default.py", line 394, in handle_async_request
    resp = await self._pool.handle_async_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../python3.12/site-packages/httpcore/_async/connection_pool.py", line 256, in handle_async_request
    raise exc from None
  File ".../python3.12/site-packages/httpcore/_async/connection_pool.py", line 236, in handle_async_request
    response = await connection.handle_async_request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../python3.12/site-packages/httpcore/_async/socks_proxy.py", line 295, in handle_async_request
    raise exc
  File ".../python3.12/site-packages/httpcore/_async/socks_proxy.py", line 231, in handle_async_request
    stream = await self._network_backend.connect_tcp(**kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../python3.12/site-packages/httpcore/_backends/auto.py", line 31, in connect_tcp
    return await self._backend.connect_tcp(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../python3.12/site-packages/httpcore/_backends/anyio.py", line 113, in connect_tcp
    with map_exceptions(exc_map):
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/hansbrende/.local/share/uv/python/cpython-3.12.10-macos-aarch64-none/lib/python3.12/contextlib.py", line 158, in __exit__
    self.gen.throw(value)
[...]

PR #1017, opened 2025-06-06
added timeouts to socks5 connection initialization

Brranton:
# Summary
added timeouts to socks5 connection initialization
# Checklist


Brranton:
this fixes issue #1015

dmittov:
Thanks! I faced this issue today.
Socks proxy doesn't respond and the code stuck on awaiting stream.read much longer than I expected passing the explicit timeout to the request.

I hope this fix is merges soon. Looking forward.

PR #1018, opened 2025-06-10
prioritize h2

soundofspace:
# Summary

When using h2 give it priority in tls alpn. In general order of tls alpn is important and servers mostly choose the first version that they support, by setting http1.1 first most server will actually negotiate http1.1

Also don't specify http1.1 if only h2 is enabled. This could otherwise lead to negotiating something that is not supported, which was especially bad when the wrong order was used.

Found this actually when forcing httpcore to use h2 only and then wireshark failing to decode traffic because tls negotiated http1.1 but we were actually using h2

# Checklist

### DIFF
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -143,7 +143,7 @@ class AsyncHTTPConnection(AsyncConnectionInterface):
                         if self._ssl_context is None
                         else self._ssl_context
                     )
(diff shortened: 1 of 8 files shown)


stijndcl:
@lovelydinosaur @zanieb sorry for pinging but any chance this one can get a quick review?

PR #1020, opened 2025-06-25
Catch exception if close() in exception manager, throw another except…

pprados:
Given this is a project maintained by volunteers, please read this template to not waste your time, or ours! 😁 -->

# Summary

If some scenario, to close the program,  the code catch en exception, and try to close itself. But, if this close generate and exception, the code crash with a stacktrace.
```
    async def __aiter__(self) -> typing.AsyncIterator[bytes]:
        try:
            async for part in self._stream:
                yield part
        except BaseException as exc:
            try:
                await self.aclose()
            except BaseException:
                pass
            raise exc from None
```

# Checklist


cclauss:
https://peps.python.org/pep-0008 contains clear advice to avoid catching BaseException.  Why should that be ignored?
* https://docs.astral.sh/ruff/rules/bare-except
* https://docs.astral.sh/ruff/rules/blind-except

PR #1030, opened 2025-09-02
Add Python 3.13 and 3.14 to test matrix

cclauss:
# Summary

Add Python 3.13 and 3.14rc2 to the test suite.

# Checklist


.github/workflows/test-suite.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.x", "3.14"]
```
zanieb: Why "3.x"?

cclauss: Tests should tell us if things are broken on the latest stable version of Python before users do.
* https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-input
> [x-ranges](https://github.com/npm/node-semver#x-ranges-12x-1x-12-) to specify the latest stable version of Python (for the specified major version):

zanieb: I'm not sure why we wouldn't just explicitly test 3.13 here? I don't think `3.x` adds any clarity and then we could lose test coverage for 3.13 unexpectedly?

cclauss: 3.x is not meant to add clarity.  It is meant to ___automatically___ run tests on the latest stable version of Python.

Automated testing has not run on Python 3.13.0, 3.13.1, 3.13.2, 3.13.4, 3.13.5, and 3.13.6.  Why would we want to repeat that trend for future Python production releases?  Must users always signal us of breakage, or can automated testing warn us instead?

The value of doing things this way becomes clearer when we see that pytest cannot run at all on
* encode/httpx#3645

lovelydinosaur:
```suggestion
        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
```

lovelydinosaur (approved):

.github/workflows/test-suite.yml
zanieb: Maybe you don't intend it to add clarity but it's a problem that it actively makes things less clear. New minor Python versions are not a frequent occurrence, it's fine to just update the matrix when support for a new version is added.

zanieb (approved):

Issue #1029, opened 2025-08-31
Very slow when sending large amounts of data using `httpx.Client` Post, either with multi-threading or a single thread


xiexianbin:
This is a consistently reproducible issue that causes significant delays when sending data across multiple threads.

**Scenario Reproduction:**

  * `server.py`: A server implemented with FastAPI to receive data.

```
# server.py
import time

from fastapi import FastAPI, Request
from loguru import logger


app = FastAPI()

@app.post("/upload")
async def upload_file(request: Request):
    req_id = request.headers.get("x-request-id", "N/A")
    start_time_str = request.headers.get("x-start-time", "0")
    content_length = request.headers.get("content-length", "0")

    logger.info(f"Request [{req_id}] received, Headers: x-start-time={start_time_str}, content-length={content_length}, attempting to read the request body...")

    received_size = 0
    start_read_time = time.time()

    async for chunk in request.stream():
        received_size += len(chunk)

    end_read_time = time.time()
    read_duration = end_read_time - start_read_time

    logger.debug(
        f"Request [{req_id}] finished reading body. "
        f"Total size received: {received_size / (1024*1024):.2f} MB. "
        f"Read duration: {read_duration:.2f} seconds."
    )

    return {
        "message": "Upload received successfully after a long wait.",
        "request_id": req_id,
        "received_size": received_size
    }


if __name__ == "__main__":
    import uvicorn
    uvicorn.run(app, host="0.0.0.0", port=8000, workers=1)
```

  * `client.py`: Implements 2 threads to simultaneously post 2048MiB of data (the larger the data, the longer the lock duration) to `server.py`.

```
# client.py
import threading
import time
import os
import uuid

import httpx
from loguru import logger


SERVER_URL = "http://127.0.0.1:8000/upload"
NUM_THREADS = 2
DATA_SIZE_BYTES = 2 * 1024 * 1024 * 1024 # 2048 MiB
TIMEOUT_SECONDS = 120


class DataGenerator:
    def __init__(self, size):
        logger.info(f"Generating a {size / (1024*1024):.2f} MB data chunk in memory...")
        self._data = os.urandom(size)
        self._size = size
        self._pos = 0
        logger.info("Data generation complete.")

    def __iter__(self):
        self._pos = 0
        return self

    def __next__(self):
        if self._pos >= self._size:
            raise StopIteration
        chunk = self._data[self._pos:]
        self._pos = self._size
        return chunk

    def __len__(self):
        return self._size

[...]

PR #1038, opened 2025-09-23
Don't hold lock unless necessary in PoolByteStream.close()

leventov:
# Summary

Don't hold lock unless necessary in PoolByteStream.close().

This lock causes significant contention and slowdown in LiteLLM when used with streaming but _sync_ HTTP transport.

Testing: this patch was validated through load testing that decreased lock contention (validated with py-spy) and significantly decreased wall clock completion time of concurrent streaming LLM requests via LiteLLM.

# Checklist


ilovelinux:
Maybe related issue:
- #1029

Issue #45, opened 2020-03-11
ConnectionTerminated error_code:ErrorCodes.NO_ERROR


victoraugustolls:
With httpx using httpcore interface, started receiving the following error:

`<ConnectionTerminated error_code:ErrorCodes.NO_ERROR, last_stream_id:1999, additional_data:None>`

Don't know h2 specifics to say, but could `NO_ERROR` be used for some indication like graceful connection closing rather than raising an error?

tomchristie:
Do you know what the traceback is on this? Really we need to be catching that case in the HTTP/2 connection class and mapping it to a `NewConnectionRequired` exception. (The server is indicating that it has served the maximum number of streams that it'll allow on a connection, and that the client should reopen a new connection.)

victoraugustolls:
I had in the past, will try to get a new one.

tomchristie:
Looks like we deal with this correctly now, so I'm going to close it off, pending re-occurrence.

victoraugustolls:
Hi @tomchristie , so... it happened again!

Here is the traceback:

```2020-07-16T15:28:11.813905853Z   File "/usr/local/lib/python3.8/site-packages/httpx/_client.py", line 1147, in request
2020-07-16T15:28:11.813909653Z     response = await self.send(
2020-07-16T15:28:11.813913353Z   File "/usr/local/lib/python3.8/site-packages/httpx/_client.py", line 1168, in send
2020-07-16T15:28:11.813917153Z     response = await self.send_handling_redirects(
2020-07-16T15:28:11.813920953Z   File "/usr/local/lib/python3.8/site-packages/httpx/_client.py", line 1195, in send_handling_redirects
2020-07-16T15:28:11.813924853Z     response = await self.send_handling_auth(
2020-07-16T15:28:11.813929653Z   File "/usr/local/lib/python3.8/site-packages/httpx/_client.py", line 1232, in send_handling_auth
2020-07-16T15:28:11.813933653Z     response = await self.send_single_request(request, timeout)
2020-07-16T15:28:11.813937453Z   File "/usr/local/lib/python3.8/site-packages/httpx/_client.py", line 1264, in send_single_request
2020-07-16T15:28:11.813941353Z     ) = await transport.request(
2020-07-16T15:28:11.813944953Z   File "/usr/local/lib/python3.8/site-packages/httpcore/_async/connection_pool.py", line 152, in request
2020-07-16T15:28:11.813948953Z     response = await connection.request(
[...]

PR #1054, opened 2026-01-29
fix: handle http2 goaway race conditions

bayoumi17m:
# Fix HTTP/2 GOAWAY Race Conditions

⚠️GenAI was used in the development of this PR. It was all reviewed and tested manually by a human⚠️

## Problem

When using HTTPX with HTTP/2 under high concurrency, users encounter errors where connections are closed for streams beyond the `last_stream_id`. This forces applications to implement retry logic for what should be transparent library behavior.

The root cause is a set of race conditions between h2's state machine and httpcore's event processing:

1. **State machine race**: h2 transitions to CLOSED before httpcore processes the GOAWAY event
2. **Disconnect without context**: Server disconnects immediately after GOAWAY, raising generic `RemoteProtocolError("Server disconnected")` with no retry opportunity
3. **Insufficient exception context**: `ConnectionNotAvailable` didn't convey *why* the connection was unavailable, preventing informed retry decisions

A user proposed changing `>` to `>=` in the stream ID comparison ([discussion #1041](https://github.com/encode/httpcore/discussions/1041)), but this would violate [RFC 7540 Section 6.8](https://datatracker.ietf.org/doc/html/rfc7540#section-6.8):

> "The last stream identifier in the GOAWAY frame contains the highest-numbered stream identifier for which the sender of the GOAWAY frame might have taken some action on..."

A stream matching `last_stream_id` may have already been processed, making automatic retry unsafe for non-idempotent operations.

## Solution

This PR introduces proper GOAWAY handling that maintains RFC compliance while enabling safe automatic retries:  
 
### 1. New `ConnectionGoingAway` Exception 
  
A dedicated exception (subclass of `ConnectionNotAvailable`) that carries GOAWAY context:

- `last_stream_id` / `request_stream_id` for RFC 7540 stream comparison
- `error_code` to distinguish graceful shutdown (NO_ERROR) from errors
- `headers_sent` / `body_sent` to track request phase at time of GOAWAY
- Helper properties: `is_safe_to_retry`, `is_graceful_shutdown`, `may_have_side_effects`
  
This exception is exported publicly, allowing applications to implement idempotency-aware retry logic when needed. 

### 2. New `DRAINING` Connection State
  
Connections receiving a graceful GOAWAY (NO_ERROR) transition to `DRAINING` rather than `CLOSED`:

- Allows in-flight streams to complete normally
- Rejects new requests (connection reports `is_available() = False`)
- Connection expires once all streams complete

This fits naturally into the existing `HTTPConnectionState` enum pattern and makes state transitions easier to reason about.
  
### 3. Request Phase Tracking

Each stream tracks `headers_sent` and `body_sent` to determine potential side effects:
- If GOAWAY arrives before headers are sent → definitely no side effects
- If headers/body were sent → server may have processed the request

This information flows through `ConnectionGoingAway.may_have_side_effects` for retry decisions.

### 4. Connection Pool Retry Logic

The connection pool now handles `ConnectionGoingAway` with RFC-compliant retry semantics:

| Condition | Behavior |
|-----------|----------|
| `stream_id > last_stream_id` | Auto-retry (guaranteed unprocessed per RFC 7540) |
| Graceful shutdown + no side effects | Auto-retry (likely safe) |
| Otherwise | Raise `RemoteProtocolError` with context for application decision |

## Behavior Summary
  
| Scenario | Before | After |  
|----------|--------|-------|  
| Stream ID > last_stream_id | `ConnectionNotAvailable` (retry worked but was implicit) | Auto-retry with explicit RFC justification |
| Graceful shutdown, headers not sent | `RemoteProtocolError` or `ConnectionNotAvailable` | Auto-retry |
| Potentially processed request | Generic error | `RemoteProtocolError` with cause chain to `ConnectionGoingAway` |
| h2 CLOSED race condition | Generic protocol error | Proper `ConnectionGoingAway` with conservative assumptions |
| Server disconnect after GOAWAY | `RemoteProtocolError("Server disconnected")` | `ConnectionGoingAway` with retry context |

## Design Decisions

**Why a new connection state vs. a boolean flag?**
The `DRAINING` state fits the existing `HTTPConnectionState` pattern and makes state transitions explicit. A boolean would work but adds cognitive overhead when reasoning about valid state combinations.  

**Why track headers and body separately?**
Both are relevant for side-effect determination. Applications can use this granular information - for example, a read-only GET that sent headers but no body might be safer to retry than a POST with body sent.

**Why raise `RemoteProtocolError` instead of `ConnectionGoingAway` for potentially-processed requests?**
Balances API simplicity with power. Most users catch `RemoteProtocolError`; advanced users can inspect `__cause__` to access the full `ConnectionGoingAway` context. Since `ConnectionGoingAway` extends `ConnectionNotAvailable`, it's also catchable directly for those who need it.

**Why conservative assumptions when h2 is CLOSED without GOAWAY event?**
When the connection is closed but we haven't processed a GOAWAY event, we can't know the true `last_stream_id`. Assuming the current stream may have been processed is the safe default - better to surface an error than silently retry a potentially-processed request.

## Related Issues

- https://github.com/encode/httpx/issues/3324
- https://github.com/encode/httpcore/issues/45
- https://github.com/encode/httpcore/issues/730
- https://github.com/encode/httpcore/discussions/1041

## Test Plan

- [X] Unit tests for `ConnectionGoingAway` exception properties
- [X] Unit tests for DRAINING state transitions
- [ ] Integration tests simulating GOAWAY during request
- [ ] Integration tests for race condition scenarios (h2 CLOSED before event processing)
- [X] Verify existing tests still pass

---

# Checklist

- [X] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.

### DIFF
--- a/httpcore/__init__.py
+++ b/httpcore/__init__.py
@@ -19,6 +19,7 @@ from ._backends.mock import AsyncMockBackend, AsyncMockStream, MockBackend, Mock
 from ._backends.sync import SyncBackend
 from ._exceptions import (
     ConnectError,
+    ConnectionGoingAway,
     ConnectionNotAvailable,
     ConnectTimeout,
     LocalProtocolError,
@@ -114,6 +115,7 @@ __all__ = [
     "default_ssl_context",
     "SOCKET_OPTION",
     # exceptions
+    "ConnectionGoingAway",
     "ConnectionNotAvailable",
     "ProxyError",
     "ProtocolError",
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -261,7 +261,8 @@ class AsyncConnectionPool(AsyncRequestInterface):
                         # Likely safe to retry.
                         continue
                     else:
-                        # Request may have been processed. Propagate error with context so application can decide whether to retry.
+                        # Request may have been processed. Propagate error with context so application can decide
+                        # whether to retry.
                         msg = "GOAWAY recieved: request may have been processed"
                         # QUESTION: What is the best way to propagate the context for the applications?
                         raise RemoteProtocolError(msg) from exc
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -7,7 +7,12 @@ import typing
 
 from .._backends.auto import AutoBackend
 from .._backends.base import SOCKET_OPTION, AsyncNetworkBackend
-from .._exceptions import ConnectionNotAvailable, UnsupportedProtocol
+from .._exceptions import (
+    ConnectionGoingAway,
+    ConnectionNotAvailable,
+    RemoteProtocolError,
+    UnsupportedProtocol,
+)
 from .._models import Origin, Proxy, Request, Response
 from .._synchronization import AsyncEvent, AsyncShieldCancellation, AsyncThreadLock
 from .connection import AsyncHTTPConnection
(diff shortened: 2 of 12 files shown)


bayoumi17m:
All checks are passing and I'm happy to make any changes if needed!

I've been using this example to reproduce and confirm the resolution:

https://github.com/encode/httpx/discussions/3549#discussioncomment-13452639

Issue #1048, opened 2025-12-21
Python 3.13.1 Incompatibility with httpcore 1.0.9 - IntEnum TypeError


yshevchenko128:
### Environment

- Python Version: 3.13.1
- Operating System: macOS (Apple Silicon)
- httpcore Version: 1.0.9
- httpx Version: 0.28.1

### Description

When using httpcore 1.0.9 with Python 3.13.1, instantiating an httpx.AsyncClient() raises a TypeError during the import of httpcore internals. The error occurs when Python 3.13's stricter enum.IntEnum validation encounters the HTTPConnectionState enum definition in httpcore/_sync/http11.py.

### Error Traceback

Traceback (most recent call last):
```shell
...
    async with httpx.AsyncClient() as client:
  File ".venv/lib/python3.13/site-packages/httpx/_client.py", line 1402, in __init__
    self._transport = self._init_transport(
  File ".venv/lib/python3.13/site-packages/httpx/_transports/default.py", line 294, in __init__
    import httpcore
  File ".venv/lib/python3.13/site-packages/httpcore/__init__.py", line 1, in <module>
    from ._api import request, stream
  File ".venv/lib/python3.13/site-packages/httpcore/_api.py", line 7, in <module>
    from ._sync.connection_pool import ConnectionPool
  File ".venv/lib/python3.13/site-packages/httpcore/_sync/__init__.py", line 1, in <module>
    from .connection import HTTPConnection
  File ".venv/lib/python3.13/site-packages/httpcore/_sync/connection.py", line 16, in <module>
    from .http11 import HTTP11Connection
  File ".venv/lib/python3.13/site-packages/httpcore/_sync/http11.py", line 36, in <module>
    class HTTPConnectionState(enum.IntEnum):
  File "/Users/user123/.local/share/uv/python/cpython-3.13.1-macos-aarch64-none/lib/python3.13/enum.py", line 568, in __new__
    enum_class = super().__new__(metacls, cls, bases, classdict, **kwds)
  File "/Users/user123/.local/share/uv/python/cpython-3.13.1-macos-aarch64-none/lib/python3.13/enum.py", line 268, in __set_name__
    enum_member = enum_class._new_member_(enum_class, *args)
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'dict'
```

### Steps to Reproduce

1. Create a virtual environment with Python 3.13.1
2. Install httpcore==1.0.9 and httpx==0.28.1
[...]

PR #1060, opened 2026-02-04
Fix TypeError due to IntEnum

tlogemann:
# Summary

The issue is described here: https://github.com/encode/httpcore/issues/1048

The problem is that at least python 3.12 raises a TypeError for the used IntEnum when using from httpx.

If this is not the intended way of fixing, feel free to close the PR. It was just a quick fix for me.

# Checklist

- [ ] I've added a test for each change that was introduced.
- [x] I tried as much as possible to make a single atomic change


tlogemann:
The `TypeError` by `IntEnum` seems not be caused by httpcore, as I also get it from other libraries.

PR #1067, opened 2026-03-22
Fix SSL certificate verification failure for hostnames with trailing dots (#1063)

### LINKED ISSUE #1063: When using httpx in a sync setting, we noticed that host names with trailing dots such as, `myhost.mycompany.internal.` (to mark them as fully qualified in the dns) were failing with
When using httpx in a sync setting, we noticed that host names with trailing dots such as, `myhost.mycompany.internal.` (to mark them as fully qualified in the dns) were failing with

```
httpx.ConnectError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Host name mismatch, certificate is not valid for 'myhost.mycompany.internal.'. (_ssl.c:1032)
```

However, the same request without the trailing dot in the host name works just fine.

I've seen this issue a few times in various libraries and in all cases, it was solved by stripping off the dot off the end of the host name before sending it to ssl.

Note, the ssl team in python has said this is an application layer issue. (https://bugs.python.org/issue31997).

Here is what urllib3 does for this, https://github.com/urllib3/urllib3/blob/9ff0acf2391c80b363ddb08a566f38d707dc9826/src/urllib3/connection.py#L170-L186 if that helps at all.

For httpx and httpcore, I think for the sync side, the change should be in

httpcore.._backends.sync.py:48

```
diff --git a/httpcore/_backends/sync.py b/httpcore/_backends/sync.py
index 4018a09..94edf08 100644
--- a/httpcore/_backends/sync.py
+++ b/httpcore/_backends/sync.py
@@ -45,7 +45,7 @@ class TLSinTLSStream(NetworkStream):  # pragma: no cover
         self.ssl_obj = ssl_context.wrap_bio(
             incoming=self._incoming,
             outgoing=self._outgoing,
-            server_hostname=server_hostname,
+            server_hostname=server_hostname.rstrip(".") if server_hostname else None,
         )

         self._sock.settimeout(timeout)
```

Forgive my lack of knowledge on the async side but I think any fix there would be in the downstream libraries, anyio and trio.

_Originally posted by @rjduffner in https://github.com/encode/httpcore/discussions/1039_

r266-tech:
## Problem

When using httpx/httpcore with FQDNs that have trailing dots (e.g., `myhost.mycompany.internal.`), SSL certificate verification fails with:

```
httpx.ConnectError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Host name mismatch, certificate is not valid for 'myhost.mycompany.internal.'.
```

This is a known issue — Python's `ssl` module doesn't handle trailing dots in `server_hostname`. The Python SSL team has acknowledged this as an application-layer issue: https://bugs.python.org/issue31997

## Solution

Strip the trailing dot from `server_hostname` before passing it to SSL backends. This follows the same approach used by [urllib3](https://github.com/urllib3/urllib3/blob/9ff0acf2391c80b363ddb08a566f38d707dc9826/src/urllib3/connection.py#L170-L186).

The fix is applied at the connection level (where `server_hostname` is constructed) rather than in each backend, keeping the change minimal and centralized across 6 files:

- `httpcore/_async/connection.py`
- `httpcore/_sync/connection.py`
- `httpcore/_async/http_proxy.py`
- `httpcore/_sync/http_proxy.py`
- `httpcore/_async/socks_proxy.py`
- `httpcore/_sync/socks_proxy.py`

DNS resolution still uses the original hostname with the trailing dot — only the SSL SNI is normalized.

Fixes #1063

### DIFF
--- a/httpcore/_async/connection.py
+++ b/httpcore/_async/connection.py
@@ -148,8 +148,9 @@ class AsyncHTTPConnection(AsyncConnectionInterface):
 
                     kwargs = {
                         "ssl_context": ssl_context,
-                        "server_hostname": sni_hostname
-                        or self._origin.host.decode("ascii"),
+                        "server_hostname": (
+                            sni_hostname or self._origin.host.decode("ascii")
+                        ).rstrip("."),
                         "timeout": timeout,
                     }
                     async with Trace("start_tls", logger, request, kwargs) as trace:
--- a/httpcore/_async/http_proxy.py
+++ b/httpcore/_async/http_proxy.py
@@ -309,7 +309,7 @@ class AsyncTunnelHTTPConnection(AsyncConnectionInterface):
 
                 kwargs = {
                     "ssl_context": ssl_context,
-                    "server_hostname": self._remote_origin.host.decode("ascii"),
+                    "server_hostname": self._remote_origin.host.decode("ascii").rstrip("."),
                     "timeout": timeout,
                 }
                 async with Trace("start_tls", logger, request, kwargs) as trace:
--- a/httpcore/_async/http_proxy.py
+++ b/httpcore/_async/http_proxy.py
@@ -309,7 +309,9 @@ class AsyncTunnelHTTPConnection(AsyncConnectionInterface):
 
                 kwargs = {
                     "ssl_context": ssl_context,
-                    "server_hostname": self._remote_origin.host.decode("ascii").rstrip("."),
+                    "server_hostname": self._remote_origin.host.decode("ascii").rstrip(
+                        "."
+                    ),
                     "timeout": timeout,
                 }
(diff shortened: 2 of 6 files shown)


r266-tech:
Rebased check: branch is up to date with master, and the lint failure has been fixed (a line exceeding the ruff format line length in `_async/http_proxy.py`). CI is now green across all Python versions (3.8-3.14).

The linked issue #1063 is still open, so this remains ready for review whenever maintainers have bandwidth. The change strips the trailing dot from `server_hostname` only (mirroring urllib3's approach), keeping the original FQDN for DNS resolution, and covers direct connections, HTTP tunnel proxies, and SOCKS proxies in both sync and async variants.

PR #1084, opened 2026-05-20
Drop stale unavailable pool connections

Mirochill:
# Summary

- remove an assigned connection from the pool if it still reports itself as available after raising `ConnectionNotAvailable`
- prevent HTTP/2 connections that have received GOAWAY from advertising availability
- add sync and async regression coverage for stale available connections

Closes #1066.

# Checklist

- [x] Documentation is not updated because this is an internal connection-pool bug fix.

# Validation

- Not run locally
- Static validation only: `git diff --check`.
- Relying on GitHub Actions for remote validation.

### DIFF
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -242,6 +242,17 @@ class AsyncConnectionPool(AsyncRequestInterface):
                     #
                     # In this case we clear the connection and try again.
                     pool_request.clear_connection()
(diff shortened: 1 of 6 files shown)


mbeijen:
There are multiple PRs for similar issues open in this repository; but there has not been much activity lately. In httpxyz, most of these issues have now been fixed and I've been upstreaming fixes (not these ones yet) to [httpx2](https://github.com/pydantic/httpx2/).

PR #1035, opened 2025-09-14
Fix 2 bugs and improve performance from O(N²+NxM) to O(N+M) in _assign_requests_to_connections

VictorPrins:
# Summary

Fixes **2 logic bugs** in `_assign_requests_to_connections` and improves complexity from O(N²+NxM) to O(N+M) where N=connections, M=requests. 

Although I attempted a smaller PR with an atomic change, I ultimately decided for a redesign of the method because the algorithm design of the method was fundamentally flawed.

## Motivation
`_assign_requests_to_connections` is in the hot path of this library and the bugs and complexity issue were severly slowing down my application.

## Bug 1: `max_keepalive_connections` is not respected
This causes idle connections to be evicted unnecessarily. This bug is also addressed in [this PR](https://github.com/encode/httpcore/pull/1000) (currently approved but not merged).

The bug is on lines 294-295. The expression on line 294 simply returns `len(self._connections)`, ignoring the number of idle connections, and compares this number to `self._max_keepalive_connections`.
https://github.com/encode/httpcore/blob/5974b03c7df89d3ee4e23779900d5349d550753c/httpcore/_async/connection_pool.py#L294-L295

[This test](https://github.com/VictorPrins/httpcore/blob/85296a47dd0aaf572449c19936625861edbf7df2/tests/_async/test_connection_pool.py#L836) demonstrates the bug. This test fails with the current code on master, but passes with the new implementation of this PR.

## Bug 2: multiple requests are assigned to the same HTTP/1.1 connection, leading to `ConnectionNotAvailable` exceptions

The bug is on lines 322-323. An available connection is assigned to a request, but the status of that connection is not actually changed, resulting in that connection also being assigned to the next request in the queue.

https://github.com/encode/httpcore/blob/5974b03c7df89d3ee4e23779900d5349d550753c/httpcore/_async/connection_pool.py#L322-L323

[This test](https://github.com/VictorPrins/httpcore/blob/ce0d1b29754c30785c607e9af2e23729ba8af2d5/tests/_async_only/test_connection_pool.py#L14) demonstrates the bug. This test fails with the current code on master, but passes with the new implementation of this PR.

## Perf: Reduce complexity from quadratic to linear
This PR reduces time complexity from O(N²+NxM) to O(N+M) where N=connections, M=requests.
In a typical use case, with for instance 100 connections in the pool and 500 requests in the queue this reduces complexity dramatically, especially since `_assign_requests_to_connections` is in the hot path of this library because it will be called twice for every request (on start and on finish).

[...]

PR #1094, opened 2026-07-16
Reclaim orphaned pool connections after cancellation

### LINKED ISSUE #1093: Cancelling requests under pool contention permanently leaks connection slots (AsyncConnectionPool → PoolTimeout)
## Summary

When requests to `AsyncConnectionPool` are **cancelled while the pool is under contention**, connections can be left in a state that the pool never reclaims. They accumulate until every slot is consumed, after which **all subsequent requests fail with `PoolTimeout`** even though the origin server is healthy and the event loop is fine. Only recreating the pool (process restart) recovers.

This reproduces on **httpcore 1.0.9** (latest) and appears to be the mechanism behind #658 (cancellation not releasing connections) and #830 (pool hangs after a cancelled connection).

## Reproduction

Pure httpcore + stdlib asyncio (tiny local HTTP/1.1 server), no httpx:

```python
import asyncio, contextlib, httpcore

MAX_CONNECTIONS, CONCURRENCY, CHURN_SECONDS = 10, 20, 5
TIMEOUT = {"connect": 2.0, "read": 2.0, "write": 2.0, "pool": 3.0}

async def _serve(reader, writer):
    with contextlib.suppress(Exception):
        while True:
            head = b""
            while b"\r\n\r\n" not in head:
                chunk = await reader.read(4096)
                if not chunk:
                    return
                head += chunk
            writer.write(b"HTTP/1.1 200 OK\r\nContent-Length: 2\r\n\r\n{}")
            await writer.drain()

async def _request(pool, port):
    return await pool.request(
        b"GET", f"http://127.0.0.1:{port}/",
        headers=[(b"Host", f"127.0.0.1:{port}".encode())],
        extensions={"timeout": TIMEOUT})

def _census(pool):
    conns = pool.connections
    stuck = sum(1 for c in conns if not c.is_idle() and not c.is_closed())
    return f"connections={len(conns)} (stuck={stuck}), waiting requests={len(pool._requests)}"

async def _trial(port, *, cancel):
    pool = httpcore.AsyncConnectionPool(max_connections=MAX_CONNECTIONS)
    deadline = asyncio.get_event_loop().time() + CHURN_SECONDS
    async def worker():
        while asyncio.get_event_loop().time() < deadline:
            task = asyncio.create_task(_request(pool, port))
            if cancel:
                await asyncio.sleep(0.005)   # start the request, then cancel mid-flight
                task.cancel()
                with contextlib.suppress(BaseException):
                    await task
            else:
                with contextlib.suppress(BaseException):
                    await task
    with contextlib.suppress(Exception):
        await asyncio.wait_for(asyncio.gather(*[worker() for _ in range(CONCURRENCY)],
                                              return_exceptions=True), timeout=CHURN_SECONDS + 10)
    await asyncio.sleep(2)   # let the pool settle; a healthy pool goes fully idle
    label = "cancelled" if cancel else "control  "
    try:
        r = await asyncio.wait_for(_request(pool, port), timeout=8)
        await r.aclose(); probe = f"HTTP {r.status}"
    except Exception as exc:
        probe = type(exc).__name__
    print(f"[{label}] {_census(pool)} -> probe: {probe}")
    with contextlib.suppress(Exception):
        await asyncio.wait_for(pool.aclose(), timeout=3)

async def main():
    print(f"httpcore {httpcore.__version__}")
    server = await asyncio.start_server(_serve, "127.0.0.1", 0)
    port = server.sockets[0].getsockname()[1]
    await _trial(port, cancel=False)
    await _trial(port, cancel=True)
    server.close()
    with contextlib.suppress(Exception):
        await asyncio.wait_for(server.wait_closed(), timeout=3)

asyncio.run(main())
```

### Output

```
httpcore 1.0.9
[control  ] connections=10 (stuck=0), waiting requests=0  -> probe: HTTP 200
[cancelled] connections=10 (stuck=10), waiting requests=3 -> probe: PoolTimeout
```

Identical load; the only difference is that the `cancelled` trial cancels each in-flight request. The control pool returns fully idle and serves the probe; the cancelled pool is wedged at 10/10 stuck connections and the probe (and every future request) raises `PoolTimeout`.

## Expected vs actual

- **Expected:** cancelling a request releases (or eventually reclaims) its pool slot, as it does for a serial/uncontended cancel.
- **Actual:** under contention the slot is never reclaimed; the pool degrades monotonically to permanent `PoolTimeout`.

## Root cause analysis

Dumping the wedged pool, the stuck connections are overwhelmingly in the **`CONNECTING`** state (`AsyncHTTPConnection` with `_connection is None` and `_connect_failed is False`):

```
<AsyncHTTPConnection [CONNECTING]>   x7
<AsyncHTTPConnection ['http://127.0.0.1:...', HTTP/1.1, ACTIVE, Request Count: 2]>   x2
<AsyncHTTPConnection ['http://127.0.0.1:...', HTTP/1.1, NEW, Request Count: 0]>      x1
```

These arise when `_assign_requests_to_connections()` creates a connection and assigns it to a pooled request, but that request's task is **cancelled before it ever calls `connection.handle_async_request()`** — so `_connect()` never starts and never fails.

Two properties then combine to make the slot unrecoverable:

1. **Not reapable.** In `_async/connection.py`, for a not-yet-established connection, `is_idle()`, `is_closed()`, and `has_expired()` **all return `self._connect_failed`** (i.e. `False`). `_async/connection_pool.py::_assign_requests_to_connections()` only removes connections that are `is_closed()` / `has_expired()` / surplus-`is_idle()` — a `CONNECTING` orphan matches **none** of these, so it is never removed from `self._connections`.
2. **Not reusable.** For the same not-yet-established connection, `is_available()` returns `False` for HTTP/1.1 (only HTTP/2 is treated as available while connecting). So the orphan cannot be handed to a queued request either.

Net: the slot is held forever by a connection that will never connect, never fail, never expire, and can never be reused.

There is also a secondary effect that defeats naive pool-side reaping: cancelled requests can remain in `self._requests`, so if orphaned connections *are* forcibly reclaimed, the freed slots are immediately re-consumed by those stranded requests.

## Notes for whoever fixes this (attempts that do **not** work)

We explored several fixes against the repro; sharing so they're not re-tried:

- Closing the owning connection in `AsyncConnectionPool.handle_async_request`'s `except BaseException` — **deadlocks**: `await conn.aclose()` blocks when the connection is cancelled *mid-connect* (live socket / in-flight connect).
- Merely removing the owning connection from `self._connections` in that `except` — insufficient: the excepting request rarely owns the connection at cancel time, and stranded requests re-consume the freed slot.
- Extending `AsyncShieldCancellation` to cover the whole of `PoolByteStream.aclose()` — no effect; the dominant orphan is `CONNECTING`, not a stream-close race.

A correct fix seems to need both (a) making an abandoned not-yet-established connection reclaimable (e.g. a distinct state, or `is_closed()`/reaper awareness), with a guaranteed non-blocking teardown for the never-started case, and (b) ensuring a cancelled request is removed from `self._requests` so reclaimed slots aren't re-consumed. Happy to help test candidate fixes against the repro.

## HTTP/1.1 vs HTTP/2

This reproduces on an HTTP/1.1 pool and appears largely HTTP/1.1-specific:

- For a not-yet-established connection, `is_available()` returns `False` for HTTP/1.1 but `True` when HTTP/2 is enabled (`self._http2 and (https or not http1) ...`). So an HTTP/2-capable `CONNECTING` connection can still be handed to queued requests, whereas the HTTP/1.1 one is stuck.
- HTTP/1.1 is one-request-per-connection, so a cancelled request wedges a whole slot and the pool must grow many connections to stay busy (contention). HTTP/2 multiplexes, so the pool rarely reaches contention and a cancellation affects a stream rather than a connection.

I have not tested whether an HTTP/2 pool has a rarer analogue, so I'm not claiming HTTP/2 is immune, just that the exhaustion demonstrated here is an HTTP/1.1 pool phenomenon.

## Environment

- httpcore 1.0.9 (`httpcore[asyncio]`), anyio/asyncio backend
- Python 3.12
- Reproduces deterministically with the script above; real-world trigger is a long-lived HTTP/1.1 pool under concurrent load where callers routinely cancel in-flight requests (e.g. client disconnects / task cancellation), which slowly accumulates orphans until the pool wedges.

---
🤖 Investigated and drafted with [Claude Code](https://claude.com/claude-code) (Claude Opus 4.8, 1M context).

tianrking:
Fixes #1093

## Summary
- Reclaim a non-idle connection when no remaining pool request owns it.
- Apply the same cleanup invariant to the synchronous and asynchronous pools.
- Add an async regression that cancels a request after assignment but before dispatching it to the connection.

## Root cause
A cancelled async request can be removed after receiving a newly created connection but before calling that connection. The connection is neither idle nor closed, and no longer belongs to any request, so it remains in the pool and consumes a connection slot indefinitely.

## Validation
- `python -m ruff format httpcore tests --check`
- `python -m ruff check httpcore tests`
- `python -m mypy httpcore/_async/connection_pool.py httpcore/_sync/connection_pool.py tests/_async/test_connection_pool.py`
- `python -m pytest -q` (216 passed, 6 xpassed)

### DIFF
--- a/httpcore/_async/connection_pool.py
+++ b/httpcore/_async/connection_pool.py
@@ -297,6 +297,12 @@ class AsyncConnectionPool(AsyncRequestInterface):
                 # log: "closing idle connection"
                 self._connections.remove(connection)
                 closing_connections.append(connection)
+            elif not connection.is_idle() and not any(
+                request.connection is connection for request in self._requests
+            ):
+                # log: "closing orphaned connection"
+                self._connections.remove(connection)
+                closing_connections.append(connection)
 
         # Assign queued requests to connections.
         queued_requests = [request for request in self._requests if request.is_queued()]
--- a/httpcore/_sync/connection_pool.py
+++ b/httpcore/_sync/connection_pool.py
@@ -297,6 +297,12 @@ class ConnectionPool(RequestInterface):
                 # log: "closing idle connection"
                 self._connections.remove(connection)
                 closing_connections.append(connection)
+            elif not connection.is_idle() and not any(
+                request.connection is connection for request in self._requests
+            ):
+                # log: "closing orphaned connection"
+                self._connections.remove(connection)
+                closing_connections.append(connection)
 
         # Assign queued requests to connections.
         queued_requests = [request for request in self._requests if request.is_queued()]
--- a/tests/_async/test_connection_pool.py
+++ b/tests/_async/test_connection_pool.py
@@ -1,15 +1,12 @@
 import logging
 import typing
 
-import anyio
 import hpack
 import hyperframe.frame
 import pytest
 import trio as concurrency
 
 import httpcore
-from httpcore._async import connection_pool
-from httpcore._async.interfaces import AsyncConnectionInterface
 
 
 @pytest.mark.anyio
@@ -452,75 +449,6 @@ async def test_connection_pool_with_connect_exception():
         "connection.connect_tcp.started",
         "connection.connect_tcp.failed",
     ]
-
-
-@pytest.mark.anyio
-async def test_connection_pool_removes_connection_after_request_cancellation(
-    monkeypatch: pytest.MonkeyPatch,
-) -> None:
-    request_assigned = anyio.Event()
-
-    class WaitingPoolRequest(connection_pool.AsyncPoolRequest):
-        async def wait_for_connection(
-            self, timeout: float | None = None
-        ) -> AsyncConnectionInterface:
-            await super().wait_for_connection(timeout)
-            request_assigned.set()
-            await anyio.sleep_forever()
-            raise AssertionError("Pool request must be cancelled")
-
-    monkeypatch.setattr(connection_pool, "AsyncPoolRequest", WaitingPoolRequest)
-
-    class ConnectingConnection(AsyncConnectionInterface):
-        def __init__(self) -> None:
-            self.closed = False
-
-        async def handle_async_request(
-            self, request: httpcore.Request
-        ) -> httpcore.Response:
-            raise AssertionError("Cancelled request must not reach the connection")
-
-        async def aclose(self) -> None:
-            self.closed = True
-
-        def can_handle_request(self, origin: httpcore.Origin) -> bool:
-            return True
-
-        def is_available(self) -> bool:
-            return False
-
-        def has_expired(self) -> bool:
-            return False
-
-        def is_idle(self) -> bool:
-            return False
-
-        def is_closed(self) -> bool:
-            return False
-
-        def info(self) -> str:
-            return "CONNECTING"
-
-    class TestPool(httpcore.AsyncConnectionPool):
-        def create_connection(
-            self, origin: httpcore.Origin
-        ) -> AsyncConnectionInterface:
-            return connection
-
-    connection = ConnectingConnection()
-    async with TestPool(max_connections=1) as pool:
-
-        async def request() -> None:
-            with pytest.raises(TimeoutError):
-                with anyio.fail_after(0.01):
-                    await pool.request("GET", "https://example.com/")
-
-        async with anyio.create_task_group() as task_group:
-            task_group.start_soon(request)
-            await request_assigned.wait()
-
-        assert pool.connections == []
-        assert connection.closed
 
 
 @pytest.mark.anyio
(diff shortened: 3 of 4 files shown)


mbeijen:
Hi, unfortunately, apparently httpx does no longer get new releases. Please note how these older PRs fix the same issue and have been open for a long time.

 - #986
 - #1035
 
Pydantic decided to fork this package to httpx2, because of accruing issues like this, and I fixed this very issue there --> https://github.com/pydantic/httpx2/pull/983

PR #1095, opened 2026-07-19
fix(http2): wait when outbound flow control window is negative

### LINKED ISSUE #1082: HTTP/2 large request body can fail with negative flow-control window
### Problem

A request with a large body failed before reaching the server with:

```text
httpx.LocalProtocolError: Cannot send 97519 bytes, flow control window is -49151
```

This surfaced through HTTPX. I'm filing it here because `encode/httpx` has issues disabled and HTTP/2 flow-control handling appears to live in httpcore.

### Expected behavior

The client should handle HTTP/2 request-body flow control internally. If the available window is too small, httpcore should split the body, wait for `WINDOW_UPDATE`, or otherwise apply backpressure. Application code should not have to manually size or buffer request bodies to avoid violating the HTTP/2 flow-control window.

### Actual behavior

httpcore appears to attempt to send a 97,519-byte chunk while the flow-control window is already negative, and the request fails with `LocalProtocolError`.

### Why this looks like a bug

The caller is not managing HTTP/2 DATA frame sizes or connection-window accounting; httpcore is. A negative flow-control window leaking into a user-visible send error suggests the transport is trying to write before flow control permits it, instead of buffering, chunking, or waiting for window credit.

If the connection is already in an unrecoverable state, the error should ideally make that explicit. As written, this reads like a client-side flow-control handling failure for large request bodies.

### Reproducer details

I do not yet have a minimal isolated script, but the failure mode is specific:

- HTTPX raised `httpx.LocalProtocolError`
- message: `Cannot send 97519 bytes, flow control window is -49151`
- trigger: sending a request with a large body over HTTP/2

I can add version numbers and a reduced repro once available.

l46983284-cpu:
## Summary

HTTP/2 request bodies could fail with `LocalProtocolError: Cannot send N bytes, flow control window is -M` when h2 reported a **negative** local flow-control window. `_wait_for_outgoing_flow` only waited while `flow == 0`, so negative windows were returned as `max_flow` and `send_data` raised.

This change waits on any **non-positive** flow (same as zero) until `WINDOW_UPDATE` credit is available.

## Test plan

- [x] `tests/_sync/test_http2.py::test_http2_wait_for_outgoing_flow_handles_negative_window`
- [x] `tests/_async/test_http2.py::test_http2_wait_for_outgoing_flow_handles_negative_window`
- [x] existing `test_http2_connection_with_flow_control`

Fixes #1082


alexchen-sys:
## CI residual

Root cause was **PR-caused style/type**, not the product wait fix:

- Job: Test Suite / Python 3.8 → `scripts/check` → mypy
- Errors: `tests/_async/test_http2.py:423` and sync twin — `Unused "type: ignore" comment [unused-ignore]`
- Other matrix jobs cancelled after first fail (`fail-fast`)

Fix: drop the unused ignores and monkeypatch via `setattr` in the #1082 regression only. Product `while flow <= 0` unchanged.

PR #1096, opened 2026-07-25
Fix nonexistent `iter_content()` reference in `stream()` docstring

cognis-digital:
The docstring for the public `httpcore.stream()` function tells users how to
read a streamed response body:

> If you want to access the response body you should either use
> `content = response.read()`, or `for chunk in response.iter_content()`.

But `Response` has no `iter_content()` method — the streaming iterator is
`iter_stream()` (and `aiter_stream()` for async). Running the documented
`for chunk in response.iter_content()` raises `AttributeError`.

The rest of the project already uses the correct name (`docs/quickstart.md`
documents `for chunk in response.iter_stream():`, and the tests use
`response.iter_stream()`). This one docstring drifted because doctests are
not run in CI.

This corrects the docstring to `iter_stream()` and adds a small regression
test pinning the documented streaming method names, so the documented API
can't silently drift to a nonexistent method again.

No library behaviour changes — docs-only correctness plus a guard test.


fallenmi (approved):
Reviewed with Codex.

I checked this on exact head `08901009255b95b7500f2a19f6695f9983e581ae`. The exact base/current `master` (`10a658221deb38a4c5b16db55ab554b0bf731707`) directs users to `Response.iter_content()`, which is not implemented, while `Response.iter_stream()` is the actual synchronous streaming API. On the head, the docstring and public API now agree.

Validation:
- repeated base/head doc-to-API oracle 100x (base mismatch, head match)
- `pytest -q tests/test_models.py::test_response_streaming_api_names` -> 1 passed
- focused Ruff format/check and mypy -> clean
- `git diff --check` -> clean
- live matrix: 7/7 successful

No blocking issues found.

PR #1111, opened 2026-09-12
Keep the pool from closing a connection assigned to a queued request

### LINKED ISSUE #1110: Connection pool can close a connection it has already assigned to a queued request
`_assign_requests_to_connections` assigns an idle connection to a queued request, but the connection stays `IDLE` until the waiting thread wakes up and calls `handle_request`. In between, another thread's pass still sees it as idle and closes it (surplus keep-alive, or "close an idle connection to make room" at `max_connections`). `HTTP11Connection.close` doesn't check the state.

If the request hasn't started yet, `handle_request` raises `ConnectionNotAvailable` and the request is re-queued. If it has, the socket is closed under a request waiting for its response. On Linux the blocked `poll()` isn't woken by the close, so the request hangs until its read timeout.

Seen with the sync pool through httpx and the openai client, 342 threads, `max_connections=100`, `max_keepalive_connections=20`: 2 to 3 requests per run hang for the full read timeout. Instrumenting the pool shows the closed connection was referenced by a `PoolRequest` in `_requests` when the pass decided to close it, and `close` running on an `ACTIVE` connection at the time of each timeout.

httpcore 1.0.9 and master (10a6582). Fix in #1111.

moldhouse:
Closes #1110.

`_assign_requests_to_connections` now treats a connection referenced by a pending `PoolRequest` as not idle, both when cleaning up connections and when picking an idle connection to close for making room. Without this a connection handed to a queued request could be closed by a concurrent pass before, or worse after, the request started on it.

The regression test drives the pool's assignment pass directly, since the race cannot be reproduced deterministically through the public API. `scripts/unasync.py` gains a `httpcore._async` -> `httpcore._sync` substitution so the test can import `AsyncPoolRequest`.


sylvesterkaczmarek (commented):
I think the “spoken for” set needs to apply to reuse too. After assigning an idle HTTP/1.1 connection, it can still report `is_available()` until that request actually starts, and `available_connections` doesn’t exclude `assigned_connections`. A second queued request in the same pass can therefore get the same connection. Could this filter out already-assigned connections as well?

PR #1116, opened 2026-09-17
Retry response cleanup after cancellation

### LINKED ISSUE #1115: Interrupted response close cannot be retried
### Problem

If response cleanup is interrupted, both `PoolByteStream.aclose()` and `HTTP11ConnectionByteStream.aclose()` mark themselves closed before their awaited cleanup finishes. A later `aclose()` therefore returns immediately instead of completing the connection-pool bookkeeping.

This is reachable when an asyncio task is cancelled again while unwinding an earlier cancellation. The result is a closed HTTP/1.1 connection that remains counted as active in the pool, eventually exhausting `max_connections`.

The behavior was reported with a full HTTPX reproducer in encode/httpx#3782.

### Cause

Both async byte-stream wrappers assign `_closed = True` before awaiting the inner close operation. If cancellation escapes that await, the state says cleanup completed even though the pool request was never removed.

### Expected behavior

An interrupted close should remain retryable. A subsequent `aclose()` should finish closing the HTTP/1.1 stream and remove the request and connection from the pool. Concurrent close calls should remain serialized.

pentaoa:
# Summary

Fixes #1115.

A response close interrupted by cancellation used to leave both byte-stream wrappers marked closed before their cleanup completed. Any later close then returned immediately, leaving the pool request and its HTTP/1.1 connection counted as active. This is the connection leak reported in encode/httpx#3782.

The wrappers now serialize close calls and set `_closed` only after the inner close and pool bookkeeping succeed. An interrupted close therefore remains retryable. The generated synchronous implementations are kept in step with the async sources, as required by the repository.

The regression test interrupts the network close with `CancelledError`, confirms the pool is still occupied, retries `response.aclose()`, and verifies that the request and connection are removed. The test fails against `master` with one active request and connection.

# Checklist

- [x] Documentation is unaffected because this restores the existing close contract.

Validation:

- Relevant async and sync HTTP/1.1, connection-pool, and cancellation tests: 99 passed, 6 existing xfails xpassed
- Repository-pinned Ruff formatting and lint checks
- Mypy on the four changed implementation files
- `scripts/unasync.py --check`

Review follow-up: detached connection cleanup now remains retryable after pool bookkeeping. The regression covers both unread and fully read responses. On Windows, 97 related tests pass and 6 existing xfails xpass; 3 immediate-expiry failures also reproduce on the previous commit. Ruff, Mypy and unasync checks pass. Wheel and sdist builds pass Twine 6.1.0 validation with explicit metadata 2.4, avoiding the CI failure caused by the newer Hatchling default.

AI assistance: OpenAI Codex assisted with the implementation, tests, and reported validation.


sylvesterkaczmarek (commented):
One retry hole still looks open here: `_closed` is set before `_close_connections(closing)`. If one of those connection closes raises, a second `aclose()` returns immediately and that cleanup is never retried. Could the flag move after `_close_connections()`, or otherwise keep those pending closes retryable?

pentaoa:
Fixed in dc2fb71. The response stream now retains detached connections until each close succeeds, then marks itself closed. Retries skip the already-completed request removal and resume pending cleanup.

The regression now covers both an unread response and a fully read response evicted with `max_keepalive_connections=0`. The latter failed before this update because the network close was never retried; both pass now, including a third close that must be a no-op.

Validation: 97 related tests passed, 6 existing xfails xpassed; Ruff, Mypy and `unasync --check` passed. Three immediate-expiry tests also fail on the previous commit on this Windows host and were excluded from that passing run.

CI then stopped at package validation: its pinned Twine rejects the metadata 2.5 emitted by the current Hatchling. Commit 6d5f158 explicitly selects metadata 2.4 for both distribution targets, retaining the existing Python support and tooling versions. The wheel and sdist both build and pass `twine check` with the repository-pinned Twine 6.1.0. CI is rerunning.

sylvesterkaczmarek (approved):
Rechecked 6d5f158b. Pending detached connections are now retained in _closing until each close succeeds, so a cancelled/failed cleanup can be retried without removing the pool request twice or losing the remaining connection cleanup. The regression covers unread and fully-read responses plus a third no-op close. My concern is resolved.

jbbottoms:
Independently checked `6d5f158bcad201a368d2d869d94882a2f865597e` against `master` at `10a658221deb38a4c5b16db55ab554b0bf731707`, through HTTPX's public streaming API with a real localhost HTTP/1.1 server. This confirms the improvement for the double-cancellation path reported in encode/httpx#3782, beyond the injected `CancelledError` test.

With `max_connections=3`, the same probe produced:

| Scenario | master | This PR |
| --- | --- | --- |
| Single cancellation, four sequential streams | Pool empty after each | Pool empty after each |
| Two `Task.cancel()` calls separated by `await asyncio.sleep(0)` | Active connections grew 1, 2, 3; fourth request raised `PoolTimeout` | Pool empty after all four; fourth request streamed successfully |

The caller uses `asyncio.shield(response.aclose())` in `finally`, as in the original report. Counts are taken after allowing that cleanup to run. No external server or mocked transport is involved.

Environment: Windows 11, Python 3.12.10, HTTPX 0.28.1, AnyIO 4.12.1, h11 0.16.0; httpcore installed editable from each checkout.

I also ran the cancellation, sync/async HTTP/1.1 and connection-pool test files: **97 passed, 1 xfailed, 5 xpassed, 3 failed**. All three failures were the immediate-expiry tests and reproduced on the baseline commit too. This is a focused HTTP/1.1 recovery check, not a full-suite or cross-platform approval.

<details>
<summary>Standalone localhost probe</summary>

Run with HTTPX 0.28.1 and the selected httpcore checkout installed in the same environment. The only socket opened is on loopback.

```python
"""Local-only reproduction of HTTPX #3782; no external requests."""
import asyncio
import contextlib
import json
import httpcore
import httpx


async def main():
    handlers = set()

    async def serve(reader, writer):
        handlers.add(asyncio.current_task())
        try:
            await reader.readuntil(b"\r\n\r\n")
            writer.write(b"HTTP/1.1 200 OK\r\nTransfer-Encoding: chunked\r\n\r\n2\r\nx\n\r\n")
            await writer.drain()
            await reader.read()
        finally:
            writer.close()
            await writer.wait_closed()
            handlers.discard(asyncio.current_task())

    server = await asyncio.start_server(serve, "127.0.0.1", 0)
    url = f"http://127.0.0.1:{server.sockets[0].getsockname()[1]}/"
    rows = []
    try:
        for double in (False, True):
            async with httpx.AsyncClient(limits=httpx.Limits(max_connections=3), timeout=0.5, trust_env=False) as client:
                for iteration in range(4):
                    ready = asyncio.Event()
                    async def consume():
                        async with client.stream("GET", url) as response:
                            try:
                                async for line in response.aiter_lines():
                                    ready.set()
                            finally:
                                await asyncio.shield(response.aclose())

                    task = asyncio.create_task(consume())
                    ready_task = asyncio.create_task(ready.wait())
                    done, _ = await asyncio.wait([task, ready_task], timeout=2, return_when=asyncio.FIRST_COMPLETED)
                    if ready_task not in done:
                        ready_task.cancel()
                        with contextlib.suppress(asyncio.CancelledError):
                            await ready_task
                        if task.done():
                            try:
                                await task
                            except Exception as exc:
                                rows.append(dict(double=double, iteration=iteration, error=type(exc).__name__))
                                break
                        raise RuntimeError("server did not start streaming")
                    task.cancel()
                    if double:
                        await asyncio.sleep(0)
                        task.cancel()
                    with contextlib.suppress(asyncio.CancelledError):
                        await task
                    # Allow shielded caller cleanup to finish before counting.
                    await asyncio.sleep(0.01)
                    pool = client._transport._pool
                    rows.append(dict(double=double, iteration=iteration, connections=len(pool.connections), active=sum(not c.is_idle() for c in pool.connections), requests=len(pool._requests)))
    finally:
        server.close()
        await server.wait_closed()
        for task in list(handlers):
            task.cancel()
        await asyncio.gather(*handlers, return_exceptions=True)
    print(json.dumps(dict(httpcore=httpcore.__file__, httpx=httpx.__version__, rows=rows), indent=2))


asyncio.run(main())
```
</details>

